PR #2078, opened 2019-05-08
Memory optimized (eliminating buffer memory per chare) jacobi-2d

pplimport:
*Original author: Ajay Nair*
*Original date: 2014-08-06 22:29:52*
*Original PR: https://charm.cs.illinois.edu/gerrit/349*

---

Change-Id: I91bbeec380758432d540a9c20cf1e428a0f64def


rbuch:
I don't think this is a useful example to distribute with the codebase, it's a Jacobi where the ghost layer data is sent to and retrieved from a group. While perhaps useful to shrink chare size, it creates migration issues and doesn't actually save any memory, making this unnecessary complexity.

PR #2079, opened 2019-05-08
Memory optimized (eliminating buffer memory per chare) jacobi-3D

pplimport:
*Original author: Ajay Nair*
*Original date: 2014-08-11 21:05:59*
*Original PR: https://charm.cs.illinois.edu/gerrit/354*

---

Change-Id: I9d5d81ff9d947297dcad3c1d7e3fbf658e38f8cb


rbuch:
I don't think this is a useful example to distribute with the codebase, it's a Jacobi where the ghost layer data is sent to and retrieved from a group. While perhaps useful to shrink chare size, it creates migration issues and doesn't actually save any memory, making this unnecessary complexity.

PR #2081, opened 2019-05-08
PAPI: Make event list dynamic and add more defaults

evan-charmworks:
*Original date: 2017-09-07 16:07:59*
*Original PR: https://charm.cs.illinois.edu/gerrit/2874*

---

Change-Id: Id188514d1f8c525fd95c8111172c8f9c4d3a52bd


matthiasdiener:
Should this include the changes to the manual? https://charm.cs.illinois.edu/gerrit/c/charm/+/2874/7/doc/projections/manual.tex

chin123:
> Should this include the changes to the manual? https://charm.cs.illinois.edu/gerrit/c/charm/+/2874/7/doc/projections/manual.tex

Yes, even in the master branch currently, PAPI is selecting the highest level of cache dynamically. We should probably change the manual to mention that the highest level of cache instructions readable by PAPI is chosen as the first and second events, as some systems do not support reading the total number of cache misses and instructions for all of L1, L2 and L3 cache.

rbuch:
> Should this include the changes to the manual? https://charm.cs.illinois.edu/gerrit/c/charm/+/2874/7/doc/projections/manual.tex

It shouldn't contain those changes (because they're no longer correct), but it should contain more up to date changes to the manual. I'll make those.

PR #2085, opened 2019-05-08
Replace the memPool lists with standard "mempool.h".  This makes one big cudaMallocHost, and vastly simplifies memory handling in the hybrid API.

pplimport:
*Original author: Dr. Orion Lawlor*
*Original date: 2016-12-20 22:00:51*
*Original PR: https://charm.cs.illinois.edu/gerrit/2070*

---

Change-Id: I8c8ba60bb8d48530284ebfdd97a1f110e4b4bba5


minitu:
This may be a good change, but will need to look into how this would affect applications that use the mempool such as ChaNGa.

Issue #947, opened 2016-01-13
AMPI support for MPI-2's MPI_Type_* routines


stwhite91:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/947*

---
http://www.mpich.org/static/docs/v3.1/www3/MPI_Type_create_resized.html

stwhite91:
*Original date: 2016-01-13 05:05:22*

---
MPI_Type related functionality from MPI-2.2 that is missing from AMPI:

MPI_Type_match_size (and MPI_TYPECLASS_INTEGER, MPI_TYPECLASS_REAL, MPI_TYPECLASS_COMPLEX)

-MPI_Type_dup-
-MPI_Type_get_true_extent- (and MPI_Type_get_true_extent_x)
-MPI_Type_create_indexed_block-
-MPI_Type_create_hindexed_block-
-MPI_Type_create_resized-
-MPI_Type_create_subarray-
-MPI_Type_get_name-
-MPI_Type_set_name-
-MPI_Type_set_attr-
-MPI_Type_get_attr-
-MPI_Type_delete_attr-
-MPI_ORDER_C-
-MPI_ORDER_FORTRAN-

-Strike-through- means it has been implemented since opening this issue.


No plan for support:
MPI_Pack_external
MPI_Pack_external_size
MPI_Unpack_external
MPI_Type_create_darray (defined in ROMIO, and possibly deprecated in MPI-4.0)
(darray constants: MPI_DISTRIBUTE_NONE, MPI_DISTRIBUTE_BLOCK, MPI_DISTRIBUTE_CYCLIC, and MPI_DISTRIBUTE_DFLT_DARG)
MPI_Type_subarray (defined in ROMIO)

stwhite91:
*Original date: 2016-01-14 23:40:08*

---
MPI_Type_create_subarray:

~~https://charm.cs.illinois.edu/gerrit/#/c/1001/~~ https://github.com/UIUC-PPL/charm/commit/548c8eb7f216cc49b59891a685a0212c7b89b2d2

stwhite91:
*Original date: 2016-02-10 15:38:15*

---
[...]

PR #2086, opened 2019-05-08
AMPI #947: Added support for Subarray datatype to CkDDT and AMPI.

pplimport:
*Original author: Edward Hutter*
*Original date: 2016-08-25 17:42:21*
*Original PR: https://charm.cs.illinois.edu/gerrit/1175*

---

Change-Id: I0bc88eda52e197a3af9ec354cc946c5b705da20d


matthiasdiener:
I think we are using ROMIO's implementation for these at the moment.

PR #2089, opened 2019-05-08
DDT: add PUP methods for derived datatypes

stwhite91:
*Original date: 2017-01-25 23:08:30*
*Original PR: https://charm.cs.illinois.edu/gerrit/2149*

---

PUP routines are needed to support RMA with derived datatypes. For now,
we restrict our RMA support to derived datatypes whose base types are
primitive types, hence the assertions in the PUP routines.

Change-Id: I50058750b0cfad59229e9730166df968766ae089


matthiasdiener:
These appear to be good to go?

Issue #2047, opened 2019-01-24
Potential GCC 8 Hang


rbuch:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2047*

---
A NAMD user has emailed the NAMD mailing list with the following:

> So you can take pure NAMD/Charmm++, build it with GCC8 compilers and
> try to run - you'll receive the same hang behavior (at least on our
> computers which are employs the same environment it's true).
> 
> Plumed team is already maintains NAMD-2.12 and 2.13 version, but once
> again - seems like the problem is related to the newest GCC compilers,
> not Plumed

> One important note: when hang occurs all dedicated CPUs (i. e., via
> namd2 +pN) are fully busy.

Thus, it seems like there may be some issue with gcc 8 or other new versions of the compiler in NAMD or in Charm++.

jcphill:
NAMD user has updated his report:
Just recompiled git version of NAMD with CUDA 9 and 10 (separately). New GCC is installed (8.3.0).
Now simulation doesn't stuck at the very beginning but after short time it naturally explodes for unknown reason. NAMD compiled with GCC-7.3.1 works fine (for both CUDA versions).

stwhite91:
Every hear anything more about this from the NAMD user? Should we close the issue?

jcphill:
I assume it is still a problem.

PR #2093, opened 2019-05-08
Cleanup (#2047): new license from OTM

ericjbohm:
*Original date: 2019-04-05 16:36:23*
*Original PR: https://charm.cs.illinois.edu/gerrit/5061*

---

specific description of non-commercial uses covered.

Change-Id: I6e316529bd71bc8681f8267910cf31fd156da692


LICENSE
```diff
@@ -1,148 +1,188 @@
...
+  of Computer Science, hereby grants You a non-commercial,
+  non-exclusive, non-transferable, royalty-free, restrictive license
+  to download, use, and create derivative works of the Software for
+  academic and research purposes only.
+
+License Restrictions.  Any uses of Software or Software derivative
+works beyond those granted in Section 2 above require a separate for a
+fee license.  To negotiate a license with additional rights, You may
+contact Illinois at the Office of Technology Management at
+otm@illinois.edu or Charmworks, Inc. at info@hpccharm.com."
```
matthiasdiener: Stray ".

LICENSE
```diff
@@ -1,148 +1,188 @@
...
+
+15 Waiver.  The failure of either party to enforce any provision of
+   this License shall not constitute a waiver of that right or future
+   enforcement of that or any other provision.
+
+Approved for legal form Illinois Counsel SJA 10/18
+
+University Technology No. TF08023 Commercial Non-Exclusive Software License
+
+
```
matthiasdiener: Stray blank lines.

ericjbohm: Export from Word creates very strange issues. Should be addressed in the fixup patch.

ericjbohm:
> Should we also add a (placeholder?) CITATION file to this PR?

Good point.  Do we have a template for one from the DOI registration site?

mprobson:
I think @rbuch set that up but it may not generate until our next release. He can correct me if I'm wrong.

ericjbohm:
see https://github.com/UIUC-PPL/charm/pull/2334 for example of CITATION new format.

mprobson:
Great, no reason not to merge this then. Unless we want to merge citation at the same time.

evan-charmworks:
Was this PR updated with the most recent version of the license as received from OTM after further discussion?

ericjbohm:
> Was this PR updated with the most recent version of the license as received from OTM after further discussion?

It was updated with the change to the citation and a little clarification.  What outstanding changes are missing?

PR #2100, opened 2019-05-08
Add Isomempool, an Isomalloc-specific batching memory allocation pool

evan-charmworks:
*Original date: 2019-05-08 04:09:28*
*Original PR: https://charm.cs.illinois.edu/gerrit/4591*

---

The design of Isomalloc requires it to interface with the system using
mmap, directly affecting virtual memory at the level of pages. This
means some form of pooling is required to avoid wasting memory when
servicing malloc calls smaller than the system page size.

The Isomempool uses a red-black tree to store the locations and sizes
of free space within the blocks provided by the system through mmap.
This tree is used to determine the best placement for user allocations.
Each user allocation is prefixed with a header that makes the allocation
a node of a doubly linked list. Deallocations mark the region available
but leave the header in place so that an adjacent deallocation can
merge the two empty regions.

Change-Id: I693de02c8750048f36c57962a20b41c42b9f665e


evan-charmworks:
I don't know how I could perform the equivalent of a self -2 on GitHub but this patch may regress AMPI+OpenMP.

matthiasdiener:
> I don't know how I could perform the equivalent of a self -2 on GitHub but this patch may regress AMPI+OpenMP.

I think by explicitly requesting changes (like I just did) you can block merging.

evan-charmworks:
I will make a new PR for the more substantial changes due for Isomalloc

PR #2101, opened 2019-05-08
Add IPv6 support

evan-charmworks:
*Original date: 2018-12-19 21:02:53*
*Original PR: https://charm.cs.illinois.edu/gerrit/4727*

---

Change-Id: I91d1f5a7dc722d351cbef0c8ac810cdd8335e5b6

### DIFF
--- a/src/arch/netlrts/machine-dgram.c
+++ b/src/arch/netlrts/machine-dgram.c
@@ -209,10 +209,8 @@ typedef struct FutureMessageStruct {
 typedef struct OtherNodeStruct
 {
   int nodestart, nodesize;
-  skt_ip_t IP;
   unsigned int mach_id;
-  unsigned int dataport;
-  struct sockaddr_in addr;
+  skt_ip_t addr;
 #if CMK_USE_TCP 
   SOCKET	sock;		/* for TCP */
 #endif
@@ -331,9 +329,6 @@ int CmiLongSendQueue(int forNode,int longerThan) {
 }
 #endif
 
-extern void CmiGmConvertMachineID(unsigned int *mach_id);
-extern void CmiAmmassoNodeAddressesStoreHandler(int pe, struct sockaddr_in *addr, int port);
-
 /* initnode node table reply format:
  +------------------------------------------------------- 
  | 4 bytes  |   Number of nodes n                       ^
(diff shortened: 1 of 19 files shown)


ericjbohm:
Is there any reason not to move forward with this change?

evan-charmworks:
> Is there any reason not to move forward with this change?

Yes, this PR needs more work to be mergeable due to portability issues. It worked well enough for the controlled environment of the Adaptivity Demo but even connecting it to an outside network was be enough to cause hangs during launch.

evan-charmworks:
I wish GitHub allowed changing an existing PR to a Draft PR, because this one is incomplete in at least the following ways:

1. There are issues with mixed use of IPv4 and IPv6 in the same session, and exclusive use of one or the other is too stringent a requirement to place on users. Currently the patch has Charmrun create a socket on only one of the two protocols, but from what I remember reading the correct approach is to always create both and use the one that is responded to. This primarily affects the case when hostnames are used in a nodelist, which may resolve to either IPv4 or IPv6. An issue may still arise with inter-node communication not having an address of the correct supported protocol.
2. If _link-local_ IPv6 addresses are used in a nodelist file and they need to be conveyed over the network to another machine, an issue arises in that they contain a suffix like `%eth0` that identifies the interface/device name on that machine that may not be applicable on any other machine. I don't know for certain how to address this. The solution may be a simple brute force approach: on the receiving end of the addresses, strip the interface name and apply the name of the one used to receive the data from Charmrun, and if that does not work then try every active interface that is present.
3. If hostnames are used in the nodelist, there should be some kind of DNS lookup caching to a file on disk because otherwise the init sequence is noticeably delayed by several seconds per-host. At the same time, if hostnames are used then both the hostname and the resolved address should be transmitted to the child nodes so that they can try the resolved address first, and use their own hostname lookup as a fallback.

evan-charmworks:
After encountering an issue where liveViz would not connect to Charmrun via its IPv6 address if it had an IPv4 address signed due to the aforementioned #1, I took another look at this patch. What's clear to me is that the skt_server API needs revision to create sockets on both protocols simultaneously and manage them as a set. That will take further refactoring, but at least this patch already touches every use of the API.

This patch also totally breaks `cpuTopoHandler` because it still assumes skt_ip_t is a char[4], and will require ripping out the use of Cmm* and replacing it with something specific to this code.

evan-charmworks:
Fixed the `cpuTopoHandler` thing for now. However we should reevaluate most/all direct uses of the skt_ip_t structure since it is not portable (such as a Charmrun running on an architecture other than the compute nodes) and potentially replace it my string comparing the hostname. Again, IPv6 link-local device names pose an obstacle.

evan-charmworks:
Just updating this for the Adaptivity Demo. All concerns above still apply.

Issue #1982, opened 2018-09-28
Use C++ STL data structures in conv-conds.C


stwhite91:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1982*

---
Once we switch Converse over to C++, we can rewrite the data structures in conv-conds.c with STL containers rather than the homegrown C lists and heap used there.

evan-charmworks:
*Original date: 2019-04-04 19:20:16*

---
https://charm.cs.illinois.edu/gerrit/c/charm/+/4932

stwhite91:
Needs some benchmarking and possibly performance optimization. Low priority for now so pushing off of 6.11

stwhite91:
PR #2103 implements this, but has a known failure

PR #2103, opened 2019-05-08
Cleanup #1982: Use C++ STL data structures in conv-conds

evan-charmworks:
*Original date: 2019-04-26 04:00:57*
*Original PR: https://charm.cs.illinois.edu/gerrit/4932*

---

Change-Id: Ie533057d1dba8db277388affce245dc8d2cb50e6

### DIFF
--- a/src/conv-core/conv-conds.C
+++ b/src/conv-core/conv-conds.C
@@ -5,6 +5,10 @@
 #include "converse.h"
 #include "charm-api.h"
 
+#include <vector>
+#include <deque>
+#include <queue>
+
 #if CMK_ERROR_CHECKING
 CpvDeclare(double, idleBeginWalltime); // used for determining the conditon for long idle
 #endif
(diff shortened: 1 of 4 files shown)


stwhite91:
Performance seems to be equivalent to the original C implementation in microbenchmarks, but this PR is failing tests/charm/megatest with:

"------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: Internal QD Error. Contact Developers.!"

stwhite91:
Fixed, it passes all tests now.

stwhite91 (approved):

stwhite91:
Fine-grained performance measured using Charm 1D array pingpong on 1 PE is unchanged (within the noise) with this PR from master.

stwhite91 (approved):

evan-charmworks:
Strange segfault in the Spack build. Unfortunately it doesn't show the most important part of the backtrace.

```
     22403      [1:15] bcast_subset_async_delayed 0x5555557717fd CsdScheduleForever
     22404      [1:16] bcast_subset_async_delayed 0x555555771a05 CsdScheduler
     22405      [1:17] bcast_subset_async_delayed 0x5555557bbdda
     22406      [1:18] bcast_subset_async_delayed 0x5555557bbe42
     22407      [1:19] libpthread.so.0 0x7ffff7f9d609
     22408      [1:20] libc.so.6 0x7ffff7b52133 clone
  >> 22409    Fatal error on PE 1> unknown signal
     22410    
     22411    real	0m0.230s
     22412    user	0m0.004s
     22413    sys	0m0.000s
  >> 22414    make[5]: *** [Makefile:48: smptest] Error 1
     22415    make[5]: Leaving directory '/tmp/runner/spack-stage/spack-stage-c
              harmpp-main-gqo6lhvz3f3e4gvegepesfz2szln437h/spack-src/tests/char
              m++/zerocopy/zc_post_async/bcast_subset_async'
```

stwhite91:
Looking at that ZC Post API Bcast test, it doesn't use any CcdCallbacks on its own (only what internal callbacks are used by the runtime).

src/conv-core/conv-conds.C
```diff
@@ -401,26 +247,23 @@ static void ccd_heap_remove(void)
  */
 static void ccd_heap_update(double curWallTime)
 {
-  ccd_heap_elem *h = CpvAccess(ccd_heap);
-  ccd_heap_elem *e = h+CpvAccess(ccd_heapmaxlen);
-  int i,ne=0;
+  auto & h = CpvAccess(ccd_heap);
+
+  std::vector<ccd_periodic_callback> expired;
```
stwhite91, line 252: Thinking about this more, we might want to hoist this vector to persistent state in order to avoid small heap allocations everytime we pull something out of the heap. Could be in a separate commit.

evan-charmworks: Despite the comment in this function, I don't see why we need to process the heap elements in two phases if handled correctly. Let's see if CI validates my thinking.

evan-charmworks:
The issue was actually a bug in my reimplementation. By default, `std::priority_queue` sorts in descending order, whereas we want to sort the time field in ascending order. I fixed that in the main commit. I also pushed other minor cleanup I made while debugging.

stwhite91 (approved):

matthiasdiener (approved):

PR #2104, opened 2019-05-08
Address data races exposed by TSan

evan-charmworks:
*Original date: 2019-03-18 23:21:44*
*Original PR: https://charm.cs.illinois.edu/gerrit/5027*

---

Change-Id: I457007c8adbab2d5930b0ef21f40b7aa10f03c14


evan-charmworks:
Should be redone from scratch if found to be something other than false positives.

PR #2105, opened 2019-05-08
AMPI: Clean up pointer refreshing after potential migration

evan-charmworks:
*Original date: 2019-05-07 16:24:00*
*Original PR: https://charm.cs.illinois.edu/gerrit/5095*

---

* Replace calls to thread->suspend() and thread->schedule() with new
  ampi::block() and ampi::yield() that refresh the parent pointer
* Use static member functions to enforce non-access of `this` after
  migration
* Replace missed `this` accesses
* Add __attribute__ ((warn_unused_result)) where helpful

Change-Id: Ifacc90178f0cfa81f677b364922bb2cfd2247652


src/libs/ck-libs/armci/armci_vp.C
```diff
@@ -195,7 +195,7 @@ void ArmciVirtualProcessor::get(pointer src, pointer dst,
   }*/
   thisProxy[src_proc].requestFromGet(src, dst, nbytes, thisIndex, -1);
   // wait for reply
-  thread->suspend();
+  TCharm * unused = thread->suspend();
```
matthiasdiener: Won't this cause lots of "variable set but not used" warnings?

evan-charmworks: I don't believe we have that warning enabled, or else we could get tons of other instances of it. I named the variable `unused` because I _think_ it will silence the warning at least on some systems but I'm not sure.

stwhite91: Eventually we can add C++17's [[maybe_unused]] attribute here and throughout the runtime

Issue #2030, opened 2018-11-19
tests/ampi/megampi crashes in MPI_Comm_free


evan-charmworks:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2030*

---
This failure shows up in autobuild every few days.

```
../../../bin/testrun  ./pgm +p2 +vp4  

Running on 2 processors:  ./pgm +vp4 
charmrun> /cygdrive/c/Program Files/Microsoft MPI/Bin/mpiexec -n 2  ./pgm +vp4 

Charm++> Running on MPI version: 2.0
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 2 processes, 1 worker threads (PEs) + 1 comm threads per process, 2 PEs total
Charm++> The comm. thread both sends and receives messages
Charm++ warning> fences and atomic operations not available in native assembly
Converse/Charm++ Commit ID: v6.9.0-0-gc3d50ef
Charm++> Disabling isomalloc because mmap() does not work.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> Running on 1 hosts (1 sockets x 4 cores x 2 PUs = 8-way SMP)
Charm++> cpu topology info is gathered in 0.016 seconds.
CharmLB> RandCentLB created.

job aborted:
[ranks] message

[0] terminated

[1] process exited without calling finalize

---- error analysis -----

[1] on CS-DEXTERITY
./pgm ended prematurely and may have crashed. exit code 0xc0000005

---- error analysis -----
make[3]: *** [Makefile:24: test] Error 127
make[2]: *** [Makefile:38: test-megampi] Error 2
make[1]: *** [Makefile:37: test-ampi] Error 2
make: *** [Makefile:165: test] Error 2
make[3]: Leaving directory '/home/nikhil/autobuild/mpi-win-x86_64-smp/charm/mpi-win-x86_64-smp/tests/ampi/megampi'
```

http://charm.cs.illinois.edu/autobuild/old.2018_11_06__01_01/mpi-win-x86_64-smp.txt
http://charm.cs.illinois.edu/autobuild/old.2018_11_10__01_01/mpi-win-x86_64-smp.txt
http://charm.cs.illinois.edu/autobuild/old.2018_11_14__01_01/mpi-win-x86_64-smp.txt

evan-charmworks:
*Original date: 2019-01-11 20:11:00*

---
[...]

PR #2106, opened 2019-05-08
AMPI #2030: Fix MPI_Comm_free

evan-charmworks:
*Original date: 2019-05-07 16:22:17*
*Original PR: https://charm.cs.illinois.edu/gerrit/5164*

---

- Call ckDestroy on proper CProxy type
- Only destroy the current rank to avoid a race condition where the
  array is destroyed before other ranks complete the barrier

Change-Id: I922670fef268afd02c928301450a284624ca0ffe


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -9364,9 +9364,7 @@ AMPI_API_IMPL(int, MPI_Comm_free, MPI_Comm *comm)
     if (*comm != MPI_COMM_WORLD && *comm != MPI_COMM_SELF) {
       ampi* ptr = getAmpiInstance(*comm);
       ptr->barrier();
```
stwhite91: There is no real reason to have a barrier here, and getting rid of it eliminates the occasional failure on linux-smp. I still don't understand the underlying issue with it, but the barrier itself is unnecessary in terms of the MPI semantics. I haven't tested on Windows.

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -9364,9 +9364,7 @@ AMPI_API_IMPL(int, MPI_Comm_free, MPI_Comm *comm)
...
-      if (ptr->getRank() == 0) {
-        CProxy_CkArray(ptr->ckGetArrayID()).ckDestroy();
-      }
+      CProxy_ampi(ptr->ckGetArrayID())[ptr->getRank()].ckDestroy();
```
stwhite91: I think it's nicer / more idiomatic to make this 'ptr->thisProxy[ptr->thisIndex].ckDestroy();'

PR #2107, opened 2019-05-08
Add assertion to CkArray::recvBroadcast deliver loop

evan-charmworks:
*Original date: 2019-05-07 16:24:02*
*Original PR: https://charm.cs.illinois.edu/gerrit/5165*

---

Change-Id: I817eb1f728bb13255e59cd48c367edabc714d460


stwhite91:
Won't this effectively disallow broadcasts to ckDestroy?

evan-charmworks:
If this assertion trips, that means that an out-of-bounds access was about to happen, which implies an implementation defect.

juanjgalvez:
I think a broadcast to ckdestroy won't work correctly right now. I think it could skip some elements.

stwhite91:
We may want to change the documentation on ckDestroy() or open a bug for the bcast issue.
Also it looks like ckio is doing bcasts to ckDestroy right now.

PR #2110, opened 2019-05-08
AMPI: Fix PMPI linking errors

evan-charmworks:
*Original date: 2019-05-08 17:01:48*
*Original PR: https://charm.cs.illinois.edu/gerrit/5182*

---

Change-Id: If77c986d2c3d66f9f9995a78a1099404d169017b

### DIFF
--- a/src/libs/ck-libs/ampi/ampi.h
+++ b/src/libs/ck-libs/ampi/ampi.h
@@ -16,7 +16,7 @@
 #define AMPI
 
 /* Declare the conditions under which AMPI supports PMPI. */
-#if defined(__linux__)
+#if defined __linux__ && !defined __clang__
 # define AMPI_HAVE_PMPI 1
 #else
 # define AMPI_HAVE_PMPI 0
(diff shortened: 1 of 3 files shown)


evan-charmworks:
clang appears to not support PMPI at all, so I disabled AMPI_HAVE_PMPI entirely.

rbuch:
> clang appears to not support PMPI at all, so I disabled AMPI_HAVE_PMPI for it entirely.

I thought that it was standard defined, not compiler specific (see 14.2.1.1 of https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report.pdf, for example). I'm not super confident in my claims, though.

stwhite91:
It is, but we only support it if the compiler supports weak symbols. Supporting it otherwise requires major changes to AMPI, i.e. every MPI_ routine has to be defined in its own file so that they can each be linked in as individual objects depending on if the MPI version is called or the PMPI version

rbuch:
In the interest of getting tests to work again, I'm going to approve and merge this, but I've also created issue #2233 to record that someone should follow up on this (feel free to close that issue immediately if one of the AMPI folks says it's alright).

PR #2138, opened 2019-05-08
cleanup: remove duplicate CmiBarrier() calls

matthiasdiener:
*Original date: 2019-01-18 19:34:40*
*Original PR: https://charm.cs.illinois.edu/gerrit/4909*

---

Change-Id: I5304013c031af6c15b0ec725bbb9982488bdd96a

### DIFF
--- a/src/conv-core/isomalloc.C
+++ b/src/conv-core/isomalloc.C
@@ -2000,6 +2000,7 @@ static void init_ranges(char **argv)
       }   /* end of barrier test */
     } /* end of rank 0 */
     else {
+      // Match the 4 CmiBarrier() calls of rank 0 in the if above
       CmiBarrier();
       CmiBarrier();
       CmiBarrier();
--- a/src/arch/gni/machine.C
+++ b/src/arch/gni/machine.C
@@ -4488,8 +4488,6 @@ void CmiTimerInit(char **argv) {
         }
     } else { /* we don't have a synchronous timer, set our own start time */
         CmiBarrier();
-        CmiBarrier();
-        CmiBarrier();
         clock_gettime(CLOCK_MONOTONIC, &start_ts);
     }
     CmiNodeAllBarrier();          /* for smp */
--- a/src/arch/mpi/machine.C
+++ b/src/arch/mpi/machine.C
@@ -1872,8 +1872,6 @@ void CmiTimerInit(char **argv) {
     } else { /* we don't have a synchronous timer, set our own start time */
 #if ! CMK_MEM_CHECKPOINT && ! CMK_MESSAGE_LOGGING
         CmiBarrier();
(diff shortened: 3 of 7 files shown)


evan-charmworks:
Have we confirmed that these barriers do not correspond to other locations where there is code actually executing between them? For example, the barriers in mpi/machine.C that match up with the isomalloc_sync sequence in isomalloc.C.

evan-charmworks:
To elaborate, take the following example:

`src/arch/mpi/machine.C`:
```
    if (*myNodeID >= num_workpes) {    /* is spare processor */
      if(CmiGetArgFlag(largv,"+isomalloc_sync")){
          MPI_Barrier(charmComm);
          MPI_Barrier(charmComm);
          MPI_Barrier(charmComm);
          MPI_Barrier(charmComm);
      }
```

`src/conv-core/isomalloc.C`:
```
  if (_sync_iso == 1)
  {
// ...
    if (CmiMyRank() == 0 && freeRegion.len > 0u)
    {
      if (CmiBarrier() == -1 && CmiMyPe() == 0)
        CmiAbort("Charm++ Error> +isomalloc_sync requires CmiBarrier() implemented.\n");
      else
      {
// ...

        CmiBarrier();

// ...

        CmiBarrier();

// ...

        CmiBarrier();

// ...
      } /* end of barrier test */
    }   /* end of rank 0 */
    else
    {
      CmiBarrier();
      CmiBarrier();
      CmiBarrier();
      CmiBarrier();
    }
```

Despite being in two separate files, removing the duplicate `MPI_Barrier(charmComm);` calls from `mpi/machine.C` would cause a regression.

Have we confirmed that all duplicate barriers removed in this PR are actually redundant and not necessary for some code elsewhere?

src/conv-core/isomalloc.C
```diff
@@ -875,6 +875,7 @@ static void CmiIsomallocInitExtent()
     }   /* end of rank 0 */
     else
     {
+      // Match the 4 CmiBarrier() calls of rank 0 in the if above
       CmiBarrier();
```
nitbhat, line 879: Why are 4 barrier calls required here?

src/conv-core/isomalloc.C
```diff
@@ -875,6 +875,7 @@ static void CmiIsomallocInitExtent()
     }   /* end of rank 0 */
     else
     {
+      // Match the 4 CmiBarrier() calls of rank 0 in the if above
```
nitbhat, line 878: Why are 4 barrier calls required here?

nitbhat:
I'm guessing we still want to merge this patch? 


The 4 Barriers in isomalloc.C and mpi/machine.C have been both reduced to 1 in the current master.

PR #2141, opened 2019-05-08
AMPI: update ROMIO to 3.2.1

matthiasdiener:
*Original date: 2019-05-02 20:04:52*
*Original PR: https://charm.cs.illinois.edu/gerrit/5151*

---

Change-Id: I9f283a646c9d0fc0a9b1443a2c179347f527be97


evan-charmworks:
I force-pushed this to rebase it over the latest in master, which involved applying the mpio.h.in --> mpio_functions.h transformation.

This PR seems to require `aclocal-1.15` to be installed, which caused the build to fail on Darwin. Seems to work and pass our AMPI tests on Linux.

evan-charmworks:
Tested on Linux and Darwin. The aclocal and double configure problems are resolved.

One remaining issue is this warning on Darwin, for all MPI_* calls made from ROMIO:

```
mpi-io/fortran/deletef.c:128:9: warning: implicit declaration of function 'PMPI_Abort' is invalid in C99 [-Wimplicit-function-declaration]
        MPI_Abort(MPI_COMM_WORLD, 1);
        ^
./adio/include/mpipr.h:14:19: note: expanded from macro 'MPI_Abort'
#define MPI_Abort PMPI_Abort
                  ^
```

matthiasdiener:
> Tested on Linux and Darwin. The aclocal and double configure problems are resolved.
> 
> One remaining issue is this warning on Darwin, for all MPI_* calls made from ROMIO:
> 
> ```
> mpi-io/fortran/deletef.c:128:9: warning: implicit declaration of function 'PMPI_Abort' is invalid in C99 [-Wimplicit-function-declaration]
>         MPI_Abort(MPI_COMM_WORLD, 1);
>         ^
> ./adio/include/mpipr.h:14:19: note: expanded from macro 'MPI_Abort'
> #define MPI_Abort PMPI_Abort
>                   ^
> ```

Does it help to add back the `#if 0` in mpipr.h?

evan-charmworks:
Yes, that fixes the warning.

I think my patch for `MPIR_Ext_cs_yield` needs more work, since it would cause ireads and iwrites to abort. I'm not sure if we should define to be empty, or if it should perform something like `AMPI_Yield`.

matthiasdiener:
> Yes, that fixes the warning.
> 
> I think my patch for `MPIR_Ext_cs_yield` needs more work, since it would cause ireads and iwrites to abort. I'm not sure if we should define it to be empty, or if it should perform something like `AMPI_Yield`.

I think it is ok for `MPIR_Ext_cs_yield` no tot do anything. In fact, it should be defined to nothing in mpioimpl.h:
```
#else /* not ROMIO_INSIDE_MPICH */
// [...]
#define ROMIO_THREAD_CS_YIELD()
```

evan-charmworks:
adio/common/ad_i{read,write}_coll.c call `MPIR_Ext_cs_yield` directly without the use of `ROMIO_THREAD_CS_YIELD`. I'm not sure if this is intentional or another oversight when building without MPICH. We would need to test this either way.

matthiasdiener:
> adio/common/ad_i{read,write}_coll.c call `MPIR_Ext_cs_yield` directly without the use of `ROMIO_THREAD_CS_YIELD`. I'm not sure if this is intentional or another oversight when building without MPICH. We would need to test this either way.

I think it is an oversight. `ROMIO_THREAD_CS_YIELD` appears to be completely unused.

stwhite91:
It would be nice to have this in 6.10

evan-charmworks:
Ready for review

matthiasdiener:
The changes to the Makefile should be reflected in AMPI's CMakeLists.txt, if they aren't already.

evan-charmworks:
CMakeLists.txt updated

matthiasdiener:
I can't approve this, since I opened the PR, but it has my +1.

PR #2142, opened 2019-05-08
Make it possible to build charm++ git without autotools

matthiasdiener:
*Original date: 2019-05-06 21:52:26*
*Original PR: https://charm.cs.illinois.edu/gerrit/5166*

---

Adds the autotools-generated files to the repo, so they don't need to be generated by the user. Also splits off the script to rerun the autotools from the package tarball creation, such that the autotools-generated files can be refreshed without creating the tarball.

Change-Id: Iba0b6afb34069950f54c38cb5ef9d431c6d7adcd


matthiasdiener:
this is ready for review

package-tarball.sh
```diff
@@ -7,54 +7,39 @@ set -v
 set -e
 
 # Set 'release' mode if requested, influences commitid.sh
-export RELEASE="0"
+export RELEASE="0"ß
```
rbuch: Is this an intentional inclusion?

matthiasdiener: No, this was a stray character.

package-tarball.sh
```diff
@@ -7,54 +7,39 @@ set -v
...
         echo Saw $arg
         RELEASE="1"
-    else
+        ;;
+    *)
         echo "Unrecognized argument '$arg'"
         exit 1
-    fi
+        ;;
+    esac
```
rbuch: What's the point of changing this? It seems like the functionality is identical, is that right? Does this help avoid using autotools somehow?

matthiasdiener: No, this isn't a functional change, but I think the argument parsing is more idiomatic. This will also make it easier to add arguments in the future.

matthiasdiener: I'll undo this part of the change.

rbuch:
Can you elaborate on how this avoids the need of using autotools? It seems like it just does the same thing as before, just spread over multiple files now.

matthiasdiener:
Basically, this PR adds the files generated by the autotools to the repo, so they don't need to be generated by the user. It also splits off the script to rerun the autotools from the package tarball creation, such that the autotools-generated files can be refreshed without creating the tarball.

I'll add this to the commit message.

rbuch:
> Basically, this PR adds the files generated by the autotools to the repo, so they don't need to be generated by the user. It also splits off the script to rerun the autotools from the package tarball creation, such that the autotools-generated files can be refreshed without creating the tarball.
> 
> I'll add this to the commit message.

Ah, gotcha. Thanks for the explanation.

matthiasdiener:
> this doesn't seem to include a way to automatically determine when these should be rebuilt. It would be handy to have something automatically warn us that a change has necessitated these be rebuilt.

Yeah, I wouldn't merge this for now. It will be for the most part superseded by the cmake build system.

stwhite91:
Should this be abandoned now that cmake has been merged?

matthiasdiener:
> Should this be abandoned now that cmake has been merged?

We'll still need a (potentially simplified) form of this patch. Even the cmake build currently requires autotools for the integrated hwloc.

matthiasdiener:
> This should also adjust the check for the presence of `autoreconf` and `aclocal` in `build`.

The presence check is only done if `src/scripts/configure` does not exist (which this PR adds). So I guess this doesn't have to be adjusted?

matthiasdiener:
> this doesn't seem to include a way to automatically determine when these should be rebuilt. It would be handy to have something automatically warn us that a change has necessitated these be rebuilt.

I've added a check to determine if configure is out of date, and rerun autoconf in that case.

evan-charmworks:
#2661 eliminated the need for autotools when using the CMake build system. Standard practice for autotools build systems is to require them to be run when building from version control. Only tarballs will come pre-generated (see package-tarball.sh).

PR #2146, opened 2019-05-08
Add UCX machine layer

nitbhat:
*Original date: 2019-05-06 21:01:50*
*Original PR: https://charm.cs.illinois.edu/gerrit/4940*

---

Change-Id: I460fc9965a735e7db34e5fba30bee7e4747646b5

### DIFF
--- a/src/arch/ucx/machine.C
+++ b/src/arch/ucx/machine.C
@@ -1,0 +1,583 @@
+/*
+ * Copyright (c) 2019, Mellanox Technologies. All rights reserved.
+ * See LICENSE in this directory.
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string>
+
+#include "converse.h"
+#include "machine.h"
+#include "pcqueue.h"
+#include "machine-lrts.h"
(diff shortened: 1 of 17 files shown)


src/arch/ucx/machine.C
```diff
@@ -0,0 +1,583 @@
...
+
+    UCX_LOG(3, "status %d len %d, buf %p, req %p, tag %zu\n",
+            status,  info->length, req->msgBuf, request, info->sender_tag);
+
+    if (ucs_unlikely(status == UCS_ERR_CANCELED)) {
+        return;
+    }
+
+#if CMK_ONESIDED_IMPL
+    if (info->sender_tag & UCX_RMA_TAG_PUT) {
```
nitbhat: Inside LrtsIssueRget, when a source buffer is unregistered, it is sent back to the source with the tag 'UCX_RMA_TAG_PUT'. However, I see that on receiving this message, you directly perform a PUT using UcxRmaOp. Isn't it required to register the source buffer using ucp_mem_map before issuing the PUT? 

(Same question for receiving UCX_RMA_TAG_GET and performing a GET).

matthiasdiener:
Merged as #2249.

PR #2160, opened 2019-05-08
ZC Bcast API: Fix hang in smp mode because of unsafe thread access

nitbhat:
*Original date: 2019-05-08 14:35:35*
*Original PR: https://charm.cs.illinois.edu/gerrit/5175*

---

Previously, the variable 'counter', which keeps a track of the number of child
Rgets completed was being incremented by both the worker thread (after
a memcpy transfer) and the comm. thread (after a CMA or an RDMA transfer).
This caused an occasional hang because the condition (counter == numChildren)
was not satisfied. This patch makes 'counter' an atomic variable for SMP
modes to avoid the unsafe access and hence fixes the hang.

Change-Id: I8e77327da3682c86b3f952ba27957cf2c363a7d3


nitbhat:
> Check the indentation of the four return statements.
Done.

evan-charmworks:
Is the addition of the changes from `srcSize` to `destSize` intentional?

nitbhat:
> Is the addition of the changes from `srcSize` to `destSize` intentional?

Looks like some other changes from another patch that I was planning to submit crept into this patch. I have removed those now. Thanks for catching this.

evan-charmworks:
It looks like commit 9664648b7c228960023914a572e69b1b96ebb05c is part of this branch now and since I authored it, GitHub is not letting me review this PR.

evan-charmworks:
Nevermind, figured out how to review it. But should that commit be part of this branch?

nitbhat:
> Nevermind, figured out how to review it. But should that commit be part of this branch?

I had the other commit to have the automatic verification pass. I've removed them now. I'll rebase this branch on top of the current master.

Issue #480, opened 2014-04-22
Calling 'migrateMe' in SDAG serial block accesses members of deallocated chare


pplimport:
*Original author: Robert Steinke*
*Original issue: https://charm.cs.illinois.edu/redmine/issues/480*

---
I have some Charm++ code in ADHydro-crash.tar.gz that crashes when run on more than one processor.  It doesn't crash if it is only run on one processor, or if I remove the migrateMe call at the end of mesh_element.cpp.  It also doesn't crash if I rearrange the code to remove a while loop from the structured dagger code in ADHydro-nocrash.tar.gz.  So it appears to be related to structured dagger and migration.

[ADHydro-crash.tar.gz](https://github.com/UIUC-PPL/charm/files/3536431/ADHydro-crash.tar.gz)
[ADHydro-crash2-output.txt](https://github.com/UIUC-PPL/charm/files/3536432/ADHydro-crash2-output.txt)
[ADHydro-crash2.tar.gz](https://github.com/UIUC-PPL/charm/files/3536433/ADHydro-crash2.tar.gz)
[ADHydro-nocrash.tar.gz](https://github.com/UIUC-PPL/charm/files/3536434/ADHydro-nocrash.tar.gz)

PhilMiller:
*Original date: 2014-04-22 22:38:41*

---
Targeting 6.6.0 since we're in the release preparation stages for that, and SDAG's implementation has been overhauled since the 6.5 release. I'll try to triage the problem this evening.

PhilMiller:
*Original date: 2014-04-22 23:29:29*

---
OK, so this is a known limitation ("documented, but rather poorly":http://charm.cs.illinois.edu/manuals/html/charm++/7.html#SECTION01640000000000000000) on the use of the `migrateMe` function. When it's called, the element calling it is packed up, destructed, and de-allocated on its original host PE in-line, immediately, at the point of the call. When called inside SDAG code, that means that the SDAG control flow code subsequent to the call will be operating on an object that's no longer there.

If you use migration triggered by `AtSync` or periodic load balancing, this won't arise, since the infrastructure-controlled migration paths operate outside execution of the subject chares' entry methods. Since that generally matches live application usage patterns, while the demonstration here is commented as 'simulating' load balancing, I'd recommend working with that approach.

As a workaround, if you _really_ want to use `migrateMe()` specifically, the way to do it would be to stick the call in a stand-alone entry method outside of any SDAG 
```
// in mesh_element.ci
[...]

PR #2164, opened 2019-05-08
Bug #480: Make migrateMe() more forgiving by executing it asynchronously

pplimport:
*Original author: Gerrit Code Review*
*Original date: 2014-07-14 19:07:58*
*Original PR: https://charm.cs.illinois.edu/gerrit/195*

---

The API routine `void migrateMe(int toPe)' has historically been very
unforgiving, in that any member access after it's called could result in
undefined behavior (reading/writing members of an object that's been
deleted). This is particularly pernicious in combination with SDAG code, in
which the user may not have written any code after the call, but the generated
code has to touch member fields to implement subsequent control flow.

There's nothing really fundamental about running this method perfectly
synchronously - the important bit is that the object migrate as ordered with no
further direction or runtime system mediation. So, add a bare entry method
ckEmigrate() that does only the inline emigration call, and nothing else, and
turn migrateMe() into an asynchronous invocation of that.

To add this everywhere requires some C++11 trickery in the present
implementation. Alternative implementations that are similarly clean are
welcome.

Change-Id: I7a200b7232bf39c5619fe2bc7f62c5097dc7ff03


juanjgalvez:
For places inside the runtime that call migrateMe, like lb framework and maybe others like ckevacuation, I assume we still want to call migrateMe directly and not schedule it as an entry method. It doesn't pose a problem for lb framework, and if you are migrating many objects it is more efficient to call directly.

Although I'm not sure which is the exact migrate function that gets called from the lb framework.

evan-charmworks:
After rebasing this patch, all tests complete with my assertion in `CkArray::recvBroadcast` from #2231 in place. We should still evaluate whether any uses of `migrateMe` in runtime code should be changed.

evan-charmworks:
I reviewed all calls to `migrateMe` in the runtime found no need to change any of them. I was able to replace the `migrateDelayed` entry method in TCharm with the general `ckEmigrate` added here.

juanjgalvez:
Do you know what the load balancing framework calls to make chares migrate? I spent some time trying to track it down but didn't get to the bottom of it. I think it might be using function pointers which made it hard to track.

evan-charmworks:
A pseudo-backtrace would look like

```
void CkLocMgr::emigrate(CkLocRec *, int)
void CkLocRec::migrateMe(int)
inline void CkMigratable::ckMigrate(int)
entry void ArrayElement::ckEmigrate(int)
inline void CBase_*::migrateMe(int)
```

It looks like `CkLocMgr::emigrate` is the highest-level point that handles the details of migration.

juanjgalvez:
Ok, I just saw that when the lb framework wants to migrate, the call leads to `CkLocRec::migrateMe` without going through `ArrayElement::migrateMe`.

Another concern is if there are applications making use of migrateMe, and if the patch could break them in some obscure way. I assume most applications that migrate objects will use the lb framework.

evan-charmworks:
I don't see any function calls to `ArrayElement::migrateMe` in NAMD, ChaNGa, or OpenAtom.

ericjbohm:
NAMD migration is an application level thing that doesn’t use ArrayElement at all.

PR #2188, opened 2019-05-08
64-bit message sizes for Charm/Converse

evan-charmworks:
*Original date: 2018-03-29 21:03:42*
*Original PR: https://charm.cs.illinois.edu/gerrit/2395*

---

Co-authored-by: Evan Ramos <evan@hpccharm.com>

Change-Id: Iebadb024639a71ec04d2cb3debd3743e69120bb7


evan-charmworks:
Draft because, at least, non-netlrts machine layers need implementation, and the documentation also needs updating.

evan-charmworks:
- CCS and other components also need updating for size_t
- NetLRTS-TCP (and therefore the Charm4Py CI) are broken because TransmitImplicitDgram inappropriately generates its message header by using DgramHeaderMake, which fails because this is not size_t-safe.

evan-charmworks:
In addition to being out-of-date and the reasons listed above, I no longer sure whether changing the message size field at the network level to 64 bits is a good idea, partially because it takes up more space in the envelope/header, but mainly because some system APIs like Cross-Memory Attach only take 32-bit sizes. Higher level message sends such as at the Charm layer could/should still support 64-bit sizes through message pipelining.

PR #2189, opened 2019-05-08
Clean up memory used for tracing in TCharm when exiting

stwhite91:
*Original date: 2017-04-28 19:12:13*
*Original PR: https://charm.cs.illinois.edu/gerrit/2463*

---

Change-Id: I9728c5ac0b58c6e1efac50e8b24d96b52cdfaf64


stwhite91:
This isn't the proper way to do this, we need to do a reduction on each node to ensure all users of the TCharm tracing stuff are finished using it before deleting it

PR #2190, opened 2019-05-08
AMPI: implement (I)Ssend's using a synchronous rendezvous protocol

stwhite91:
*Original date: 2019-05-03 17:24:05*
*Original PR: https://charm.cs.illinois.edu/gerrit/2541*

---

Instead of sending the message payload eagerly and then getting an ack
when the recv'er actually matches it, now we use a rendezvous protocol,
only sending the payload after it has been matched.

The initial sync message (to ampi::genericSync) has no payload and its
refnum is the index of the sender's request. When the sync message is
matched by the recv'er, we delete that placeholder msg and send a
message back to the sender (to ampi::ssendAck) to get the payload back
in a normal AmpiMsg (via ampi::generic or ampi::genericRdma).

Change-Id: I6a8db194f63321a2247f707042d624d39fa3bef8

### DIFF
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -3380,8 +3380,8 @@ void ampi::processGathervMsg(CkReductionMsg *msg, void* buf, MPI_Datatype type,
   delete [] results;
 }
 
-ampi* ampi::blockOnIReq(void* buf, int count, MPI_Datatype type, int src,
-                        int tag, MPI_Comm comm, MPI_Status* sts) noexcept
+CMI_WARN_UNUSED_RESULT ampi* ampi::blockOnIReq(void* buf, int count, MPI_Datatype type, int src,
+                                               int tag, MPI_Comm comm, MPI_Status* sts) noexcept
 {
   MPI_Request request = postReq(parent->reqPool.newReq<IReq>(buf, count, type, src, tag, comm, getDDT(),
                                                              AMPI_REQ_BLOCKED));
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -2700,22 +2700,72 @@ CMI_WARN_UNUSED_RESULT ampi* ampi::static_blockOnColl(ampi *dis) noexcept {
   return dis;
 }
 
-void ampi::ssend_ack(int sreq_idx) noexcept {
-  if (sreq_idx == 1)
-    thread->resume();           // MPI_Ssend
-  else {
-    sreq_idx -= 2;              // start from 2
-    AmpiRequestList& reqs = getReqs();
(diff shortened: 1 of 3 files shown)


evan-charmworks:
Needs a force-push to remove leading commits

evan-charmworks:
It looks like something went wrong here. It might fix it to run `git checkout 3fb8d79fe378c87458f994611f3a621a0c2a05eb && git rebase --onto master HEAD~1` and then force pushing again

stwhite91:
That seems to have worked but now I have this random commit in there for some reason "AMPI: make multiple worlds/modules support a build option"

evan-charmworks:
You probably need to `git checkout master && git reset --hard origin/master` to reset your master branch, then `git checkout 7839adc7f58ce255fd99654f1740b973722cdef0 && git rebase --onto origin/master HEAD~1` to for sure isolate just the one commit

evan-charmworks:
I think this commit's content implicitly depends on the pointer refreshing patch. Not sure if it can be avoided

stwhite91:
Yeah, the pointer one seems to be ready for merge, so I'll just update this one once that gets in.

src/libs/ck-libs/ampi/ampiimpl.h
```diff
@@ -1636,15 +1644,16 @@ class AmpiMsg final : public CMessage_AmpiMsg {
...
+  inline void setSsendReq(int s) noexcept { ssendReq = s; }
   inline void setSeq(CMK_REFNUM_TYPE s) noexcept { CkAssert(s >= 0); UsrToEnv(this)->setRef(s); }
   inline void setSrcRank(int sr) noexcept { srcRank = sr; }
   inline void setLength(int l) noexcept { length = l; }
   inline void setTag(int t) noexcept { tag = t; }
   inline void setComm(MPI_Comm c) noexcept { comm = c; }
-  inline CMK_REFNUM_TYPE getSeq() const noexcept { return UsrToEnv(this)->getRef(); }
-  inline int getSsendReq() const noexcept { return ssendReq; }
-  inline int getSeqIdx() const noexcept {
+  inline CMK_REFNUM_TYPE getSeq(void) const noexcept { return UsrToEnv(this)->getRef(); }
```
matthiasdiener: no need for the added `(void)`s

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -2793,7 +2843,7 @@ void AmpiRequestList::free(AmpiRequestPool &reqPool, int idx, CkDDT *ddt) noexce
   startIdx = std::min(idx, startIdx);
 }
 
-void ampi::inorder(AmpiMsg* msg) noexcept
+bool ampi::inorder(AmpiMsg* msg) noexcept
```
matthiasdiener: Could add a comment here on the meaning of the returned value.

PR #2195, opened 2019-05-08
AMPI: use zero-copy direct API to avoid intermedate copies

stwhite91:
*Original date: 2019-05-03 17:45:18*
*Original PR: https://charm.cs.illinois.edu/gerrit/3080*

---

For messages sent within the same PE or process, we pack an
AmpiNcpyShmBuffer object into the AmpiMsg data. This allows direct
handling of DDTs in a single message, and avoids the cost of memory
pinning when it is unnecessary. We still have to handle the case where
the recv'er turns out to not be in the same process, so the recv'er
then requests a put() of the message payload.

For messages sent across processes, we pack a CkNcpyBuffer object into
the AmpiMsg data and use that to do a get() from the recv'er.

Change-Id: I9e8126bc8bf321b4a5ab64e08b3283e448fec4c0


src/libs/ck-libs/ampi/ampiimpl.h
```diff
@@ -2497,8 +2594,8 @@ class ampi final : public CBase_ampi {
   ampi() noexcept;
   ampi(CkArrayID parent_,const ampiCommStruct &s) noexcept;
   ampi(CkMigrateMessage *msg) noexcept;
-  void ckJustMigrated() noexcept;
-  void ckJustRestored() noexcept;
+  void ckJustMigrated(void) noexcept;
+  void ckJustRestored(void) noexcept;
```
evan-charmworks: Don't need these voids

stwhite91: Done

stwhite91:
This patches fixes the failures in tests/ampi/datatype that have been silently happening for a couple months

stwhite91:
Needs another review

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -6092,7 +6195,7 @@ AMPI_API_IMPL(int, MPI_Wait, MPI_Request *request, MPI_Status *sts)
              *request, reqs[*request], (int)(reqs[*request]->tag));
   AMPI_DEBUG("MPI_Wait: request=%d, reqs.size=%d, &reqs=%d\n",
              *request, reqs.size(), reqs);
```
matthiasdiener: `reqs.size()` probably won't compile (see my previous patch to restore AMPI_DEBUG)

evan-charmworks:
Still has a conflict

src/libs/ck-libs/ampi/ampi.C
stwhite91: Fixed

PR #2197, opened 2019-05-08
AMPI: maintain a free list of CkNcpyBuffer objects on each PE

stwhite91:
*Original date: 2019-05-03 19:19:28*
*Original PR: https://charm.cs.illinois.edu/gerrit/3946*

---

Change-Id: I6907e15381eae157bdf767665284e48d27e958de

### DIFF
--- a/src/conv-core/converse.h
+++ b/src/conv-core/converse.h
@@ -131,6 +131,15 @@
 #  endif
 # endif
 
+// must be placed before return type and at both declaration and definition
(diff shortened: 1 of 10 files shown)


stwhite91:
This is no longer applicable. The finalized zero copy API doesn't require the application to dynamically allocate CkNcpyBuffers. Instead we may want to keep freelists or pools of CkNcpyBuffers inside Charm/Converse, but that would be an entirely separate effort.

PR #2213, opened 2019-05-08
Implement malloc_info to enhance mall_info

pplimport:
*Original author: Shaoqin Lu*
*Original date: 2018-04-16 17:18:29*
*Original PR: https://charm.cs.illinois.edu/gerrit/3968*

---

mall_info does not support 64 bit sized memory reporting.
malloc_info is used as a fall through after 32 bit is overflowed

malloc_info is streamed in a xml format. program parses the
string to extract mmap and total usage from the summary section
and reports the difference compared to the first malloc_info run.

Change-Id: I7884e23716581cc7589fe1bac0e445ce2521b0af


matthiasdiener:
This patch is probably not worth keeping.

PR #2215, opened 2019-05-08
Fixed CmiNodeBarrier() for PAPI Initialization:

pplimport:
*Original author: Steven Qiou*
*Original date: 2018-03-13 00:47:53*
*Original PR: https://charm.cs.illinois.edu/gerrit/3796*

---

Initial implementation of CmiNodeBarrierCount() assumed a NodeAllBarrier
will complete before a NodeBarrier is called (and vice versa), however
the workers in initPAPI() need to call NodeBarrier before they can complete
the NodeAllBarrier previously called by comm threads. Original
CmiNodeBarrierCount could not handle this situation with one set of
synchronization variables, so I added a second copy of everything, allowing
CmiNodeBarrierCount to handle both barrier types at the same time.

The pairs of elements are stored as the first and last elements of arrays
that are large enough that the two elements will exist onseparate cache
lines as to avoid false sharing.

Change-Id: I3a4c154adb391f8e2361a53fa8411c749c720fb5

### DIFF
--- a/doc/projections/manual.tex
+++ b/doc/projections/manual.tex
@@ -1059,10 +1059,11 @@ proportional to the number of processors to be displayed.
 
 This view is enabled when \charmpp{} is compiled with the {\tt papi}
 option, which enables the collection of performance counter data using
-PAPI. Currently, this collects {\tt PAPI\_L1\_TCM}, {\tt PAPI\_L1\_TCA},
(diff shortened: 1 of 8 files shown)


rbuch:
Superseded by #2359. This change contains an implementation that is hardened against false sharing, so it may be worth resuscitating if performance is a problem down the line (but it shouldn't be considering that the implementation uses condition variables, not spin locks, so cache ping-ponging probably won't occur).

PR #2226, opened 2019-05-08
fix the bug of using CkMyRank before ConverseInit

pplimport:
*Original author: YanhuaSun*
*Original date: 2014-08-28 22:51:13*
*Original PR: https://charm.cs.illinois.edu/gerrit/362*

---

Change-Id: I82423c1ec95220a7365ff7bb4c1de2d93a2d1c43

### DIFF
--- a/src/ck-core/register.h
+++ b/src/ck-core/register.h
@@ -269,11 +269,6 @@ public:
 	/// Add a heap-allocated registration record,
 	///  returning the index used.
 	int add(T *t) {
-#if CMK_ERROR_CHECKING
-		/* Make sure registrations only happen from rank 0: */
-		if (CkMyRank()!=0)
-			CkAbort("Can only do registrations from rank 0 processors");
-#endif
 		vec.push_back(t);
 		return vec.size()-1;
 	}


matthiasdiener:
Per the discussion in the gerrit log, its probably not relevant anymore.

---

Charm Automation
Patch Set 1:

Build Started http://poise:8080/job/charm-commit-triggered/286/

Aug 28, 2014 17:51 
Eric Bohm
added to REVIEWER: 
Eric Bohm
Aug 28, 2014 17:58 
Eric Bohm
Patch Set 1:

The original code was written intending to avoid some bug associated with calling that code off of rank 0.  Was that original code introduced erroneously?  If not, then how is the error it is guarding against being avoided?

Aug 28, 2014 17:58 
Charm Automation
added to REVIEWER: 
Charm Automation
Aug 28, 2014 18:09 
Charm Automation Verified +1
Patch Set 1: Verified+1

Build Successful 

http://poise:8080/job/charm-commit-triggered/286/ : SUCCESS

Aug 28, 2014 18:09 
Yanhua sun
Patch Set 1:

The purpose of avoiding not rank 0 calling that is right. However, using CkMyRank is improper. In some cases, this function is called before rank is initialized. Calling CkMyRank gives wrong value.

Aug 28, 2014 22:59 
Nikhil Jain
Patch Set 1:

Who is calling it before ranks are initialized? This data structure is primarily for Charm++ core, while ranks are assigned at startup by Converse, so we may want to revisit this use case.

Sep 16, 2014 10:12 
Yanhua sun
Patch Set 1:

It is triggered by a readonly variable registration

Sep 16, 2014 12:59 
Nikhil Jain
Patch Set 1:

But that will most likely be after the Converse init is done and during execution of charm init. Why would rank be not defined at that time?

Sep 16, 2014 14:15 
Yanhua sun
Patch Set 1:

It is in megatest. If you add some print in CmiMyRank(), you will find it is printed before converseInit. Actually it is even before main .

Sep 17, 2014 11:28 
Yanhua sun
Patch Set 1:

To be specific, it fails because readonly<CkGroupID> nodering_thisgroup;.  The constructor is called before main

Sep 17, 2014 11:39 
Nikhil Jain
added to REVIEWER: 
Nikhil Jain
Sep 17, 2014 11:49 
Nikhil Jain
Patch Set 1:

Ah, this is because of megatest. Seems like megatest is using a really odd (old?) method of deploying readonly. 

Eric B, Phil, (since you may have seen it before) is this reasoning correct? Should megatest be updated to avoid this?

Sep 17, 2014 11:49 
Yanhua sun
Patch Set 1:

Since charm still supports readonly<>, only modifying megatest does not solve the problem. We do not know whether other programs use this. Also, we can not prevent it is used.

Sep 17, 2014 12:04 
Phil Miller
added to REVIEWER: 
Phil Miller
Sep 17, 2014 12:08 
Phil Miller
Patch Set 1:

If readonly<> is the root cause of this, we can easily enough look around for uses in real applications. It's *never* been documented in our manual, AFAIK. If it really does need CkMyRank in the static constructor context, then it's critically flawed, and should be deprecated and removed.

PR #2234, opened 2019-05-09
charm4py patches

juanjgalvez:


juanjgalvez:
This is ready for review.

Note that **all** of the code in these commits is inside `#if CMK_CHARMPY` blocks

juanjgalvez:
> Looks good, I just hope all the #ifs and #endifs are in the right place.

Yes, they must be because Charm++ passes verification and I also tested charm4py separately and it also works.

PR #2162, opened 2019-05-08
ZC Post API: Allow users to modify size in Post Entry Method

nitbhat:
*Original date: 2019-05-08 18:49:06*
*Original PR: https://charm.cs.illinois.edu/gerrit/5183*

---

Change-Id: Ie50c45c92e32a46a1ea330e12ff2463ae1d14b02

PR #2236, opened 2019-05-09
ZC Direct API: Use min(srcSize, destSize) as the size for get/put

nitbhat:
Change-Id: Id6af431bef5ab3dc302574fa0e3ce6b68d81f1de


stwhite91:
I think we should abort with a clear error message if a user tries to do a get() or put() with a size larger than that exposed by the target of the operation. Do we already have that?

nitbhat:
> I think we should abort with a clear error message if a user tries to do a get() or put() with a size larger than that exposed by the target of the operation. Do we already have that?

In the Direct API, the user cannot modify the 'size'. The only restriction is that the source buffer should be smaller than the destination buffer. (target buffer). 
For the Direct API, we do have checks in CkNcpyBuffer::get and CkNcpyBuffer::put to ensure that (srcSize <= destSize). I'll add relevant error messages to make it more user friendly. 

So, basically, for the Direct API, get and put with srcSize <= destSize are allowed. (and srcSize > destSize is not allowed). 


For the EM Post API (which performs a get), the 'size' variable refers to the amount of data to be read from the source buffer. In this case, I would be throwing an error if 'size' is larger than 'srcSize'. (Because it is invalid to read more data than allowed by the sender). 

For the EM Post API, I am going to rebase the patch (https://github.com/UIUC-PPL/charm/pull/2162) on top of this one and add error messages if the user modifies the 'size' variable in the Post Entry Method to a value larger than the source specified size in CkSendBuffer.

PR #2238, opened 2019-05-10
Use independent ccache per Travis build

rbuch:
According to https://docs.travis-ci.com/user/caching/#caches-and-build-matrices, all of the linux builds will point to the same cache, which will then become corrupt. Adding unique environment variables to each build should make them use independent caches.


rbuch:
| Build | Original Time | New Time |
| -- | -- | -- |
| NetLRTS-Linux | 16 min 57 sec | 8 min 1 sec |
| NetLRTS-Linux SMP | 17 min 13 sec | 6 min 11 sec |
| MPI-Linux SMP | 13 min 23 sec | 12 min 48 sec |
| NetLRTS-Darwin | 16 min 4 sec | 16 min 45 sec |

I'm not terrible sure why this doesn't seem to be having much of any effect on the MPI and Darwin builds.

PR #2246, opened 2019-05-15
Bug #2241: Delete .gitreview

rbuch:
This configuration file was used for Gerrit, which is not longer being used in favor of Github.


rbuch:
Closes #2241

PR #2249, opened 2019-05-16
Add UCX machine layer

brminich:
This one replaces #2146


src/arch/ucx/machine.C
```diff
@@ -0,0 +1,583 @@
...
+#define UCX_MSG_PROBE_THRESH      32768
+#define UCX_MSG_NUM_RX_REQS       64
+#define UCX_MSG_NUM_RX_REQS_MAX   1024
+#define UCX_TAG_MSG_BITS          4
+#define UCX_TAG_RMA_BITS          4
+#define UCX_MSG_TAG_DIRECT        UCS_BIT(0)
+#define UCX_MSG_TAG_PROBE         UCS_BIT(1)
+#define UCX_RMA_TAG_GET           UCS_BIT(UCX_TAG_MSG_BITS + 1)
+#define UCX_RMA_TAG_PUT           UCS_BIT(UCX_TAG_MSG_BITS + 2)
+#define UCX_RMA_TAG_ACK           UCS_BIT(UCX_TAG_MSG_BITS + 3)
```
nitbhat: We can probably call UCX_RMA_TAG_ACK as UCX_RMA_TAG_DEREG_AND_ACK.

src/arch/ucx/machine.C
```diff
@@ -330,6 +330,20 @@ static void UcxRxReqCompleted(void *request, ucs_status_t status,
         resetNcpyOpInfoPointers(ncpyOpInfo);
         ncpyOpInfo->freeMe = CMK_FREE_NCPYOPINFO; // It's a message, not a real ncpy Obj
         UcxRmaOp(ncpyOpInfo, UCX_RMA_OP_GET);
+    } else if (info->sender_tag & UCX_RMA_TAG_GET) {
```
nitbhat: This should be UCX_RMA_TAG_DEREG_AND_ACK.

nitbhat: This should be modified to ( info->sender_tag & UCX_RMA_TAG_DEREG_AND_ACK)

src/arch/ucx/machine-onesided.C
```diff
@@ -70,7 +70,7 @@ void LrtsDeregisterMem(const void *ptr, void *info, int pe, unsigned short int m
 
     UCX_LOG(4, " %p, PE %d, info %p, memh %d", ptr, pe, ucxInfo, ucxInfo->memh);
```
nitbhat: Currently, the Charm++ layer assumes that both source and destination memory is registered and invokes CmiDeregisterMem(which calls LrtsDeregisterMem). 

However, from our conversation yesterday, I found out that the UCX layer only requires peer remote memory to be registered. (and not local memory).

For that reason, /examples/charm++/zerocopy/entry_method_api/unreg/simpleZeroCopy is failing with a segfault inside this method. (Because it is trying to de-register a region that was not registered). 

Is there a way to find out with the UCX layer if the memory is registered? And only then de-register? (to bypass that de-registration?). 

If not, I can modify the LrtsDeregisterMem API to have the charm layer send a flag and not de-register in specific cases.

brminich: The flag option sounds optimal. UCX does not provide such an API, so I'd need to hash all addresses in UCX machine layer to be able to make a decision.

nitbhat:
In the process of reviewing, I built LIBS on the thor cluster and I am getting the following error when I run programs:
```
[nitinb@thor001 tmp]$ cd /global/home/users/nitinb/charm/ucx-linux-x86_64-debug/tests/util/
[nitinb@thor001 util]$ ls
charmrun  check  check.C  check.o  headerpad.C  headerpad.ci  headerpad.h  Makefile
[nitinb@thor001 util]$ ls
charmrun  check  check.C  check.o  headerpad.C  headerpad.ci  headerpad.h  Makefile
[nitinb@thor001 util]$ make test
../../bin/testrun  +p4 ./check 4

Running on 4 processors:  ./check 4
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./check 4
------- Partition 0 Processor 0 Exiting: Called CmiAbort ------
Reason: UCX: UcxInitEps: runtime_kvs_put error failed: 5
------- Partition 0 Processor 0 Exiting: Called CmiAbort ------
Reason: UCX: UcxInitEps: runtime_kvs_put error failed: 5
[0] Stack Traceback:
  [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0xb3  [0x4487cc]
  [0:1] CmiGetNonLocal+0  [0x448805]
[0] Stack Traceback:
  [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0xb3  [0x4487cc]
  [0:1] CmiGetNonLocal+0  [0x448805]
  [0:2]   [0x44cc99]
  [0:3] _Z8LrtsInitPiPPPcS_S_+0x561  [0x44e0a5]
  [0:4] ConverseInit+0x2d0  [0x4482e1]
  [0:5] main+0x2d  [0x445dc0]
  [0:6] __libc_start_main+0xf5  [0x2aaaaca6e3d5]
  [0:7]   [0x4459a9]
  [0:2]   [0x44cc99]
  [0:3] _Z8LrtsInitPiPPPcS_S_+0x561  [0x44e0a5]
  [0:4] ConverseInit+0x2d0  [0x4482e1]
  [0:5] main+0x2d  [0x445dc0]
  [0:6] __libc_start_main+0xf5  [0x2aaaaca6e3d5]
  [0:7]   [0x4459a9]
```

I have the following modules:
```
[nitinb@login01 charm]$ module list
Currently Loaded Modulefiles:
  1) intel/2018.1.163   2) mkl/2018.1.163     3) hpcx/2.4.0
```

And I built charm with the following command:
`./build LIBS ucx-linux-x86_64 --enable-error-checking --suffix=debug -j16 -g -O0`

brminich:
can u please try using impi mpirun? By default UCX will not work with OMPI, as we discussed

nitbhat:
Okay, understood. 

However, on Thor cluster, since HPCX module is built with OMPI, it conflicts with IMPI (and vice versa). Is there any HPCX module that doesn't depend on OMPI, so that I can have HPCX and IMPI loaded in my environment? (Or is there any other way of doing this).

brminich:
If you want to use UCX from hpcx module you can do the following when charm is built:
$ module unload hpcx/2.4.0
$ module load impi/2018.1.163

// Now you have impi mpirun
$ which mpirun
/global/software/centos-7/modules/langs/intel/2018.1.163/impi/2018.1.163/bin64/mpirun

nitbhat:
> If you want to use UCX from hpcx module you can do the following when charm is built:
> $ module unload hpcx/2.4.0
> $ module load impi/2018.1.163
> 
> // Now you have impi mpirun
> $ which mpirun
> /global/software/centos-7/modules/langs/intel/2018.1.163/impi/2018.1.163/bin64/mpirun

Yes, I did that. However, after I unload hpcx/2.4.0, while running the executable, I get an error because some .so from hpcx is missing. 

```
Running on 4 processors:  ./check 4
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./check 4
./check: error while loading shared libraries: libucp.so.0: cannot open shared object file: No such file or directory
```

In any case, I tested simple pmi on another machine (iforge) and both the non-smp and smp modes are running smoothly.

brminich:
ok, let me double check thor once it is available

nitbhat:
Hi @brminich, 

I am almost done with the integration of the ZC API for the UCX layer. However, I have hit a roadblock for the “unreg” mode. The de-registration issue that we faced earlier is solved. The new issue that I’m seeing for the “unreg” mode is when a PUT operation is performed instead of a GET. From the UCX API semantics, I see that the “ucp_put_nb” call invokes the ucx callback when the local buffer is ready to be reused, but does not guarantee remote completion. However, in the higher Charm++ layer, we assume that the PUT operation guarantees remote completion. Is there a way to guarantee remote completion for a UCX PUT call? 

I was seeing this crash in /examples/charm++/zerocopy/entry_method_api/unreg/simpleZeroCopy because of the receiver buffer not having the data that it was supposed to have. My guess was that the “ucp_put_nb” call triggers the local callback even when the remote operation is incomplete i.e. the data has not been received by the receiver. 

Thanks,
Nitin

brminich:
Hi @nitbhat, 
We do not have corresponding API now. Can we overcome this situation by some other way? Or remote completion is required in any case?

nitbhat:
Oh, okay. So, for a PUT operation, is there no way to determine when the data has been received in the receiver buffer? (and the only notification is when it is safe to reuse the source buffer?)

One of the ways in which we can overcome the situation is by not doing the PUT, but sending back the data to the destination process and doing a GET, so that we know when the data transfer has been complete.

brminich:
We can try to use flush or performing fence(); get(); after the put. But probably sending request for registration and performing get would be more efficient.

Issue #2059, opened 2019-03-15
Inconsistent CPU affinity options for running SMP programs


nitbhat:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2059*

---
It seems to me that on different machines (and different allocations/partitions), to get Charm++ SMP threads to be pinned to cores, one needs to provide different options.

And in some cases, it is not possible to get the desired pinning. (As in the case of Golub)

I was attempting to run a simple pingpong benchmark on 2 nodes. ( 2 cores on each node, 1 core for the worker thread and another for the comm thread).
Following are the examples of my run commands on different machines. 

*Example 1 - Golubs - verbs-linux-x86_64-smp build:
Allocated a job with 2 nodes, 2 cores each*
```
[nbhat4`golub213 p2pPingpong]$ ./charmrun +p2 ./megaZCPingpong ++ppn 1 +pemap 0 +commap 1

Running on 2 processors:  ./megaZCPingpong +ppn 1 +pemap 0 +commap 1
./charmrun: line 99: mvapich2-start-mpd: command not found
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_SINGLE (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 2 processes, 1 worker threads (PEs) + 1 comm threads per process, 2 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: v6.9.0-136-g37f84a4ac
Warning> Randomization of virtual memory (ASLR) is turned on in the kernel, thread migration may not work! Run 'echo 0 > /proc/sys/kernel/randomize_va_space' as root to disable it, or try running with '+isomalloc_sync'.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled.
Charm++> cpuaffinity PE-core map : 0
Charm++> Running on 1 hosts (1 sockets x 2 cores x 1 PUs = 20-way SMP)
Charm++> Warning: Internally-determined PU count does not match hwloc's result!
Charm++> cpu topology info is gathered in 0.044 seconds.
*WARNING: Multiple PEs assigned to same core, recommend adjusting processor affinity or passing +CmiSleepOnIdle to reduce interference.*
[...]

Issue #1907, opened 2018-05-11
Reimplement CmiNumCores et al using hwloc


evan-charmworks:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1907*

---
With hwloc in place, there is no need to maintain our custom querying of hardware PU counts.

stwhite91:
*Original date: 2018-10-02 17:26:51*

---
The current API also doesn't expose information about sockets or anything much beyond NumCores, NumPhysicalNodes, and NumPesPerPhysicalNode

ericjbohm:
*Original date: 2018-10-02 18:15:10*

---
The basic functionality of the current API needs to be maintained.  Being able to get basic topology information with a single function call is incredibly convenient.  HWLOC is much richer for the within node information, but requires the user to do a lot more work to obtain the same information.  

The top level of the HWLOC ontology is "machine".  We can make collections of those, but any questions about the collection would be answered by us, not HWLOC.

That said, the implementation of the within host aspects of the API using HWLOC would help clean up some redundant code and exposing more of the functionality of HWLOC would be welcome.

evan-charmworks:
#2250 was the first step of this.

PR #2250, opened 2019-05-16
Cleanup #1907: Reimplement CmiNumCores using hwloc

evan-charmworks:
See #1907, #2059


stwhite91:
We should add more direct wrappers for querying hwloc's hardware entities, since it's unfortunate to have a function CmiNumCores() that returns hwloc's PU count when hwloc has a different measure of cores. This patch is good on its own though.

PR #2254, opened 2019-05-20
doc: add charmci syntax highlighting

matthiasdiener:


juanjgalvez:
What happens if someone is building the manual with an older version that doesn't recognize charmci? Does it fall back to c++ or no syntax highlighting?

matthiasdiener:
It will print a warning and not do any syntax highlighting for that code block. We could add a version check to the makefile if that is of interest.

PR #2255, opened 2019-05-20
ZC API: Refactor multiple converse handlers into one Ncpy Ack Handler

nitbhat:
Additionally, move the handler code applicable to CMK_ONESIDED_IMPL
from conv-core/conv-rdma.C to ck-core/ckrdma.C.


src/ck-core/ckrdma.C
```diff
@@ -1742,5 +1744,33 @@ void readonlyGetCompleted(NcpyOperationInfo *ncpyOpInfo) {
   // Free ncpyOpInfo allocated inside readonlyGet
   CmiFree(ncpyOpInfo);
 }
+
+void _ncpy_ack_handler(ncpyHandlerMsg *msg) {
```
matthiasdiener: maybe add `inline`

matthiasdiener: I think that names that start with an underscore are reserved in C/C++. Maybe remove the underscore.

nitbhat: Done.

nitbhat: I added the underscore because other converse handlers in Charm++ were also named with that format (_initHandler, _processHandler, etc).

PR #2257, opened 2019-05-21
charmrun: Add ++interactive option for tty ssh connection to process 0

juanjgalvez:
This is necessary to support interactive charm4py sessions when
using charmrun+ssh.

### DIFF
--- a/src/util/charmrun-src/charmrun.C
+++ b/src/util/charmrun-src/charmrun.C
@@ -4451,7 +4451,7 @@ static void finish_nodes(std::vector<nodetab_process> & process_table) {}
 static void start_one_node_ssh(nodetab_process & p) {}
 static void start_nodes_mpiexec() {}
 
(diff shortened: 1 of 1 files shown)


juanjgalvez:
I tested this running a charm4py interactive session on 4 PPL lab machines and it works fine.

src/util/charmrun-src/charmrun.C
```diff
@@ -4446,7 +4451,7 @@ static void finish_nodes(std::vector<nodetab_process> & process_table) {}
 static void start_one_node_ssh(nodetab_process & p) {}
 static void start_nodes_mpiexec() {}
 
-static void finish_set_nodes(std::vector<nodetab_process> & process_table, int start, int stop) {}
+static void finish_set_nodes(std::vector<nodetab_process> & process_table, int start, int stop, bool charmrun_exiting=false) {}
```
evan-charmworks: I think having the `=false` here is a compilation error.

juanjgalvez: Good catch. I hadn't tested this on Windows.

PR #2258, opened 2019-05-21
Resolve hang in user-driven-interop on Windows

evan-charmworks:
Necessary code was entirely missing from the Windows version of the
call_startfn function.


juanjgalvez:
Travis is not going to test this on Windows. I assume you have tested in relevant windows smp modes.

PR #2259, opened 2019-05-22
Avoid use-after-free in CmiIssueRput

evan-charmworks:


evan-charmworks:
It caused a crash on Windows.

src/conv-core/conv-rdma.C
```diff
@@ -113,7 +114,7 @@ void CmiIssueRput(NcpyOperationInfo *ncpyOpInfo) {
   ncpyDirectAckHandlerFn(ncpyOpInfo);
```
stwhite91: We might want to add a comment here and by other uses of this function that it frees the ncpyOpInfo object

PR #2261, opened 2019-05-22
Add examples to Travis testing

evan-charmworks:


rbuch:
I think we should ideally not need these to be in Travis. If there are deficiencies in the tests that the examples fill in, then tests should be altered or added to increase coverage.

That being said, pragmatically, I agree that we should add these now.

evan-charmworks:
I would say that while not all tests are examples, all examples are tests, rather than the two functions being mutually exclusive.

epmikida:
I agree that all examples are tests (which means its somewhat problematic in my mind that not all examples are tested), but I could see the argument to eventually only test examples in nightly builds to expedite commit-triggered tests and/or allow for a wider set of features/applications to be tested. A lot of examples are (necessarily) redundant in what they test, and in theory no feature in examples should be without some testing in tests. But I agree for now they should be added since there are a number of examples which are the only test cases of various features.

PR #2267, opened 2019-05-23
 ZC P2P API: Deregister buffers after completion of memcpy and CMA transfers

nitbhat:


nitbhat:
Only review the commit relating to De-registering of buffers for CMA and memcpy transfers. 

The previous commit has been reviewed and approved in this PR: https://github.com/UIUC-PPL/charm/pull/2255

src/ck-core/ck.C
```diff
@@ -1433,8 +1433,10 @@ void _skipCldEnqueue(int pe,envelope *env, int infoFn)
 
 #if CMK_ONESIDED_IMPL
   // Store source information to handle acknowledgements on completion
-  if(CMI_IS_ZC_BCAST(env))
-    CkRdmaPrepareBcastMsg(env);
+  if(CMI_IS_ZC(env)) {
+    int node = CkNodeOf(pe);
+    CkRdmaPrepareZCMsg(env, node);
```
stwhite91: Maybe nicer to make this one line here and below: CkRdmaPrepareZCMsg(env, CkNodeOf(pe));

nitbhat: Yeah, I initially had that. But since I declared CkRdmaPrepareZCMsg with the second parameter being passed as reference, I get the following error:

```
ck.C: In function 'void _noCldEnqueue(int, envelope*)':
ck.C:1546:37: error: invalid initialization of non-const reference of type 'int&' from an rvalue of type 'int'
     CkRdmaPrepareZCMsg(env, CkNodeOf(pe));
```

matthiasdiener: Why is `node` passed as reference?

evan-charmworks: I don't understand why it is either. `CkRdmaPrepareZCMsg` only uses it once when calling `findTransferModeWithNodes` but that is pass-by-value.

nitbhat: I think I initially had it pass by reference because I wanted to avoid copying `node`. But, for the more common case, I'd have to explicitly create the copy anyway (by declaring `int node = CkNodeOf(pe)`), so it doesn't make a lot of sense. I'll change it to pass by value. 

I'll modify this

PR #2274, opened 2019-05-25
 PAPI: dynamically check for availability of highest cache level events to add

chin123:
In some systems PAPI is unable to read cache events in certain cache levels, thus the highest cache level which can be read must be used.

### DIFF
--- a/src/ck-perf/trace-common.C
+++ b/src/ck-perf/trace-common.C
@@ -1004,7 +1004,6 @@ void initPAPI() {
 #else
   // just uses { PAPI_L2_DCM, PAPI_FP_OPS } the 2 initialized PAPI_EVENTS
 #endif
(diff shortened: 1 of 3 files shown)


src/ck-perf/trace-common.C
```diff
@@ -1004,7 +1004,25 @@ void initPAPI() {
 #else
   // just uses { PAPI_L2_DCM, PAPI_FP_OPS } the 2 initialized PAPI_EVENTS
 #endif
-  papiRetValue = PAPI_add_events(CkpvAccess(papiEventSet), papiEvents, NUMPAPIEVENTS);
+  //papiRetValue = PAPI_add_events(CkpvAccess(papiEventSet), papiEvents, NUMPAPIEVENTS);
```
rbuch:
```suggestion
```

PR #2276, opened 2019-05-28
ZC Bcast Post API: Deregister buffers after completion of memcpy and CMA transfers

nitbhat:
Previously, buffers were de-registered after completion of RDMA transfers.
However, when buffers are declared with CK_BUFFER_REG (and PREREG for verbs
and ucx), it is required to de-register these buffers after the completion
of the memcpy and CMA transfers to avoid leaking of pinned memory.

### DIFF
--- a/src/ck-core/ckrdma.C
+++ b/src/ck-core/ckrdma.C
@@ -545,6 +545,10 @@ void performEmApiMemcpy(CkNcpyBuffer &source, CkNcpyBuffer &dest, ncpyEmApiMode 
   else if (emMode == ncpyEmApiMode::BCAST_SEND || emMode == ncpyEmApiMode::BCAST_RECV) {
     // Invoke the bcast handler
     CkRdmaEMBcastAckHandler((void *)source.refAckInfo);
+
+    // De-register dest if it has been registered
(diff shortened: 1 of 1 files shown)


src/ck-core/ckrdma.C
```diff
@@ -570,6 +574,11 @@ void performEmApiCmaTransfer(CkNcpyBuffer &source, CkNcpyBuffer &dest, int child
         CmiSetRdmaBufferInfo(dest.layerInfo + CmiGetRdmaCommonInfoSize(), dest.ptr, dest.cnt, dest.regMode);
         dest.isRegistered = true;
       }
+      // Buffers on intermediate nodes are left registered to have their child nodes rget from them
+    } else {
+      // De-regsiter dest on child nodes if it has been registered
```
matthiasdiener: Register*

nitbhat:
@matthiasdiener Fixed spelling based on your review. 

Is there a way to push to the same commit (and PR) without force push? 

The force push just dismisses previous reviews.

matthiasdiener:
I think the best is to just add another commit to the branch, and then squash them when merging.

PR #2282, opened 2019-05-29
AMPI: Fail compilation when unable to get binary path

matthiasdiener:
In this case, any AMPI application would abort anyway on startup, so better to fail at AMPI compilation already.

### DIFF
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -987,6 +987,7 @@ static void getAmpiBinaryPath() noexcept
     ampi_binary_path[n] = '\0';
   }
 #else
+// FIXME: We do not need to abort here, only if user requests pipglobals or fsglobals
 #  error "AMPI: No known way to get path to current binary."
 #endif
 }
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -987,7 +987,7 @@ static void getAmpiBinaryPath() noexcept
     ampi_binary_path[n] = '\0';
   }
 #else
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Sure, but if I understand the current code correctly, `getAmpiBinaryPath` is called on startup of each AMPI application, independent of whether it uses pipglobals or fsglobals. Are you suggesting to change the `getAmpiBinaryPath` call to only execute when pipglobals or fsglobals are requested?

evan-charmworks:
I think I made a mistake by putting an abort in this location. pipglobals and fsglobals still need to fail if no support is available here (and compile time would be better than runtime) but AMPI in general does not.

I'm not sure of the best way to only get the binary path when it is needed.

matthiasdiener:
IMHO we could put `#error` instead of the `CkAbort()`, like in this PR. It is very unlikely that the `#else` case gets executed, and this would most likely indicate some configuration error, not lack of support in the system. At least, this is how I found the issue (missed configure update when working on #2142).

matthiasdiener:
Got hit again by this bug when working on the cmake build system...

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -987,7 +987,7 @@ static void getAmpiBinaryPath() noexcept
     ampi_binary_path[n] = '\0';
   }
 #else
-  CkAbort("Could not get path to current binary!");
+#  error "AMPI: No known way to get path to current binary."
```
stwhite91: Could add a comment somewhere here that says we only really need the binary path when using pipglobals or fsglobals.

matthiasdiener: done

PR #2293, opened 2019-06-04
Zerocopy Readonly Bcast: Fix bugs to ensure that RO Bcast completion is detected correctly

nitbhat:


src/ck-core/init.C
```diff
@@ -116,8 +116,12 @@ UInt  _numInitMsgs = 0;
 CksvDeclare(UInt,_numInitNodeMsgs);
 
 #if CMK_ONESIDED_IMPL
-UInt  numZerocopyROops;
-UInt  curROIndex;
+#if CMK_SMP
+std::atomic<UInt> numZerocopyROops;
```
matthiasdiener: could probably add `= {0};` here

nitbhat: Done.

src/ck-core/init.C
```diff
@@ -914,20 +918,19 @@ static inline void _processRODataMsg(envelope *env)
...
-#endif
-
     //Unpack each readonly
     PUP::fromMem pu((char *)EnvToUsr(env));
     CmiSpanningTreeInfo &t = *_topoTree;
 
 #if CMK_ONESIDED_IMPL
-    pu|numZerocopyROops;
+#if CMK_SMP
+    UInt numZerocopyROopsTemp;
```
matthiasdiener: Just to clarify, this is necessary because pup doesn't support atomic?

matthiasdiener: also, would it be possible/easier to do a cast, ie, `pu|(Uint) numZerocopyROOps` (or call `pup()` explicitly)?

nitbhat: Yeah, I added because it seemed like pup didn't support atomics. 

```
init.C:932:9: error: invalid conversion from 'std::__atomic_base<unsigned int>::__int_type' {aka 'unsigned int'} to 'mCastEntryPtr' {aka 'mCastEntry*'} [-fpermissive]
     pu|((UInt)numZerocopyROops);
        ~^~~~~~~~~~~~~~~~~~~~~~~
```

```
init.C:933:8: error: invalid user-defined conversion from 'std::atomic<unsigned int>' to 'PUP::able**' [-fpermissive]
     pu(numZerocopyROops);
        ^~~~~~~~~~~~~~~~
```

matthiasdiener: Hmm, interesting. Do these also not work?:
`p(numZerocopyROops);` and `p((UInt)numZerocopyROops);`?

nitbhat:
```
init.C:933:8: error: invalid conversion from 'std::__atomic_base<unsigned int>::__int_type' {aka 'unsigned int'} to 'CkMigrateMessage*' [-fpermissive]
     pu((UInt)numZerocopyROops);
        ^~~~~~~~~~~~~~~~~~~~~~
```

matthiasdiener: Ok, this is unfortunate.

stwhite91: We should be able to add a generic pup routine for std::atomic<T> that loads/stores the value in and out of a bare T, then this code can be simplified.

Issue #1436, opened 2017-02-17
trace CcdCallFnAfter() causality


jcphill:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1436*

---
NAMD uses CcdCallFnAfter() to poll for GPU completion, which breaks the causality chain.  It should be possible to trace a chain of CcdCallFnAfter() calls back to the entry method that called the first one.

PhilMiller:
*Original date: 2017-10-11 20:14:59*

---
Ronak or Karthik, please get whatever further details are necessary, and decide if this should be addressed in the next release timeframe.

rbuch:
*Original date: 2019-04-04 19:15:14*

---
I think this is doable, but the detail of functions coming from a CcdCallFnAfter may be a bit suspect (since it just calls a function directly, so it's hard for the RTS to know what's executing) and the ordering causality chain may include extraneous calls due to differing levels of delays for different invocations of  CcdCallFnAfter. Regardless, getting causation to the right place may be a bit tricky without adding significantly more disambiguation code to adding stuff to the CcdCallFnAfter call. I've started toying with an implementation to see if I can get it to work correctly or not.

PR #2296, opened 2019-06-05
Feature #1436: Trace CcdCallFnAfter() causality

rbuch:
Creates dummy message sends to trace invocations of CcdCallFnAfter back to the
original entry method


src/conv-core/conv-conds.C
```diff
@@ -421,6 +446,17 @@ void CcdModuleInit(char **ignored)
    CpvInitialize(int, ccd_heapmaxlen);
    CpvInitialize(int, _ccd_numchecks);
 
+#if CMK_TRACE_ENABLED
+   CpvInitialize(envelope*, dummyEnv);
```
stwhite91: Can you add a comment here describing the use of dummyMsg?

PR #2303, opened 2019-06-07
Ampi privatisation tutorial

raj-aayush:


matthiasdiener:
Why was the previous pull request with this patch closed?

raj-aayush:
> Why was the previous pull request with this patch closed?

The Makefile in the MPI version of the PENNANT program created a slight issue with the executable. I didn't notice that when I first created the pull request. It's fixed in this version now.

examples/ampi/privatisation_tutorial/nm_globals.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
matthiasdiener: Is this copy of nm_globals.sh required?

raj-aayush: This copy is placed there for convenience. I could get rid of it if needed.

matthiasdiener: Would it make sense to replace the copy with a symlink or so?

raj-aayush: That's a good point. I will make that change. So will this pull request need to closed?

matthiasdiener: No need to close this pull request, you can just push another commit to the branch to update it.

raj-aayush: Made the changes to nm_globals.sh. Please verify.

matthiasdiener: looks good to me

evan-charmworks:
It looks like some other files too. Maybe `make clean` should be run.

evan-charmworks:
This is good work, but I don't think we should bring external MPI codes into the main Charm repository.

ericjbohm:
We already have a few, such as the mpich tests.  So its more of a question
of quantity and what utility they bring.

PR #2318, opened 2019-06-18
UCX: Add support for ZC API

nitbhat:


nitbhat:
@brminich: Ronak has sent you an invite to become a collaborator for our repo. Can you accept the invite and let me know? Once you do, I'll add you as a reviewer for this patch.

src/arch/ucx/machine-onesided.C
```diff
@@ -131,12 +142,9 @@ void LrtsIssueRget(NcpyOperationInfo *ncpyOpInfo)
     } else {
         // Remote buffer is not registered, ask peer to perform put
```
brminich: minor: this comment is not relevant anymore

src/arch/ucx/machine-onesided.C
```diff
@@ -147,21 +155,38 @@ void LrtsIssueRput(NcpyOperationInfo *ncpyOpInfo)
...
-    } else {
-        // Remote buffer is not registered, ask peer to perform get
-        UcxSendMsg(CmiNodeOf(ncpyOpInfo->destPe), ncpyOpInfo->destPe,
-                   ncpyOpInfo->ncpyOpInfoSize, (char*)ncpyOpInfo,
-                   UCX_RMA_TAG_GET, UcxRmaSendCompleted);
-        UCX_LOG(4, "Sending Get REQ to %d, mem size %d", ncpyOpInfo->destPe,ncpyOpInfo->srcSize);
-
-    }
+    // Since UCX doesn't guarantee destination completion with PUT, it is implemented
+    // as a reverse operation with the remote destination performing a GET
```
brminich: maybe it is worth to register src here (if needed)? Otherwise remote side will have to request registration

brminich:
> @brminich: Ronak has sent you an invite to become a collaborator for our repo. Can you accept the invite and let me know? Once you do, I'll add you as a reviewer for this patch.

Just accepted (sorry missed it at the first time)

stwhite91:
bump, needs review comments addressed

src/arch/ucx/machine-onesided.C
nitbhat: Since `LrtsIssueRput` is invoked on the source, the higher layer code (ck-core) ensures that the local (source) buffer is always registered if it is required. (at /ck-core/ckrdma.C:190).

src/arch/ucx/machine-onesided.C
nitbhat: Good point, fixed in the latest push.

nitbhat:
All review comments have been addressed. The patch is ready for a final review.

src/arch/gni/machine-onesided.C
```diff
@@ -367,6 +367,9 @@ void _performOneRputForWorkerThread(MSG_LIST *ptr) {
 
 void LrtsInvokeRemoteDeregAckHandler(int pe, NcpyOperationInfo *ncpyOpInfo) {
 
+  if(ncpyOpInfo->opMode == CMK_BCAST_EM_API)
```
stwhite91: This should presumably be in a separate patch?

src/arch/ofi/machine-onesided.C
```diff
@@ -333,6 +333,9 @@ void LrtsDeregisterMem(const void *ptr, void *info, int pe, unsigned short int m
 
 void LrtsInvokeRemoteDeregAckHandler(int pe, NcpyOperationInfo *ncpyOpInfo) {
 
+  if(ncpyOpInfo->opMode == CMK_BCAST_EM_API)
```
stwhite91: Same here

src/arch/gni/machine-onesided.C
nitbhat: This change of opMode being `CMK_BCAST_EM_API` was introduced because of the UCX layer requiring de-registration support. For that reason Gni, OFI and Verbs also had to support that opMode, where they had to simply not do anything (and hence return).

stwhite91: Can you explain this in the body of the commit msg?

nitbhat: Yes, I can.

nitbhat:
I tested the SMP and a few tests are crashing. I'll debug those and update here when things are passing.

nitbhat:
Hi Mikhail, I am seeing some bugs when I test the ZC API in the SMP mode. I wanted your help in understanding some of the code and debugging these issues. 

1. Firstly, in the ZC API, when small messages are sent with the tag UCX_RMA_TAG_REG_AND_SEND_BACK, UCX_RMA_TAG_GET and UCX_RMA_TAG_DEREG_AND_ACK, in the non-SMP mode, the message buffer (req->msgBuf) is not freed. However, I see that there's a `CmiFree(req->msgBuf)` in `ProcessTxQueue`, for the SMP mode. This free is causing an issue. I think we should selectively free the buffer when the tags are not the ones above mentioned. 

2) Why are we freeing the request in `ProcessTxQueue`? (since the callback is called in the SMP mode and the freeing code already exists in the different callbacks like `UcxTxReqCompleted`).

3) In `UcxRmaOp`, I see that after posting the operation, you have a check with `UCS_PTR_IS_PTR`. What exactly does that function do? In the SMP mode, I see that after posting the RDMA operation (GET or PUT), the callback `UcxRdmaReqCompleted` is invoked but gets `req->ncpyAck` as NULL.

brminich:
1) You mean that we free `ncpyOpInfo` object in SMP mode? We probably need to pass some flag to UcxSendMsg function which would indicate whether the message needs to be freed (and save it in SMP request for future use in `ProcessTxQueue`). Or try to inspect tag to identify whether it is ZC API (i.e whether message needs to be freed). I could fix it, if you have not started to do that yet

2) We free SMP request, not UCX. Completion callbacks would free UCX request

3) Some UCX communication calls (including RMA) may return `ucs_status_ptr_t`. Which means that it is either a valid request or just status (error or ok). If status is returned, it means the operation completed immediately and there is no need to have request. 
BTW, should ZC API communication calls be deferred to communication thread in SMP mode? AFAIR, I noticed that some machine layers invoke comm calls directly even in SMP mode, so implemented it the same way.

nitbhat:
Hi Mikhail, 

Thanks for getting back. 

1. Yes, I think we end up freeing `ncpyOpInfo` in the SMP mode. Yes, there is a flag inside `ncpyOpInfo` called `ncpyOpInfo->freeMe` and if it is equal to `CMK_FREE_NCPYOPINFO`, the object has to be freed. I tried implementing this by looking at the tag value, but for some reason, I was getting weird tag values (100+), if I remember correctly (which seemed odd). Yes, but what we'd have to do is look at the tag and cast `msgBuf` into `NcpyOperationInfo` and then free only if `freeMe` is equal to `CMK_FREE_NCPYOPINFO`. 

2. Oh okay, I thought both the requests were the same.

3.  Okay. When I was testing out the SMP code, I saw that this if block was being executed in machine-onesided.C.
```
127     if (!UCS_PTR_IS_PTR(statusReq)) {
128         CmiEnforce(UCS_PTR_STATUS(statusReq) == UCS_OK);
129         CmiInvokeNcpyAck(ncpyOpInfo);
130         ucp_rkey_destroy(rkey);
131     } else {
```
However, the CmiEnforce on line 128 was failing. 

That's a good point. Ideally, it'll be good to post the RMA calls from the comm thread in order to parallelize it and have the worker thread return to executing other things.

brminich:
Hi Nitin,

1. The tag contains info about send protocol ("direct" eager or "probe" rendezvous) and also RMA op id (which is zero in case of regular send). So you can check if tag contains any bits from UCX_RMA_TAG_MASK mask. If yes, no need to delete, it is ncpyOpInfo object. It can be done in `ProcessTxQueue`, where tag is available. 

3. Can you please check what is exact error when it is not UCS_OK?

nitbhat:
Hi Mikhail, 

Thanks for getting back. 

1. I was working with the tags incorrectly. Previously, I hadn't noticed that the tags were modified with bitwise ORs, so I was trying to equate their value directly. I've fixed this in the latest push. 

2. As I was implementing the posting of RGET/RPUT from the comm thread, the other error I was seeing (related to UCS_OK) also ended up fixed. Now, I've modified the code to post RGET/RPUT from the comm thread only. 

I tested both the nonsmp and smp versions and both seem to working fine. The patch is ready for a set of final reviews and merge. 

Additionally, I enabled `CMK_IMMEDIATE_MSG` because it can help applications that send immediate messages (which directly execute on the comm thread, skipping the scheduler). It is also required by the Zerocopy API for Quiescence Detection. (as previously, the ZC QD Test was failing). 

Thanks,
Nitin

PR #2327, opened 2019-06-19
Require all Charm++ tests to meaningfully test SMP

epmikida:
This is the first step in a larger effort to cleanup/improve Charm++ tests.
This adds a new make target, smptest, which is run on every test. Tests that
don't define it will fail. Almost all existing tests now also run on a single
multi-PE process (+p2 ++ppn 2) and mutliple multi-PE processes (+p4 ++ppn 2).

### DIFF
--- a/tests/charm++/queue/Makefile
+++ b/tests/charm++/queue/Makefile
@@ -8,6 +8,8 @@ test: $(TARGETS)
 	$(call run, ./pgm  +p1)
 	$(call run, ./msgqtest  +p1)
 
+smptest: $(TARGETS)
+
 pgm.C: main.decl.h
 
 msgqtest.C: main.decl.h
--- a/tests/charm++/zerocopy/zc_post_modify_size/Makefile
+++ b/tests/charm++/zerocopy/zc_post_modify_size/Makefile
@@ -21,6 +21,10 @@ test: all
 	$(call run, +p4 ./zc_post_modify_size)
 	$(call run, +p4 ./zc_post_modify_size +noCMAForZC)
 
+smptest: all
+	$(call run, +p2 ./zc_post_modify_size ++ppn 2)
+	$(call run, +p4 ./zc_post_modify_size ++ppn 2)
+
 clean:
 	rm -f *.decl.h *.def.h *.o
 	rm -f zc_post_modify_size charmrun cifiles
--- a/tests/charm++/zerocopy/zc_post_modify_size/Makefile
+++ b/tests/charm++/zerocopy/zc_post_modify_size/Makefile
@@ -24,6 +24,8 @@ test: all
 smptest: all
 	$(call run, +p2 ./zc_post_modify_size ++ppn 2)
 	$(call run, +p4 ./zc_post_modify_size ++ppn 2)
+	$(call run, +p2 ./zc_post_modify_size +noCMAForZC ++ppn 2)
+	$(call run, +p4 ./zc_post_modify_size +noCMAForZC ++ppn 2)
 
 clean:
 	rm -f *.decl.h *.def.h *.o
--- a/tests/charm++/Makefile
+++ b/tests/charm++/Makefile
@@ -1,3 +1,5 @@
+-include ../../include/conv-mach-opt.mak
+
 DIRS = \
   megatest \
   alignment \
@@ -39,6 +41,13 @@ TESTDIRS = $(filter-out $(FTDIRS),$(DIRS))
 all: $(foreach i,$(DIRS),build-$i)
 
 test: $(foreach i,$(TESTDIRS),test-$i)
+ifeq ($(CMK_SMP),1)
+ifneq ($(CMK_MULTICORE),1)
+	make smptest
+endif
+endif
+
+smptest: $(foreach i,$(TESTDIRS),smptest-$i)
 
 bgtest: $(foreach i,$(filter $(BGDIRS),$(TESTDIRS)),bgtest-$i)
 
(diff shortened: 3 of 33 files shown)


tests/charm++/queue/Makefile
```diff
@@ -8,6 +8,8 @@ test: $(TARGETS)
 	$(call run, ./pgm  +p1)
 	$(call run, ./msgqtest  +p1)
 
+smptest: $(TARGETS)
```
rbuch: Should this have some run commands with it?

epmikida: These test is a sequential test that just creates a queue and does some tests on it. Does need to be run on multiple pes/cores/nodes/etc.

epmikida:
This may have caused some flaky tests in SDAG, so we should probably hold off in the merge. I'll look into what breaks the SDAG tests when they run in SMP mode.

epmikida:
> todo: look at SDAG tests

So it turns out they weren't flaky. It was just that the anytimeMigration test took an absurd amount of time which made it look like a hang. I've reduced the size of the test for now and I think this change is ready for merge.

I am separately looking at rewriting the SDAG tests because they dont have great coverage, so I will make sure to re-examine the anytime migration test then as well.

evan-charmworks:
Is there a benefit to adding these new `smptest` rules instead of simply fixing `make test TESTOPTS="<some ppn argument>"` to work, assuming we update `testrun` so that any `++ppn` passed in the Makefile will override the user's choice?

epmikida:
> Is there a benefit to adding these new `smptest` rules instead of simply fixing `make test TESTOPTS="<some ppn argument>"` to work by updating `testrun` so that any `++ppn` passed in the Makefile will override the user's choice?

I chose this approach for a few reasons, but it's certainly something that could be changed:
1) Putting default ++ppn inside of testrun is different from how we handle +p, which we still require explicit choices for in the tests.
2) Testing fails if the tests do not include an `smptest` target which means for newly added tests there is an explicit check to force people to consciously think about how their test interacts with SMP.
3) IMO how the tests are run would be hidden/unclear if testrun adds in too much behind the scenes, with no indication in the makefile of what is happening.

Point #1 is certainly something we could change further down the line, but I do think +p and ++ppn should be handled consistently.

evan-charmworks:
1. We wouldn't put `++ppn` inside testrun, we would change it so that any `+p` or `++ppn` passed in the Makefile will override anything the user passes into `TESTOPTS`.
2. I agree that all test writers should be required to think about SMP, but I don't agree that adding a new target is the way to do it. Ideally our testing infrastructure would run `make test TESTOPTS="<some SMP options>"` and validate functionality that way.
3. Testrun wouldn't add anything on its own. Any options written in the makefile would have final say over any TESTOPTS the user passes in for the whole test run.

epmikida:
> 1. We wouldn't put `++ppn` inside testrun, we would change it so that any `+p` or `++ppn` passed in the Makefile will override anything the user passes into `TESTOPTS`.
> 2. I agree that all test writers should be required to think about SMP, but I don't agree that adding a new target is the way to do it. Ideally our testing infrastructure would run `make test TESTOPTS="<some SMP options>"` and validate functionality that way.
> 3. Testrun wouldn't add anything on its own. Any options written in the makefile would have final say over any TESTOPTS the user passes in for the whole test run.

So what would a Makefile look like in your plan? It would seem to me that we'd end up going back to all test targets having the ifeq(CMK_SMP,1) tests an passing in ++ppn in that way? The difference being that if someone does not include that, we'd have default ++ppns passed in from somewhere else?

evan-charmworks:
Except for tests that require an exact number of PEs like pingpong or test specific functionality like provisioning, ideally most tests would not specify any `+p` or `++ppn` in the Makefile and let the user choose how they are run through TESTOPTS.

That said, there are more preexisting `+p` parameters in our test Makefiles than I anticipated:
```
$ grep -r --include=Makefile '+p' | wc -l
     718
```
It seems many of these run the same test with a few different combinations of values. We might need to take a closer look to see which of these are justified and which are better served by running _all_ tests with various permutations through TESTOPTS.

evan-charmworks:
The more I look into my idea the more it seems like it would take more effort than the solution in this PR. We should move forward with merging this for now so we can improve our SMP test coverage right away.

evan-charmworks:
Needs manual conflict resolution

epmikida:
> The more I look into my idea the more it seems like it would take more effort than the solution in this PR. We should move forward with merging this for now so we can improve our SMP test coverage right away.

I agree that the testing infrastructure needs more comprehensive changes, and this is only a step along the way. I'll resolve the conflicts now so it can be merged.

nitbhat:
Yes, I agree. This is a quick way of integrating SMP tests with simple changes. We could parallely work on generic solutions that help improve our test infrastructure. (like multiple combination testing of +p and ++ppn for all tests in the test suite).

nitbhat:
I think one of the newly added tests (tests/charm++/zerocopy/bcast_nonzero_root/) requires an `smptest` target.

nitbhat:
Nvm, just saw the latest commit.

tests/charm++/zerocopy/zc_post_modify_size/Makefile
```diff
@@ -21,6 +21,10 @@ test: all
 	$(call run, +p4 ./zc_post_modify_size)
 	$(call run, +p4 ./zc_post_modify_size +noCMAForZC)
 
+smptest: all
+	$(call run, +p2 ./zc_post_modify_size ++ppn 2)
+	$(call run, +p4 ./zc_post_modify_size ++ppn 2)
```
stwhite91: We probably want a smptest with multiple processes and `+noCMAForZC` as well

stwhite91: Same goes for all ZC tests

stwhite91:
I think we want tests/ and benchmarks/ for all of charm++/, converse/, and ampi/. Could also do examples/ but maybe lower priority

epmikida:
> I think we want tests/ and benchmarks/ for all of charm++/, converse/, and ampi/. Could also do examples/ but maybe lower priority

Agreed. Are you suggesting holding this patch until the rest of tests/ have been added, or are those fine in separate PRs?

stwhite91:
I'm fine with it being in a separate commit, but as soon as we have the smptests for tests/charm++/ passing, we should get the rest running too

epmikida:
Fixed ZC tests, making other branches for ampi and conv tests.

evan-charmworks:
Could use a review and merge before autobuild begins

PR #2329, opened 2019-06-20
Fix de-registration bug in benchmarks/charm++/pingpong/

nitbhat:
This commit changes the deregMode of the buffer info objects using
the Direct API in benchmarks/charm++/pingpong to CK_BUFFER_NODEREG.
This is required to ensure that the buffers are not de-registered
after the completion of the operation.


nitbhat:
While changing DEREG -> NODEREG for benchmarks, I missed out on this program. 

This should fix the autobuild failure on ofi and verbs.

PR #2336, opened 2019-06-20
Fix Charm4py registering extra entry methods, and some cleanup

juanjgalvez:
Minor bug fixes and cleanup.

This is all code inside `#if CMK_CHARMPY` blocks.

### DIFF
--- a/src/ck-core/register.C
+++ b/src/ck-core/register.C
@@ -99,17 +99,17 @@ void CkRegisterGroupIrr(int chareIndex,int isIrr){
 #if CMK_CHARMPY
 
 // TODO give a unique name to entry methods when calling CkRegisterEp
+// (no need has appeared for this so this is very low priority)
+
 void CkRegisterGroupExt(const char *s, int numEntryMethods, int *chareIdx, int *startEpIdx) {
   int __idx = CkRegisterChare(s, sizeof(GroupExt), TypeGroup);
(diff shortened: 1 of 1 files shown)


src/ck-core/register.C
```diff
@@ -99,40 +99,36 @@ void CkRegisterGroupIrr(int chareIndex,int isIrr){
...
-  CkRegisterGroupIrr(__idx, true); // isIrreducible?
 
   int epIdxCtor = CkRegisterEp(s, GroupExt::__GroupExt, CkMarshallMsg::__idx, __idx, 0+CK_EP_NOKEEP);
   CkRegisterDefaultCtor(__idx, epIdxCtor);
 
-  for (int i=0; i < numEntryMethods; i++)
-    int epidx = CkRegisterEp(s, GroupExt::__entryMethod,
-                             CkMarshallMsg::__idx, __idx, 0+CK_EP_NOKEEP);
+  for (int i=1; i < numEntryMethods; i++)
+    CkRegisterEp(s, GroupExt::__entryMethod, CkMarshallMsg::__idx, __idx, 0+CK_EP_NOKEEP);
```
stwhite91: Why always nokeep?

juanjgalvez: The current scheme in Charm4py for every method invocation is similar to parameter marshalling. From the Charm message buffer, Python objects are unpickled, or Numpy arrays are created by copying the contents into new buffers. I have not thought about alternative schemes yet.

PR #2343, opened 2019-06-21
Add within node broadcast functionality to PAMI

epmikida:


evan-charmworks:
Does uth also need this code?

epmikida:
> Does uth also need this code?

Yes that is possible. I'm not sure how many of the machine layers rewrite their own broadcast functions.

epmikida:
> Fixes pami compile, not sure if bug exists other layers.

Do the tests pass as well? Specifically tests/charm++/within_node

nitbhat:
```
./shmem/machine.C:void CmiSyncBroadcastFn(int size, char *msg)
./sim/machine.C:void CmiSyncBroadcastFn(int size,  char *msg)
./uth/machine.C:void CmiSyncBroadcastFn(int size, char *msg)
```

These layers don't use LRTS and would need their own within node bcast fn.

jcphill:
I'm actually getting segfaults on Summit for simplearrayhello now.

jcphill:
Segfaults…

evan-charmworks:
> Nevermind. The segfaults are in --with-debug, while --with-production works fine.

That still sounds concerning. Do these segfaults predate the introduction of WithinNodeBroadcast?

PR #2344, opened 2019-06-23
Change maximum statistic calculation in PICS

chin123:
From my understanding, there is no reason why the calculation for the maximum value of the statistics is done in this way, thus I have changed it to be more straightforward.


chin123:
The reason that the current implementation is skipping every other integer when combining the `MAX` fields is because every alternate field in the enum contains the PE associated with the previous field. The other fields (those with the prefixes `MIN` and `AVG`, do not contain the PE associated with each field. This could probably be made more clear through documenting this file, and perhaps similar fields could be added to `MIN` as well. Thus I am closing this PR.

PR #2208, opened 2019-05-08
CHANGES: add release notes for v6.10.0

evan-charmworks:
*Original date: 2019-05-08 04:12:33*
*Original PR: https://charm.cs.illinois.edu/gerrit/5152*

---

Change-Id: Iee9fd0da868856c0f6461800fcdd0b99a1c4523e

PR #2348, opened 2019-06-24
Documentation: specify that users should not free the migration ctor msg

stwhite91:


juanjgalvez:
The migration constructor message should not be deleted by the user under no circumstance? Where is it deleted then? By charmxi generated code?

The ArrayElemExt class (used by Charm4py) deletes it explictly inside the migration constructor, and the constructor is registered as "keep". But I don't remember why I did this, I guess I just copied the pattern from charmxi generated code.

epmikida:
Yeah I think that means examples and tests are wrong. I've always deleted the migration ctor message for array elements. I just tested it as well by adding the delete to the migration ctor for tests/charm++/load_balancing/lb_test. So its inconsistent across migration ctors AND our examples are wrong. I don't know which way we want to swing though, have RTS always delete migration ctor messages, or leave that to the users. I think I'd lean towards the latter, especially because part of Jozsef's confusion came from the fact that its not how main ctor works either.

But then again, if we think we are moving to the default behavior for other charm messages to become nokeep then maybe we should try to make the runtime system delete messages unless the user specifically says not to.

epmikida:
Actually, explicitly double freeing the message in that test also works fine, so maybe something else is wrong. Still looking into it.

epmikida:
Update for this:
For mainchares, the CkMigrateMessage* is just a cast of CkArgMsg. But for some reason, in checkpoint restart, we delete it, and during normal startup we don't. This seems like a weird choice.

For array elements, the CkMigrateMessage* is only for the name to disambiguate ctor calls, and is typedef'd to a struct { int ...; }. Furthermore, it looks like NULL is always what is passed in anyways, in both migration and checkpoint restart. So no leaked memory, and it doesn't matter what the user does.

For group elements, CkMigrateMessage* is an actual message (but looks like just the envelope), who's message type is set to BocInitMsg and follows the regular entry method invocation path. This message is not deleted.

IMO, we should remove the automatic deletion of the arg message (and also maybe rename it so its more explicit and users don't have to cast it?) and then make it clear in the documentation that users are responsible for deleting these messages. Thoughts?

juanjgalvez:
Does the migration constructor serve some point other than telling the user that the chare has migrated? That part has never been clear to me. Wouldn't a `void Chare::migrated()` or similar serve this purpose better?

epmikida:
The code path (at least for array elements and groups) is the same for regular construction and migration construction, so without more significant changes to those paths, you'd need a constructor to call when the elements are being created. And at that time you don't have arguments to pass so in theory we need an overloaded ctor to call.

And we do already have ckJustMigrated, so you could argue that the migration constructor doesn't even serve that purpose. However I've heard its for cases where you need to do stuff before pupping occurs. I'm not 100% sure what that would be though...

In Charades I've used it to have chares register with local group branches after migration but before pup occurs, but I don't think theres anything preventing me from doing that at beginning of the pup() function instead.

epmikida:
Also, for main chares of course, the arg message serves a purpose.

juanjgalvez:
Ok. Regarding the CkMigrateMessage msg, it seems weird to me that the user would be responsible for deleting something that he is not going to use and doesn't care about. The only point of that type seems to be to allow constructor overloading. Can we define a dummy type that is not a pointer and just let the users ignore it? If internally there is some msg associated with migration (you mentioned that for group elements it is an actual msg) then the runtime could delete it. Basically, if there is some good way of letting the user ignore that parameter I think that would be best.

epmikida:
Yes, I'd also be for having the RTS delete the messages. But if we do that, I think we should also change the behavior of mainchare constructors. Since the mainchare constructor and the mainchare migration ctor are basically the exact same, with the exact same message, it seems silly to have them behave differently. Plus, I also don't think theres any reason to make applications responsible for the arg message either. Right now its just us requiring them to add another easy-to-forget delete to their main constructor.

evan-charmworks:
It is likely important to distinguish construction of an object after migration from its original construction so that the default constructor does not run instead. This is the same principle used for `PUP::reconstruct` constructors.

evan-charmworks:
Was this intentionally closed?

stwhite91:
I suppose we still should merge this, perhaps with modifications

nitbhat:
I think we should update https://github.com/UIUC-PPL/charm/pull/2208 with this new change.

Isn't the actual change (of deleting `CkMigrateMessage` in the RTS) an application breaking change for Apps that have already deleted `CkMigrateMessage` in the application because of valgrind pointing to it as a leak? Are we okay with that?

epmikida:
> I think we should update #2208 with this new change.
> 
> Isn't the actual change (of deleting `CkMigrateMessage` in the RTS) an application breaking change for Apps that have already deleted `CkMigrateMessage` in the application because of valgrind pointing to it as a leak? Are we okay with that?

This was just a documentation change. Also, has anyone made changes to this? I was working on a patch to make it so the RTS automatically freed group migrate messages, but it seems like that is either already done, or something else weird is happening. I can't find where the messages are being deleted anywhere, but deleting them myself is causing crashes. It's causing crashes before there would be a second free though.

epmikida:
> > I think we should update #2208 with this new change.
> > Isn't the actual change (of deleting `CkMigrateMessage` in the RTS) an application breaking change for Apps that have already deleted `CkMigrateMessage` in the application because of valgrind pointing to it as a leak? Are we okay with that?
> 
> This was just a documentation change. Also, has anyone made changes to this? I was working on a patch to make it so the RTS automatically freed group migrate messages, but it seems like that is either already done, or something else weird is happening. I can't find where the messages are being deleted anywhere, but deleting them myself is causing crashes. It's causing crashes before there would be a second free though.

I found the problem. They are being allocated as system messages (although I still think they may be completely unused). So delete is not the correct way to free them, but nowhere else frees them either. Working on a patch now.

PR #2350, opened 2019-06-24
PICS: fix the way steps are counted

chin123:
Previously, if the collected data was requested to be analyzed more than once in a single step, the subsequent requests would be ignored. However, this change ensures that it is requested at a maximum of once per step, thus removing the need for a check to see whether the number of steps since the last analysis is greater than 0.


src/ck-pics/picsautoperf.C
```diff
@@ -735,6 +739,7 @@ extern "C" void traceAutoPerfExitFunction() {
   }
 
   CkpvAccess(isExit) = true;
+
```
rbuch: Extraneous whitespace.

chin123:
Closed in favour of #2393.

PR #2352, opened 2019-06-24
Fix breakages in UTH from recent changes

epmikida:
Within node broadcasts need to be defined for the UTH layer and the MetaLB test
must assume 0 migrations when running on UTH.

### DIFF
--- a/tests/charm++/load_balancing/meta_lb_test/period_selection.C
+++ b/tests/charm++/load_balancing/meta_lb_test/period_selection.C
@@ -41,10 +41,15 @@ public:
     if (iteration < MAX_ITER) {
       arrayProxy.balance(iteration);
     } else {
-      if (CkNumPes() > 1 && migrations != 10) {
+      int expected_migrations = 10;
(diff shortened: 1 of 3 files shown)


tests/charm++/load_balancing/meta_lb_test/period_selection.C
```diff
@@ -41,10 +41,15 @@ public:
     if (iteration < MAX_ITER) {
       arrayProxy.balance(iteration);
     } else {
-      if (CkNumPes() > 1 && migrations != 10) {
+      int expected_migrations = 10;
+      if (CkNumPes() == 1) {
```
evan-charmworks: Should this CkNumPes() and the one below be CkNumNodes()? If so, would that eliminate the need for the `#if CMK_UTH_VERSION` part?

evan-charmworks: Nevermind, I bet within-node balancing is something we do.

PR #2359, opened 2019-06-25
Fix and replace SMP barriers with C++11 version

rbuch:
Previously, CmiNodeBarrier and CmiNodeAllBarrier had a bug on
non-Windows platforms because the same state variables were shared
between the two different kinds of barriers. This fixes that and is
also more portable, so it can replace both the previous Windows and
pthreads implementations of barriers.

### DIFF
--- a/src/arch/util/machine-smp.C
+++ b/src/arch/util/machine-smp.C
@@ -81,14 +81,13 @@ CmiIdleLock_checkMessage
 #include "machine-smp.h"
 #include "sockRoutines.h"
 
+#include <mutex>
+#include <condition_variable>
+
 void CmiStateInit(int pe, int rank, CmiState state);
 void CommunicationServerInit(void);
 
 static struct CmiStateStruct Cmi_default_state; /* State structure to return during startup */
-
-#define CMI_NUM_NODE_BARRIER_TYPES 2
-#define CMI_NODE_BARRIER 0
-#define CMI_NODE_ALL_BARRIER 1
 
 /************************ Win32 kernel SMP threads **************/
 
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Fails to compile with the following error: 

```
In file included from machine-common-core.C:322:0,
                 from machine.C:228:
machine-smp.C: In function ‘void CmiDestroyLocks()’:
machine-smp.C:525:26: error: ‘barrier_mutex’ was not declared in this scope
   pthread_mutex_destroy(&barrier_mutex);
                          ^
Fatal Error by charmc in directory /home/circleci/project/netlrts-linux-x86_64-smp/tmp
   Command g++ -DCMK_GFORTRAN -D_REENTRANT -I../bin/../include -D__CHARMC__=1 -I. -g -Werror=vla -O3 -fno-stack-protector -std=c++11 -fno-lifetime-dse -c machine.C -o machine.o returned error code 1
charmc exiting...
Makefile:605: recipe for target 'machine.o' failed
make: *** [machine.o] Error 1
```

rbuch:
> Fails to compile with the following error:
> 
> ```
> In file included from machine-common-core.C:322:0,
>                  from machine.C:228:
> machine-smp.C: In function ‘void CmiDestroyLocks()’:
> machine-smp.C:525:26: error: ‘barrier_mutex’ was not declared in this scope
>    pthread_mutex_destroy(&barrier_mutex);
>                           ^
> Fatal Error by charmc in directory /home/circleci/project/netlrts-linux-x86_64-smp/tmp
>    Command g++ -DCMK_GFORTRAN -D_REENTRANT -I../bin/../include -D__CHARMC__=1 -I. -g -Werror=vla -O3 -fno-stack-protector -std=c++11 -fno-lifetime-dse -c machine.C -o machine.o returned error code 1
> charmc exiting...
> Makefile:605: recipe for target 'machine.o' failed
> make: *** [machine.o] Error 1
> ```

Yeah, I think I tested it before I deleted the declaration of that variable. I've just pushed a fix.

src/arch/util/machine-smp.C
```diff
@@ -589,21 +531,60 @@ static void CmiDestroyLocks(void)
...
+public:
+  Barrier(const Barrier&) = delete;
+  Barrier& operator=(const Barrier&) = delete;
+
+  explicit Barrier(unsigned int count) :
+    curCount(count), barrierCount(count), curGeneration(0) {
+  }
+
+  void wait() {
+    std::unique_lock<std::mutex> lock(barrierMutex);
```
juanjgalvez: Lock is released automatically when going out of scope?

rbuch: Yes.

src/arch/util/machine-smp.C
```diff
@@ -589,21 +531,60 @@ static void CmiDestroyLocks(void)
...
+  unsigned int curGeneration;
+  unsigned int curCount;
+  const unsigned int barrierCount;
+};
+
 /* Wait for all worker threads */
-void  CmiNodeBarrier(void) {
-  CmiNodeBarrierCount(CmiMyNodeSize(), CMI_NODE_BARRIER);
+void CmiNodeBarrier(void) {
+  static Barrier nodeBarrier(CmiMyNodeSize());
```
juanjgalvez: I'm wondering how static initialization works here when there are multiple threads. AFAIK the static is initialized the first time the function is executed. What happens with multiple threads calling at the same time?

rbuch: Yes, that's right. Since C++11, ["If multiple threads attempt to initialize the same static local variable concurrently, the initialization occurs exactly once."](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables)

juanjgalvez: Cool

rbuch:
> Does this allow us to collapse the triple Barrier call in init.C to one?

This doesn't change the semantics of the barriers, so that's a fully orthogonal question to this change.

PR #2360, opened 2019-06-26
Fix a race in MetaLB regarding 0 object PEs

epmikida:
If the intial periodic call to check for 0 object PEs triggered at a weird
time (ie during migration), it could incorrectly detect an empty PE and cause
a mismatched reduction contribution. This makes it so the periodic call is only
checked if we have not received any load contributions at the time it triggers.

### DIFF
--- a/src/ck-ldb/MetaBalancer.C
+++ b/src/ck-ldb/MetaBalancer.C
@@ -163,7 +163,7 @@ void MetaBalancer::init(void) {
   lb_in_progress = false;
 
   // After the first reduction completes, we have a way to manually check for
-  // PEs withouth any objects. This flag is set to true if load is ever added
+  // PEs without any objects. This flag is set to true if load is ever added
   // so that the periodic call can be ignored if it comes late. Otherwise, it
   // may come at a weird time and trigger an incorrect contribution.
   ignore_periodic = false;
--- a/src/ck-ldb/MetaBalancer.C
+++ b/src/ck-ldb/MetaBalancer.C
@@ -162,6 +162,12 @@ void MetaBalancer::init(void) {
   // to false so that it doesn't clear the handles.
   lb_in_progress = false;
 
+  // After the first reduction completes, we have a way to manually check for
+  // PEs withouth any objects. This flag is set to true if load is ever added
+  // so that the periodic call can be ignored if it comes late. Otherwise, it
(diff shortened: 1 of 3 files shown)


src/ck-ldb/MetaBalancer.C
```diff
@@ -162,6 +162,12 @@ void MetaBalancer::init(void) {
   // to false so that it doesn't clear the handles.
   lb_in_progress = false;
 
+  // After the first reduction completes, we have a way to manually check for
+  // PEs withouth any objects. This flag is set to true if load is ever added
```
evan-charmworks: typo

src/ck-ldb/MetaBalancer.C
```diff
@@ -162,6 +162,12 @@ void MetaBalancer::init(void) {
   // to false so that it doesn't clear the handles.
   lb_in_progress = false;
 
+  // After the first reduction completes, we have a way to manually check for
```
rbuch: Is this supposed to be "we have a way to" or "we have to?"

epmikida: We have a way to. As in, we have a way to do it otherwise, so we can ignore the periodic method.

tests/charm++/load_balancing/meta_lb_test/period_selection.C
```diff
@@ -28,7 +31,9 @@ private:
 public:
   Main(CkArgMsg* msg) : iteration(0), migrations(0) {
     delete msg;
-    arrayProxy = CProxy_TestArray::ckNew(CkNumPes() * 8);
+    // Test currently requires an equal number objects on every PE
```
rbuch:
```suggestion
    // Test currently requires an equal number of objects on every PE
```

juanjgalvez:
I forgot, what's the manual way to check for PEs without any objects after the first reduction?

epmikida:
> I forgot, what's the manual way to check for PEs without any objects after the first reduction?

After load balancing, each MetaLB object checks the LBDatabase for the number of objects on its PE. It also updates state when objects are deleted or created. So the only time it is a problem is at the beginning when there is no "kick" to have it check. The Ccd call after is that kick.

If it ever detects it has 0 objects it also registers that with PE 0 which will ping them when its time for another AtSync. These callbacks are cleared and re-registered after LB.

PR #2363, opened 2019-06-27
Charm4py: support sending same msg copy to multiple PEs in a Group

juanjgalvez:
This essentially exposes CkSendMsgBranchMulti to Charm4py, useful
for list sends, and sending the same msg to children in a spanning
tree.

### DIFF
--- a/src/ck-core/charm.h
+++ b/src/ck-core/charm.h
@@ -405,9 +405,9 @@ extern void CkChareExtSend(int onPE, void *objPtr, int epIdx, char *msg, int msg
 extern void CkChareExtSend_multi(int onPE, void *objPtr, int epIdx, int num_bufs, char **bufs, int *buf_sizes);
 /// Send msg to group with ID 'gid'. if pe == -1, msg will be broadcasted, else
 /// it will go to the group instance in that PE
-extern void CkGroupExtSend(int gid, int pe, int epIdx, char *msg, int msgSize);
+extern void CkGroupExtSend(int gid, int npes, const int *pes, int epIdx, char *msg, int msgSize);
 /// Send msg to group copying data into CkMessage from multiple input buffers
-extern void CkGroupExtSend_multi(int gid, int pe, int epIdx, int num_bufs, char **bufs, int *buf_sizes);
+extern void CkGroupExtSend_multi(int gid, int npes, const int *pes, int epIdx, int num_bufs, char **bufs, int *buf_sizes);
 /// Send msg to array with ID 'aid'. idx is index of destination and ndims the number
 /// of dimensions of the index. If ndims <= 0, msg will be broadcasted to all array elements
 extern void CkArrayExtSend(int aid, int *idx, int ndims, int epIdx, char *msg, int msgSize);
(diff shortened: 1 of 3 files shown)


juanjgalvez:
Changes only affect code inside `#CMK_CHARMPY` blocks.

src/ck-core/charm.h
```diff
@@ -405,9 +405,9 @@ extern void CkChareExtSend(int onPE, void *objPtr, int epIdx, char *msg, int msg
 extern void CkChareExtSend_multi(int onPE, void *objPtr, int epIdx, int num_bufs, char **bufs, int *buf_sizes);
 /// Send msg to group with ID 'gid'. if pe == -1, msg will be broadcasted, else
 /// it will go to the group instance in that PE
-extern void CkGroupExtSend(int gid, int pe, int epIdx, char *msg, int msgSize);
+extern void CkGroupExtSend(int gid, int npes, const int *pes, int epIdx, char *msg, int msgSize);
```
matthiasdiener: Small suggestion: if it is an array, it is sometimes a bit clearer to write as `const int pes[]`.

src/ck-core/ckcallback.C
```diff
@@ -254,7 +255,7 @@ static void CkCallbackSendExt(const CkCallback &cb, void *msg)
     case CkCallback::sendGroup: // Send message to a group element
       CreateCallbackMsgExt(data, dataLen, reducerType, cb.d.group.refnum,
                                   extResultMsgData, extResultMsgDataSizes);
-      CkGroupExtSend_multi(cb.d.group.id.idx, cb.d.group.onPE, cb.d.group.ep,
+      CkGroupExtSend_multi(cb.d.group.id.idx, 1, &(cb.d.group.onPE), cb.d.group.ep,
```
evan-charmworks: I don't think the extra parentheses are necessary, but they don't hurt.

src/ck-core/ck.C
```diff
@@ -2449,10 +2447,12 @@ void CkGroupExtSend_multi(int gid, int pe, int epIdx, int num_bufs, char **bufs,
   CkGroupID gId;
   gId.idx = gid;
 
-  if (pe == -1)
```
evan-charmworks: If most callers of this function have arguments known at compile time that determine which of these three branches they will take, it might be nice to have an API that does not have a runtime branch on compile time information. This would mean `CkGroupExtSend`, `CkGroupExtSendMulti`, `CkGroupExtBroadcast` functions. They could call a static function that does the first half of `CkGroupExtSend_multi` to avoid code duplication. This is very detail-oriented though and not terribly important.

juanjgalvez: Yeah, I was considering doing something like this but fine grained performance testing hasn't shown any difference due to the branching, and I'm not sure how much code duplication it would result in (there are two variants of CkGroupExtSend right now which don't really share the first half of the code). My idea is to do this in the future if it does turn into a bottleneck.

juanjgalvez: Also, the branch is necessary in any case because a group proxy could be for a broadcast or an individual element and the check is done at runtime. What is not necessary is to add `CkSendMsgBranchMulti` to the branch (it could have its own separate function), but like I said I haven't seen any performance regression.

PR #2367, opened 2019-06-27
Remove leftover references to old Windows barrier implementation

rbuch:
I think this should fix the problem, but I still haven't tested, I'll update this when I do.


rbuch:
I was able to build and run a Charm++ program with this change, so I think it fixes the build breakage.

PR #2362, opened 2019-06-27
Propagate main()'s return code in AMPI

matthiasdiener:
AMPI renames the application’s `main` function to `AMPI_Main`.
The return value from `AMPI_Main` is ignored, i.e. the runtime always assumes a return code 0.

Workaround: change `return` to `exit()`.

We need to propagate the real return code throughout AMPI and TCharm.

~~Potential issue:
How to handle `AMPI_Main` without return? E.g. when defined as void main(...), or with implied return 0 at end.
This might then return a random value from the stack.~~

Edit: 
- `void main()` doesn't work with AMPI (even without this PR).
- implied return 0 does actually return 0, so seems to be fine.


matthiasdiener:
This is ready for review.

PR #2372, opened 2019-07-01
ZC Bcast API: Add support to broadcast from non-zero root node

nitbhat:


evan-charmworks:
What implications does `CkLocalBranchOther` have for thread safety?

nitbhat:
> What implications does `CkLocalBranchOther` have for thread safety?

`CkLocalBranchOther` will simply give access to the group's local pointer on another PE. I think right now, it's the user's responsibility to ensure thread safety by protecting writes using atomics or locks. (much like how I have used atomics for `bcastNo`, which is updated by both the worker thread and the comm thread).

stwhite91:
> > What implications does `CkLocalBranchOther` have for thread safety?
> 
> `CkLocalBranchOther` will simply give access to the group's local pointer on another PE. I think right now, it's the user's responsibility to ensure thread safety by protecting writes using atomics or locks. (much like how I have used atomics for `bcastNo`, which is updated by both the worker thread and the comm thread).

It would be good to add comments warning potential users of this function about the need to ensure thread safety when accessing another's group's members

nitbhat:
> > > What implications does `CkLocalBranchOther` have for thread safety?
> > 
> > 
> > `CkLocalBranchOther` will simply give access to the group's local pointer on another PE. I think right now, it's the user's responsibility to ensure thread safety by protecting writes using atomics or locks. (much like how I have used atomics for `bcastNo`, which is updated by both the worker thread and the comm thread).
> 
> It would be good to add comments warning potential users of this function about the need to ensure thread safety when accessing another's group's members

Done.

nitbhat:
This PR needs review and it'll be good to have it merged soon to have it tested in nightly autobuild (since it is release critical).

src/ck-core/ckarray.C
```diff
@@ -1386,16 +1371,49 @@ void CProxy_ArrayBase::ckBroadcast(CkArrayMessage *msg, int ep, int opts) const
...
+
+    // Send back to CkArray on that index
+    thisProxy[srcPe].sendZCBroadcast(zcBcastNo, w);
+  }
+
+void CkArray::incrementBcastNo(){
+  int zcBcastNo = broadcaster->incrementBcastNo();
+}
+
+void CkArray::sendZCBroadcast(int zcBcastNo, MsgPointerWrapper w) {
```
stwhite91: Why does this take zcBcastNo as an argument if it doesn't use it?

juanjgalvez:
The changes affect the existing code for regular broadcasts. I'm wondering, does this add overhead to regular broadcasts if CMK_SMP=1 and CMK_ONESIDED_IMPL=1? And if so, has it been quantified (for example, running a benchmark that does very small broadcasts on a low overhead network layer?)

src/ck-core/ckarray.C
nitbhat: Good point, I'll change this.

nitbhat:
> The changes affect the existing code for regular broadcasts. I'm wondering, does this add overhead to regular broadcasts if CMK_SMP=1 and CMK_ONESIDED_IMPL=1? And if so, has it been quantified (for example, running a benchmark that does very small broadcasts on a low overhead network layer?)

It hasn't been quantified. But yes, converting bcastNo from a regular variable to atomic could add some overhead. I can quantify it on some of the layers to determine if it is okay to merge this change.

juanjgalvez:
Regarding the spanning trees, what this will do is that on every process, the first time it receives a bcast from a new root, it will create the spanning tree for that root (takes O(n) where n is the global number of processes) and store the parent and children for that tree. Subsequent bcasts will use the stored information.

Not sure how much this will affect the performance of the first broadcast compared to subsequent broadcasts. A disparity might be weird. Not sure if we care about that.

The other issue is that this means potentially storing on every process the parent and children of N spanning trees (where N is the number of processes). CmiSpanningTreeInfo is 6 ints per tree, assuming the default branching factor of 4. Not sure how likely it would be that an application does ZC bcasts from every process but I guess we shouldn't be making assumptions about applications. Maybe this is not a big deal for current machines, but intrinsically it is not scalable.

For non-smp mode I assume this code path is also used? If so, since there will be much more processes compared to smp mode this would make this situation worse.

Thoughts?

nitbhat:
I think it's a good idea that it doesn't recompute the spanning tree parent and children for every bcast. I think it's okay to have that disparity between the first bcast and subsequent bcasts. 

The storage issue is interesting and we should think of smarter ways of doing it as this would cause a scaling bottleneck. ( I wonder what other solutions exist)

Yes, the non-smp mode also uses this code path and will scale even poorly. 

Btw, this patch is introducing the bcast from non-zero roots for ZC broadcasts for arrays. Other that this, even currently, groups and nodegroups use the same scheme to broadcast from non-zero roots, right?

juanjgalvez:
> I think it's a good idea that it doesn't recompute the spanning tree parent and children for every bcast. I think it's okay to have that disparity between the first bcast and subsequent bcasts.
> The storage issue is interesting and we should think of smarter ways of doing it as this would cause a scaling bottleneck. ( I wonder what other solutions exist)
> Yes, the non-smp mode also uses this code path and will scale even poorly.
> Btw, this patch is introducing the bcast from non-zero roots for ZC broadcasts for arrays. Other that this, even currently, groups and nodegroups use the same scheme to broadcast from non-zero roots, right?

You are right about the group broadcasts. I just checked the code. I actually wrote that code lol, but I don't remember what was the rationale behind that. I guess I figured, or someone suggested, that using arrays is more common than using groups, and array broadcasts from any PE are more common than group broadcasts from any PE (because arrays have much more elements and the possibility of broadcast sources migrating to other PEs).

nitbhat:
Results from iForge: (The numbers on the 'master' tab shows speedup with atomic `bcastNo` as compared to before, when it was non-atomic)
https://docs.google.com/spreadsheets/d/1Epd_OIDc1rAU0BjkuYqWPEpTRhp-16MjZDdVGLYLSCQ/edit?usp=sharing
 

Results from Stampede2: (The numbers on the 'master' tab shows speedup with atomic `bcastNo` as compared to before, when it was non-atomic) https://docs.google.com/spreadsheets/d/1BnedfmCrSXvbKKeP6s6MjGpC4AUXEsv79pPRFzLc4_U/edit?usp=sharing



Conclusion - There's a lot of variabilities with the results, sometimes the performance is higher and sometimes it's lower. This goes to show that making `bcastNo` atomic is not really impacting the performance. And the difference in results seems to be primarily because of variability across runs. This seems to be confirmed by another experiment in which I ran two iterations of the 'before' case. 

iforge: (The numbers on the 'master' tab shows speedup of 1 run over another, before `bcastNo` was made atomic)https://docs.google.com/spreadsheets/d/13xbamdf5uISO8G2aPQPPSz9XGdKO2et3BRj_Pd7CK1Q/edit?usp=sharing

Stampede2: (The numbers on the 'master' tab shows speedup of 1 run over another, before `bcastNo` was made atomic)
https://docs.google.com/spreadsheets/d/1EfYwUhZ7qe7mfI-12wfQG0yjjXdQ__YL3kwoY7zW_q4/edit?usp=sharing

PR #2374, opened 2019-07-01
 Fix incorrect AtSync barrier check

epmikida:
This is a bug in LB, which had been incorrectly discussed in core in the past. This fix should make its way to the new LB framework as well.

### DIFF
--- a/src/ck-ldb/LBDBManager.C
+++ b/src/ck-ldb/LBDBManager.C
@@ -623,7 +623,6 @@ void LocalBarrier::CheckBarrier()
   if (!on) return;
 
   // If there are no clients, resume as soon as we're turned on
-
   if (client_count == 0) {
     cur_refcount++;
     CallReceivers();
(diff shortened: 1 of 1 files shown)


juanjgalvez:
The changes look good to me, as far as fixing the bug. But it is weird that this bug was there, since the code seems to have been written to handle the case where a client can call AtSync multiple times before the barrier is reached (otherwise it could have been written in a simpler way).

Eric, do you have a test case where it previously failed and it works with your fix?

My main concern is if we really need to handle this case. Do we have a use case for it? For me it seems it would be easier for users to understand if we require a resumeFromSync to happen after an AtSync, and thus it would be an error to call AtSync two times in a row without resumefromsync in between. I'm also wondering if allowing multiple AtSyncs in a row would work well without triggering bugs in other parts of the runtime or application.

Also, the current mechanism to check the local barrier is not very efficient (going through all the clients to see if all of them have reached the barrier). Granted, this only happens when at least `client_count` have called AtSync, but if you have thousands of chares and you can achieve the same with a simple counter that would be better.

PR #2380, opened 2019-07-03
Fix SDAG forall code generation

### LINKED ISSUE #2378: SDAG forall code generation doesn't match forall semantics when using -step
The way code is generated for SDAG forall loops, it always switches bounds and step to make it so you loop from small to large by a positive step size. This breaks for cases like the following:

forall [index] (9:0, -2) { serial { CkPrintf("%i\n", index); } }

Expected output is the odd numbers in [0,9] (technically in any order).
But because generated code sees 9 > 0, it sets lower bound to 0, upper bound to 9, and step to 2. Therefore the actual output is the even numbers in [0,9].

epmikida:
Fixes #2378


epmikida:
> Could add a test for this usage

In the process of overhauling the SDAG tests, which is how I found this bug in the first place. Will be pushing that soon.

PR #2381, opened 2019-07-08
AMPI: Fix handling of MPI_PROC_NULL

matthiasdiener:
In particular, do not clear the all of the status fields, just the ones
required by the standard.

### DIFF
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -3556,11 +3556,16 @@ static inline void clearStatus(MPI_Status sts[], int idx) noexcept {
   }
 }
 
-static inline bool handle_MPI_PROC_NULL(int src, MPI_Comm comm, MPI_Status* sts) noexcept
+// Handle a MPI_PROC_NULL src argument according to Section 3.11 of the MPI-3.1 standard.
+// Relevant for MPI_Recv, MPI_Probe, MPI_Iprobe, MPI_Improbe
+static inline bool handle_MPI_PROC_NULL(int src, MPI_Status* sts) noexcept
 {
   if (src == MPI_PROC_NULL) {
-    clearStatus(sts);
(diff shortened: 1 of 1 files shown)


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -3692,9 +3698,11 @@ void ampi::static_mprobe(ampi *dis, int t, int s, MPI_Comm comm, MPI_Status *sts
 #endif
 }
 
+// Returns whether there is a message that can be received (return 1) or not (return 0) 
 int ampi::iprobe(int t, int s, MPI_Comm comm, MPI_Status *sts) noexcept
```
matthiasdiener: Should this be a static function (as well as improbe)? @evan-charmworks

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -3556,20 +3556,25 @@ static inline void clearStatus(MPI_Status sts[], int idx) noexcept {
...
+      sts->MPI_SOURCE = MPI_PROC_NULL;
+      sts->MPI_TAG = MPI_ANY_TAG;
+      sts->MPI_LENGTH = 0;
+    }
     return true;
   }
   return false;
 }
 
 int ampi::static_recv(ampi *dis, int t, int s, void* buf, int count, MPI_Datatype type, MPI_Comm comm, MPI_Status *sts) noexcept
```
matthiasdiener: static_recv doesn't seem to return anything useful, but its return value appears to be used in various other places (e.g. in intercomm_scatter: `if(-1==recv(...)`). @stwhite91 @evan-charmworks

src/libs/ck-libs/ampi/ampi.C
evan-charmworks: I don't think that is necessary, because they don't have anything like `dis = dis->blockOnRecv();` where the `this` pointer would be invalidated.

PR #2382, opened 2019-07-08
AMPI: make tests/datatype abort on error

stwhite91:


stwhite91:
This test has been silently failing on Issend's since we merged the first AMPI ZC patches a couple months ago.

tests/ampi/datatype/datatype.C
```diff
@@ -506,10 +520,12 @@ void typefree_ialltoall_test(MPI_Comm comm) {
             if (recvbuf[i] != i) {
                 errs++;
                 if (errs < 10) {
-                    printf("recvbuf[%d] = %d, expected %d\n", i, recvbuf[i], i);
+                    printf("Ialltoallv: recvbuf[%d] = %d, expected %d\n", i, recvbuf[i], i);
```
matthiasdiener: should be `ialltoall` (same below)?

PR #2393, opened 2019-07-11
PICS: Fix step counters, complete implementation of more fields, and remove redundant ones

chin123:


src/ck-pics/picsautoperfAPI.h
```diff
@@ -7,8 +7,26 @@
...
+  char **phaseNames;
+  PicsConfig() {
+    fromGlobal = true;
+    numPhases = PICS_INVALID;
+    collectionMode = PICS_INVALID;
+    evaluationMode = PICS_INVALID;
+    warmupSteps = PICS_INVALID;
+    pauseSteps = PICS_INVALID;
+    phaseNames = nullptr;
+  }
```
evan-charmworks: We require C++11 support, so you can safely use default member initializers instead of a constructor method.

src/ck-pics/picsautoperf.C
```diff
@@ -341,6 +342,7 @@ void TraceAutoPerfBOC::setNumOfPhases(int num, const char names[]) {
     strcpy(name, names + i*40);
     CkpvAccess(phaseNames)[i] = name;
   }
+  contribute(0, NULL, CkReduction::nop, cb);
```
evan-charmworks: Would `contribute(cb);` work instead?

src/ck-pics/picsautoperfAPI.h
chin123: Ok, done. Thanks!

src/ck-pics/picsautoperf.C
chin123: Yes, it would. Thanks!

Issue #2375, opened 2019-07-02
++debug has no effect on my application.


marmistrz:
I'm using AMPI with the OpenMPI backend on Arch Linu using `--with-production`, using the package I maintain: https://aur.archlinux.org/packages/ampi-openmpi/

I'm trying to debug the following test program under gdb: https://github.com/marmistrz/heat_solver
I'm using the following cmdline:

```
./charmrun +p 4 ./heat_solver +vp60 --size 7200 --steps 2000 +balancer RotateLB +LBDebug 3 ++debug
```
I'd expect that passing `++debug` would launch an xterm window with gdb. This is not the case. Neither xterm nor gdb is launched. Moreover, no error message is printed to stdout/stderr.

The log until the application starts:
```
Running on 4 processors:  ./heat_solver +vp60 --size 7200 --steps 2000 +balancer RotateLB +LBDebug 3 ++debug 
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./heat_solver +vp60 --size 7200 --steps 2000 +balancer RotateLB +LBDebug 3 ++debug 
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_SINGLE (desired: MPI_THREAD_SINGLE)
CHARMDEBUG> Processor 0 has PID 31049
CHARMDEBUG> Processor 1 has PID 31050
CHARMDEBUG> Processor 2 has PID 31051
CHARMDEBUG> Processor 3 has PID 31052
Charm++> Running in non-SMP mode: 4 processes (PEs)
Converse/Charm++ Commit ID: v6.9.0-0-gc3d50efa1
CharmLB> Verbose level 3, load balancing period: 0.5 seconds
CharmLB> Topology torus_nd_5 alpha: 3.500000e-05s beta: 8.500000e-09s.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> Running on 1 hosts (1 sockets x 6 cores x 2 PUs = 12-way SMP)
Charm++> cpu topology info is gathered in 0.002 seconds.
CharmLB> RotateLB created.
```

matthiasdiener:
I believe `++debug` is only available in the netlrts version of Charm++.

evan-charmworks:
Please see if #2394 implements this suitably.

evan-charmworks:
Implemented by #2394. Since no feedback has been provided, I am considering this complete.

PR #2394, opened 2019-07-11
Feature #2375: Add support for ++debug to charmrun shell scripts

evan-charmworks:


evan-charmworks:
> I also can't imagine how any of these would actually work as expected
Whether they work or not will depend on the environment and particular use. I frequently run `mpirun -np N xterm -e gdb -ex r --args ./myProgram +whatever` with the MPI machine layer successfully. If it works it works, if it doesn't it doesn't, but at least this PR adds support for at least trying.

> other than multicore, which only has one process anyway so doesn't need to launch xterms. 
I did this for consistency. Other machine layers also don't need to be launched with Charmrun if only one PE will be used and launch in an xterm window with just one PE. If you don't want xterm, launch the process with your debugger directly (skipping Charmrun), and this is true for all machine layers that support standalone invocation.

> What would be more useful is documentation of how to run a Charm++ program in a parallel debugger on each platform.
That's outside the scope of the issue this PR resolves, and also not something I can write given no experience with parallel debuggers. I don't want to to block this PR based on this separate issue. If you want, create a new issue and assign it to someone with actual parallel debugger experience.

PR #2398, opened 2019-07-15
DDT: remove using statements

matthiasdiener:


stwhite91:
Why the change? AMPI has the same using statement, so I'd prefer we keep it consistent either way

matthiasdiener:
> Why the change? AMPI has the same using statement, so I'd prefer we keep it consistent either way

ddt.h is now a user-facing header file (due to AMPI/Charm++ interop), so we should probably not have `using` statements in this header. We could put these into ddt.C, but would have then slightly different function prototypes between the header and .C file.

PR #2400, opened 2019-07-15
AMPI: Fix checkBuf for MPI_BOTTOM

matthiasdiener:


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -158,6 +160,7 @@ int errorCheck(const char* func, MPI_Comm comm, bool ifComm, int count,
   }
   if (ifData) {
     ret = checkData(func, data);
+    isAbsolute = getDDT()->getType(data)->getAbsolute();
```
stwhite91: This should go after the possible return statement below it, so if `data` is MPI_DATATYPE_NULL the user gets back MPI_ERR_TYPE

matthiasdiener: Good point, fixed.

PR #2401, opened 2019-07-15
AMPI: replace custom complex classes with std::complex

matthiasdiener:


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -193,72 +193,6 @@ static mpi_comm_worlds mpi_worlds;
...
-    CkAbort("AMPI> Cannot compare long double complex numbers with MPI_MAX\n");
-    return 0;
-  }
-  int operator<(const AmpiLongDoubleComplex &a) noexcept {
-    CkAbort("AMPI> Cannot compare long double complex numbers with MPI_MIN\n");
-    return 0;
-  }
-};
-
 typedef struct { float val; int idx; } FloatInt;
```
stwhite91: Could remove these?

matthiasdiener: but aren't these struct still necessary?

stwhite91: Ah yeah nvm I misread

matthiasdiener:
> ddt.h also has definitions of these complex types that should match the ones here. I think we should switch to std::complex wherever possible, assuming it is compatible with MPI's definition of complex types for the C and Fortran bindings

hmm, are you sure they need to match? currently, there seem to be different definitions already (class with methods vs. struct with just fields).

stwhite91:
> hmm, are you sure they need to match? currently, there seem to be different definitions already (class with methods vs. struct with just fields).

Only the memory layout of the structs/classes need to match with each other (ampi.C and ddt.h), and with the MPI standard. I see now that for the first problem, C++11 does guarantee that `std::complex<T>` has the same memory layout as a struct with two `T` members, so I suppose it's not really an issue though it seems nicer to have them match. I also see now that the MPI standard has language specific bindings to address the second problem for pre-C++11 and for Fortran: MPI_C_FLOAT_COMPLEX, MPI_CXX_FLOAT_COMPLEX, MPI_COMPLEX8, MPI_TYPE_CREATE_F90_COMPLEX, etc.

matthiasdiener:
> > hmm, are you sure they need to match? currently, there seem to be different definitions already (class with methods vs. struct with just fields).
> 
> Only the memory layout of the structs/classes need to match with each other (ampi.C and ddt.h), and with the MPI standard. I see now that for the first problem, C++11 does guarantee that `std::complex<T>` has the same memory layout as a struct with two `T` members, so I suppose it's not really an issue though it seems nicer to have them match. I also see now that the MPI standard has language specific bindings to address the second problem for pre-C++11 and for Fortran: MPI_C_FLOAT_COMPLEX, MPI_CXX_FLOAT_COMPLEX, MPI_COMPLEX8, MPI_TYPE_CREATE_F90_COMPLEX, etc.

Ok, considering that the complex types in ddt are only used to calculate offsets within the structure, I think it's safer to leave them as they are (as opposed to determining the offsets from std::complex).

matthiasdiener:
This needs another review.

PR #2402, opened 2019-07-15
Rewrite SDAG tests

epmikida:
SDAG tests were lacking a lot of coverage, and in some cases didn't actually check the results of the tests. This rewrite tests all SDAG constructs, including advanced usage of refnums, and migration. It also specifically adds a test that caught Issue #2378.


tests/charm++/sdag/migration.ci
```diff
@@ -0,0 +1,129 @@
+// Tests that SDAG works correctly when migrations occur in the middle of an
+// SDAG invocation. Tests both synchronized and anytime migration. Tests that
+// messages received both before an after migration are correctly received, and
```
stwhite91: "an" -> "and"

tests/charm++/sdag/refnum.ci
```diff
@@ -0,0 +1,151 @@
...
+
+    entry void run(int messages) {
+      // Send out the requested number of messages, each with unique refnum
+      // which is set in one of 3 ways:
+      // 1) via the first integer of a parameter marshalled method
+      // 2) via the refnum explicitly set for a reduction
+      // 3) via the refnum explicitly set for a callback
+      serial {
+        // The first third of the messages will be marshalled. The second third
+        // will be via reductions, and the last message will be via callback.
```
stwhite91: "last message" -> "last third"

tests/charm++/sdag/sdag_tests.C
```diff
@@ -0,0 +1,176 @@
...
+
+/* readonly */ CProxy_Main mainProxy;
+/* readonly */ CProxy_BasicTest basicTestProxy;
+/* readonly */ CProxy_RefnumTest refnumTestProxy;
+/* readonly */ CProxy_MigrationTest migrationTestProxy;
+
+class Main : public CBase_Main {
+public:
+  Main(CkArgMsg *m) {
+    mainProxy = thisProxy;
```
stwhite91: m is leaked

epmikida:
One other issue that needs discussion:

Previously, the SDAG migration test was broken with randomized queues. This was due to it having multiple broadcasts and anytime migration. In this new iteration of the tests, there are no broadcasts overlapping with the anytime migration, so it works with randomized queues.

I also have a version that is slightly different, that has bcasts overlapping with anytime migration.

I'm not sure what version to use since I don't know if we came to a consensus on how we are going to restrict broadcast ordering, and what we are doing about anytime migration. It also might be a good idea to have a dedicated test for such a feature, rather than just happening to have it caught up in the SDAG migration tests. But again, that depends on what exactly we decide to support.

stwhite91:
I think having a standalone test for that would be best. It would enable faster development of a potential solution for it

epmikida:
> I think having a standalone test for that would be best. It would enable faster development of a potential solution for it

Agreed. I missed last core meeting, did we actually decide what behavior we want to support?

stwhite91:
We discussed it and I think we said that in Charm v7.0 we don't want to support anytime migration by default, but that we should support unique (not ordered) delivery of broadcast messages in the face of anytime migration when it is used

PR #2405, opened 2019-07-16
ZC API: Allow users to pass switch +noCMAForZC to not use CMA for ZC

nitbhat:
This is useful to test the RDMA code path during autobuild on many machines that support CMA. Additionally, it can be used by users when RDMA performs better than CMA.

### DIFF
--- a/src/ck-core/ckrdma.C
+++ b/src/ck-core/ckrdma.C
@@ -14,6 +14,7 @@
 
 // Integer used to store the ncpy ack handler idx
 static int ncpy_handler_idx;
+bool useCMAForZC;
 
 /*********************************** Zerocopy Direct API **********************************/
 // Get Methods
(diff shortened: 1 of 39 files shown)


src/ck-core/ckrdma.C
```diff
@@ -474,7 +475,7 @@ CkNcpyMode findTransferMode(int srcPe, int destPe) {
   if(CmiNodeOf(srcPe)==CmiNodeOf(destPe))
     return CkNcpyMode::MEMCPY;
 #if CMK_USE_CMA
-  else if(CmiDoesCMAWork() && CmiPeOnSamePhysicalNode(srcPe, destPe))
+  else if(useCMAForZC && CmiDoesCMAWork() && CmiPeOnSamePhysicalNode(srcPe, destPe))
```
stwhite91: It seems potentially nicer to rename CmiDoesCMAWork to CmiIsCMAEnabled() or something and to check useCMAForZC inside of it

nitbhat: `CmiDoesCMAWork` is implemented in the LRTS layer and returns `true` if the machine has support for CMA (after CMA Init). 

I could add a wrapper function `CmiUseCMAForZC` that returns `useCMAForZC && CmiDoesCMAWork`.

stwhite91: I forgot this only disables CMA for ZC, not for everything, so it's not needed

PR #2406, opened 2019-07-17
More tests

epmikida:
Rewrite simplearrayhello (as it didn't really test anything relevant), and add a simplegrouptest to match (as some group functionality seems to have previously been untested).


epmikida:
One other thing I need to do before merge is update the quickstart docs to point to a new example program (it previously sort of pointed to tests/charm++/simplearrayhello). I'm not sure which example I should point to yet though. Probably examples/charm++/hello/1darray, although it's a bit more verbose than it has to be IMO.

Issue #1999, opened 2018-10-18
Bug record/replay may hang in SMP


ericjbohm:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1999*

---
pretty sure tests exist for record replay, but they are either not being run, or are not exercising a lot of code in ck.C

try some experiments with that in the context of coverage analysis to find out what needs to be done to improve measured coverage.

ericjbohm:
Record replay can be shown to hang on current master on netlrts-linux-x86_64-smp using simplearrayhello.

./charmrun ./hello 20 +p8 ++ppn 8 +record
.
.
.


./charmrun ./hello 20 +p8 ++ppn 8 +replay
Charmrun> scalable start enabled. 
Charmrun> started all node programs in 1.378 seconds.
Charm++> Running in SMP mode: 1 processes, 8 worker threads (PEs) + 1 comm threads per process, 8 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: v6.9.0-383-g97572b27c
Warning> Randomization of virtual memory (ASLR) is turned on in the kernel, thread migration may not work! Run 'echo 0 > /proc/sys/kernel/randomize_va_space' as root to disable it, or try running with '+isomalloc_sync'.
Charm++> scheduler running in netpoll mode.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> replay mode.
Charm++> Warning: disabling message integrity detection during replay (requires linking with -memory charmdebug)
Charm++> Running on 1 hosts (1 sockets x 6 cores x 2 PUs = 12-way SMP)
Charm++> cpu topology info is gathered in 0.000 seconds.
Running Hello on 8 processors for 20 elements
[0] Hello 0 created
[0] Hello 1 created
[0] Hello 2 created
[3] Hello 9 created
[6] Hello 16 created
[2] Hello 6 created
[5] Hello 14 created
[1] Hello 3 created
[5] Hello 15 created
[1] Hello 4 created
[6] Hello 17 created
[1] Hello 5 created
[4] Hello 12 created
[2] Hello 7 created
[7] Hello 18 created
[4] Hello 13 created
[7] Hello 19 created
[2] Hello 8 created
[3] Hello 10 created
[3] Hello 11 created
[0] Hi[17] from element 0
[0] CkMessageReplay> Message EP changed during replay org: [0 112 58 136] got: [0 80 58 59]
[...]

PR #2408, opened 2019-07-17
bugfix #1999: record/replay may hang in SMP

ericjbohm:
envelopeEventID is the ckpv that monotonically increases
with each envelope (and therefore message).  However, in SMP
the envelope allocated and enqueued in the initDone process
is invisible to the record/replay process and can occur
on any PE in the node.  That creates a non-deterministic
race condition in the PE:envelopeEventID tuple space.

This workaround simply avoids incrementing the envelopeEventID
on the PE that triggers doneInit. This allows otherwise
deterministic SMP applications to record/replay with proper
fidelity.

### DIFF
--- a/tests/charm++/Makefile
+++ b/tests/charm++/Makefile
@@ -1,17 +1,11 @@
 DIRS = \
-  megatest \
   alignment \
   simplearrayhello \
   provisioning \
-  load_balancing \
-  chkpt \
   delegation \
   queue \
-  sdag \
-  ckAllocSysMsgTest \
   method_templates \
   demand_creation \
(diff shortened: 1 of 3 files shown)


tests/charm++/Makefile
```diff
@@ -1,17 +1,11 @@
 DIRS = \
-  megatest \
```
stwhite91: Changes to to this file seem unintentional

ericjbohm: yeah none of the makefile changes should be in there.  those bled over from a different test branch entirely.

PR #2409, opened 2019-07-18
 ZC non-zero bcast test: Rename identifiers for Windows builds

nitbhat:
Previously, the group was named 'grp1', which was a global namespace
macro on Windows. The group is renamed from 'grp1' to 'grp'. For
consistency, the chare array was renamed from 'arr1' to 'arr' and
the nodegroup was renamed from 'nodegrp1' to 'nodegrp'.

### DIFF
--- a/tests/charm++/zerocopy/bcast_nonzero_root/bcast_nonzero_root.C
+++ b/tests/charm++/zerocopy/bcast_nonzero_root/bcast_nonzero_root.C
@@ -3,9 +3,9 @@
 #define SIZE 129
 #define DEBUG(x) //x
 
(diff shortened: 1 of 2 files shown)


stwhite91:
Didn't we run into the same or similar issue before?

nitbhat:
> Didn't we run into the same or similar issue before?

Yes, we did, for another test. This was a new test I added and I totally forgot about the windows issue and ended up using my favorite identifier for groups - `grp1`.

nitbhat:
This patch fixes the non-smp windows build failures. It's weird how the windows smp builds didn't run into this issue.

PR #2416, opened 2019-07-19
AMPI: fix bug in zero copy buffer deregistration

stwhite91:
AMPI previously kept around the original CkNcpyBuffer objects
used in all zero copy operations. Instead, one should get a
CkNcpyBuffer object back from zero copy completion callbacks
via the CkDataMsg::data field.


stwhite91:
With the updated patch from this morning? I think I just removed the 'delete' showing up there

PR #2420, opened 2019-07-23
Allow printf-style formatting in CkAbort/CmiAbort

matthiasdiener:


src/libs/ck-libs/datatransfer/charm_standalone.C
```diff
@@ -5,8 +5,11 @@
 #include <stdlib.h>
 #include <stdarg.h> /*<- for va_start & friends */
 
-extern "C" void CmiAbort(const char *why) {
-	fprintf(stderr,"Fatal error> %s\n",why);
+extern "C" void CmiAbort(const char *fmt, ...) {
+	fprintf(stderr,"Fatal error>");
+    va_list p; va_start(p, fmt);
```
evan-charmworks: whitespace

matthiasdiener: fixed.

matthiasdiener:
> We should tag CmiAbort (and probably also CmiPrintf and CmiError) with `__attribute__ ((format (printf, 1, 2)))` on GCC and clang so that it is checked for -Wformat-security warnings.
> 
> https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes
> https://clang.llvm.org/docs/AttributeReference.html#format

Agreed, but this seems better in a separate PR?

evan-charmworks:
> Agreed, but this seems better in a separate PR?

Potentially. There might be instances where the string passed to CmiAbort is runtime generated, which was previously perfectly valid, but now that the first argument is passed as a printf format string, could in theory have issues with % characters. This may not be an issue in practice.

matthiasdiener:
> > Agreed, but this seems better in a separate PR?
> 
> Potentially. There might be instances where the string passed to CmiAbort is runtime generated, which was previously perfectly valid, but now that the first argument is passed as a printf format string, could in theory have issues with % characters. This may not be an issue in practice.

Yeah, that's a good point. I changed '%' to '%%' in our source. This might be worth mentioning in the release notes.

evan-charmworks:
In that case, maybe we should add a new abort call that supports printf arguments instead of redefining what the current one is.

matthiasdiener:
> In that case, maybe we should add a new abort call that supports printf arguments instead of redefining what the current one is.

I think we don't need a new call. Having a '%' in the string will only print a slightly incorrect message and should be easy to fix in the application (clang at least will also print a warning).

evan-charmworks:
Something else worth discussing is whether the abort calls should add a trailing newline. Our own tests and examples and even NAMD are not consistent about whether the strings passed in have one or not.

stwhite91:
IMO CmiAbort should add a newline to the string passed to it since having an extra newline is better than not having one at all in an abort statement

matthiasdiener:
> Something else worth discussing is whether the abort calls should add a trailing newline. Our own tests and examples and even NAMD are not consistent about whether the strings passed in have one or not.

Updated the patch with a newline.

PR #2421, opened 2019-07-23
Enable CMK_IMMEDIATE_MSG on mpi-win64 platforms

nitbhat:
This fixes the mpi-win-smp ZC QD test, which was hanging because
the QD counter from the comm thread was not being accounted for in the
QD detection process.

I tested autobuild style 'make test' with this patch and all tests seem to be running fine.

### DIFF
--- a/src/arch/mpi-win64/conv-mach.h
+++ b/src/arch/mpi-win64/conv-mach.h
@@ -40,7 +40,7 @@
 #define CMK_LBDB_ON					   1
 
 #undef  CMK_IMMEDIATE_MSG
-#define CMK_IMMEDIATE_MSG				   0
+#define CMK_IMMEDIATE_MSG				   1
 
 /*
 #define CMK_PCQUEUE_LOCK                                   1


nitbhat:
Do not merge. Planning to test this on windows with the smptest changes.

jcphill:
> This fixes the mpi-win-smp ZC QD test, which was hanging because
> the QD counter from the comm thread was not being accounted for in the
> QD detection process.

Does this mean there is code somewhere that is only correct if CMK_IMMEDIATE_MSG is true? That seems bad and fixable.

I'm guessing the reason that CMK_IMMEDIATE_MSG was false is that for the non-smp MPI layer these messages merely skip the scheduler rather than being driven by an interrupt. I don't know if any of the layers use interrupts anymore. They seem dangerous.

ericjbohm:
Immediate means they get executed by the comm thread, if there is one.  I don’t think we have any semantics that support interrupting messages.

nitbhat:
The travis failure is because of a hang in Stencil3d (with DistributedLB). I'll update the branch and have the tests run again. . 

Yes, when QD counters are modified on the comm thread, it requires `CMK_IMMEDIATE_MSG` to be true to contribute its count. Otherwise, QD will be not be reached. The ZC API in the SMP mode modifies the QD counters on the comm thread because that is where completion is detected. There is a workaround where I can send a message to the first node to update the counters. But, I'm wondering if that change is worth it considering that we'll be inclined to keep `CMK_IMMEDIATE_MSG` enabled for mpi-win-smp. 

I tested this commit with the SMP testing patch by @epmikida and it successfully passed on Windows.

ericjbohm:
My vague recall of the original disabling is that there was a rare hanging problem with NAMD on mpi-smp.  So I suggest someone try a few NAMD experiments at various scales with this patch.

nitbhat:
I'll test NAMD with this patch for mpi-win-SMP.

nitbhat:
Except for a file open error (to write the output) at the end of the simulation, I was able to run NAMD with apoa1 on cs-dexterity about 4 times with the command `mpirun -n 4 ./namd2.exe ../apoa1/apoa1.namd ++ppn 1 +pemap 0-3 +commap 4-7`. This test took about 1.5-2 hours to run to completion. Is there any other test I can run to validate the `CMK_IMMEDIATE_MSG` change? 

Also, are there bigger Windows HPC clusters where larger tests can run faster?

jcphill:
5 s/step is horribly slow.

Issue #2426, opened 2019-07-25
Megampi crashes on netlrts-linux-x86_64-bigsim


nitbhat:
```
make[3]: Entering directory '/scratch/autobuild/bluegene/charm/netlrts-linux-x86_64-bigsim/tests/ampi/megampi'
../../../bin/ampicxx  -DBIGSIM_TEST=1 -o bg_test.o -c test.C
../../../bin/ampicxx  -o bg_pgm bg_test.o -balancer RandCentLB
../../../bin/testrun  ./bg_pgm +p1 +vp1 +x1 +y1 +z1  ++local ++no-va-randomization
Charmrun> scalable start enabled. 
Charmrun> started all node programs in 0.027 seconds.
Charm++> Running in non-SMP mode: 1 processes (PEs)
Converse/Charm++ Commit ID: 96e1466
Charm++> scheduler running in netpoll mode.
BG info> Simulating 1x1x1 nodes with 1 comm + 1 work threads each.
BG info> Network type: bluegene.
alpha: 1.000000e-07	packetsize: 1024	CYCLE_TIME_FACTOR:1.000000e-03.
CYCLES_PER_HOP: 5	CYCLES_PER_CORNER: 75.
BG info> cpufactor is 1.000000.
BG info> floating point factor is 0.000000.
BG info> Using WallTimer for timing method. 
CharmLB> Load balancer ignores processor background load.
CharmLB> Load balancer assumes all CPUs are same.
Trace: traceroot: ./bg_pgm
CharmLB> RandCentLB created.
[0] Testing: Testing...
Charmrun> error on request socket to node 0 '127.0.0.1'--
Socket closed before recv.
Makefile:34: recipe for target 'bgtest' failed
make[3]: Leaving directory '/scratch/autobuild/bluegene/charm/netlrts-linux-x86_64-bigsim/tests/ampi/megampi'
make[3]: *** [bgtest] Error 1
Makefile:48: recipe for target 'bgtest-megampi' failed
make -C stacksize bgtest OPTS='' TESTOPTS='++local ++no-va-randomization'
make[2]: *** [bgtest-megampi] Error 2
```

evan-charmworks:
Fixed by #2428

PR #2428, opened 2019-07-25
BigSim #2426: fix crash in megampi test

stwhite91:


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -5960,7 +5960,7 @@ CMI_WARN_UNUSED_RESULT ampiParent* IReq::wait(ampiParent* parent, MPI_Status *st
       complete = true;
       parent->resumeOnRecv = false;
 #if CMK_BIGSIM_CHARM
-      *result = 0;
+      if (result) *result = 0;
```
matthiasdiener: Could maybe add to `int* result /*=nullptr*/` to the function prototype.

stwhite91: Done

evan-charmworks:
Fixes #2426

PR #2434, opened 2019-07-26
Add smptest target for AMPI tests

epmikida:


epmikida:
megampi currently fails when running with `+p4 ++ppn 2 +vp4` with the following error: `Recv'ed bad value for large message send/recv`

I'm not sure if this is an issue with the test, the test configuration, or a bug in AMPI, but I figured I'd put the PR up for discussion.

epmikida:
@evan-charmworks @SamTWhite This is another AMPI related PR that should probably be discussed by the AMPI subgroup as to its relevance. Not sure if something else ended up fixing the issues we were seeing here or not.

PR #2435, opened 2019-07-26
cmake: allow passing extra options to compiler

matthiasdiener:


matthiasdiener:
> Is there documentation on using cmake to build charm anywhere?

here: https://charm.readthedocs.io/en/latest/charm++/manual.html#installation-with-cmake

Issue #2445, opened 2019-07-30
+setcpuaffinity fails on Bridges for SMP builds


nitbhat:
```[nbhat4@r005 megatest]$ make test OPTS="" TESTOPTS="+setcpuaffinity"
../../../bin/testrun  ./pgm +p1  +setcpuaffinity

Running on 1 processors:  ./pgm +setcpuaffinity
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 1  ./pgm +setcpuaffinity
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 1 processes, 1 worker threads (PEs) + 1 comm threads per process, 1 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: v6.9.0-508-gd537163
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled.
Error: CmiSetCPUAffinity failed to bind PE #0 to PU #0.
[0] Stack Traceback:
  [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0xc1  [0x81256e]
  [0:1] CmiGetNonLocal+0  [0x812648]
  [0:2]   [0x832ede]
  [0:3] CmiHandleMessage+0xb2  [0x816c3d]
  [0:4] CmiDeliverSpecificMsg+0xcf  [0x81743a]
  [0:5] CmiInitCPUAffinity+0x7ea  [0x83431b]
  [0:6] _Z10_initCharmiPPc+0x1688  [0x6fc797]
  [0:7]   [0x81225b]
  [0:8]   [0x80e503]
  [0:9] +0x7e25  [0x7ff87fbc6e25]
  [0:10] clone+0x6d  [0x7ff87d34634d]
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: CmiSetCPUAffinity failed!
Abort(1) on node 0 (rank 0 in comm 496): application called MPI_Abort(comm=0x84000002, 1) - process 0

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   RANK 0 PID 26266 RUNNING AT r005
=   EXIT STATUS: 1
===================================================================================
make: *** [test] Error 1```

I tried mpi-smp and ofi-smp and both show the same error.

[...]

PR #2448, opened 2019-07-30
Bug #2445: Use hwloc logical PU indices during cpuAffinityRecvHandler

evan-charmworks:


src/conv-core/cpuaffinity.C
```diff
@@ -241,6 +241,47 @@ int CmiSetCPUAffinity(int mycore)
...
 
+int CmiSetCPUAffinityLogical(int mycore)
+{
+  int core = mycore;
+  if (core < 0) {
+    core = CmiHwlocTopologyLocal.num_pus + core;
+  }
+  if (core < 0) {
+    CmiError("Error: Invalid parameter to CmiSetCPUAffinityLogical: %d\n", mycore);
+    CmiAbort("CmiSetCPUAffinityLogical failed!");
```
stwhite91: Could remove this since it's covered in the error message above

nitbhat:
On ofi-smp, the code doesn't crash, but on passing `+setcpuaffinity`, it looks like all threads are being mapped to the same PE. 

```Running on 2 processors:  ./pgm +setcpuaffinity +showcpuaffinity
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 2  ./pgm +setcpuaffinity +showcpuaffinity
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 2 processes, 1 worker threads (PEs) + 1 comm threads per process, 2 PEs total
Converse/Charm++ Commit ID: v6.9.0-510-gd7489d7
HWLOC> [0] Thread 140670688073472 bound to cpuset: 0x00000002
[2] thread CPU affinity mask is 0x00000002
[0] thread CPU affinity mask is 0x00000002
[3] thread CPU affinity mask is 0x00000002
[1] thread CPU affinity mask is 0x00000002
Charm++> Running on 1 hosts
Charm++> cpu topology info is gathered in 0.350 seconds.
WARNING: Multiple PEs assigned to same core, recommend adjusting processor affinity or passing +CmiSleepOnIdle to reduce interference.
Megatest is running on 2 nodes 2 processors.
test 0: initiated [groupring (milind)]```

**Comment by nitbhat** [comment]
And explicitly specifying `++ppn, +pemap, +commap` seems to cause a crash because of unsuccessful binding. 

```[nbhat4@r005 megatest]$ make test OPTS="" TESTOPTS=" ++ppn 1 +pemap 0-11 +commap 12-20 +showcpuaffinity"
../../../bin/testrun  ./pgm +p1  ++ppn 1 +pemap 0-11 +commap 12-20 +showcpuaffinity

Running on 1 processors:  ./pgm +ppn 1 +pemap 0-11 +commap 12-20 +showcpuaffinity
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 1  ./pgm +ppn 1 +pemap 0-11 +commap 12-20 +showcpuaffinity
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 1 processes, 1 worker threads (PEs) + 1 comm threads per process, 1 PEs total
Converse/Charm++ Commit ID: v6.9.0-510-gd7489d7
Charm++> cpuaffinity PE-core map : 0-11
Charm++> set PE 0 on node 0 to core #0
[0] Stack Traceback:
  [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0xc1  [0x8125ae]
  [0:1] CmiGetNonLocal+0  [0x812688]
  [0:2] CmiInitCPUAffinity+0x633  [0x8342c7]
  [0:3] _Z10_initCharmiPPc+0x1688  [0x6fc7d7]
  [0:4]   [0x81229b]
  [0:5]   [0x80e543]
  [0:6] +0x7e25  [0x7ff87fbc6e25]
  [0:7] clone+0x6d  [0x7ff87d34634d]
Error: CmiSetCPUAffinity failed to bind PE #0 to PU #0.
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: CmiSetCPUAffinity failed!
Abort(1) on node 0 (rank 0 in comm 496): application called MPI_Abort(comm=0x84000002, 1) - process 0

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   RANK 0 PID 30153 RUNNING AT r005
=   EXIT STATUS: 1
===================================================================================
make: *** [test] Error 1```

**Comment by evan-charmworks** [comment]
How many cores are provisioned when the job is launched? It may be that only one is available so all PEs are mapped to it.

I would expect `+pemap` involving index 0 to fail by design, given that it uses absolute OS indices and index 0 is unavailable to the job.

**Comment by nitbhat** [comment]
> How many cores are provisioned when the job is launched? It may be that only one is available so all PEs are mapped to it.
> 
> I would expect `+pemap` involving index 0 to fail by design, given that it uses absolute OS indices and index 0 is unavailable to the job.

I was allocating an entire node with 28 cores.

**Comment by evan-charmworks** [comment]
Interesting. I don't believe there is a bug in this patch that would cause all PEs to bind to a single PU, because I don't see that behavior on other machines. What does hwloc's `lstopo` utility report on Bridges when run in a job launch context?

**Comment by nitbhat** [comment]
Ah, I see. The mistake was on my end. When `interact` allocates, turns out the default is just 1 core, unless specified. (just saw this in the documentation). If I specify more cores, `+setcpuaffinity` works fine. Sorry for the confusion.

File: src/conv-core/cpuaffinity.C (hunk 1 above)
Comment (evan-charmworks) [reply src/conv-core/cpuaffinity.C]: Most of this function is copied from `CmiSetCPUAffinity`, including these lines. There are plenty of locations throughout the runtime that could now have multiple print calls coalesced into a single CmiAbort. For consistency, we could leave this as it is for now, and revisit it as part of broader cleanup.

**Comment by nitbhat** [comment]
Why was this failing on Bridges and not on other machines? (Any idea? i.e. was `cmi_hwloc_get_pu_obj_by_os_index` failing previously)

**Comment by evan-charmworks** [comment]
The hwloc documentation states, OS indexes "may be completely arbitrary, non-unique, non-contiguous, not representative of logical proximity, and may depend on the BIOS configuration." If only one core was being allocated, it may not have had OS index 0, therefore any attempted use of that index would be invalid.

File: src/conv-core/cpuaffinity.C (hunk 1 above)
Comment (nitbhat) [reply src/conv-core/cpuaffinity.C]: > The hwloc documentation states, OS indexes "may be completely arbitrary, non-unique, non-contiguous, not representative of logical proximity, and may depend on the BIOS configuration." If only one core was being allocated, it may not have had OS index 0, therefore any attempted use of that index would be invalid.

Right, I understand that for `interact` which was only allocating 1 core. But the autobuild job that failed with `+setcpuaffinity` had the following configuration:
```
 30 #!/bin/sh$
 31 #SBATCH -p RM-small$
 32 #SBATCH -N 2$
 33 #SBATCH --ntasks-per-node 4$
 34 #SBATCH -t 01:20:00$
 35 #SBATCH -o $output$
 36 #SBATCH -e $error$
```

**Comment by nitbhat** [comment]
> The hwloc documentation states, OS indexes "may be completely arbitrary, non-unique, non-contiguous, not representative of logical proximity, and may depend on the BIOS configuration." If only one core was being allocated, it may not have had OS index 0, therefore any attempted use of that index would be invalid.

Right, I understand that for `interact` which was only allocating 1 core. But the autobuild job that failed with `+setcpuaffinity` had the following configuration:
```
 30 #!/bin/sh$
 31 #SBATCH -p RM-small$
 32 #SBATCH -N 2$
 33 #SBATCH --ntasks-per-node 4$
 34 #SBATCH -t 01:20:00$
 35 #SBATCH -o $output$
 36 #SBATCH -e $error$
```

evan-charmworks:
Even if all cores were available, OS indexes are arbitrary and index 0 could be entirely meaningless. You could use a debugger to break inside the `if (thread_obj != nullptr)` block and print `thread_obj->os_index` to see what a valid value is.

Issue #2446, opened 2019-07-30
Record/replay segfaults when built with --disable-tracing


stwhite91:
```
   frame #0: 0x00000001000ef491 pgm`openReplayFile(prefix="ckreplay_", suffix=".log", permissions="w") at ck.C:2543:23
   2540
   2541 #if CMK_REPLAYSYSTEM
   2542 static FILE *openReplayFile(const char *prefix, const char *suffix, const char *permissions) {
-> 2543   std::string fName = CkpvAccess(traceRoot);
   2544   fName += prefix;
   2545   fName += std::to_string(CkMyPe());
   2546   fName += suffix;
  thread #3, stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
    frame #0: 0x00000001000ef491 pgm`openReplayFile(prefix="ckreplay_", suffix=".log", permissions="w") at ck.C:2543:23
   2540
   2541 #if CMK_REPLAYSYSTEM
   2542 static FILE *openReplayFile(const char *prefix, const char *suffix, const char *permissions) {
-> 2543   std::string fName = CkpvAccess(traceRoot);
   2544   fName += prefix;
   2545   fName += std::to_string(CkMyPe());
   2546   fName += suffix;
```

ericjbohm:
Record replay depends on parts of the tracing framework in a variety of places. So I think the right answer here is to reject attempts to compile charm that way.

matthiasdiener:
Fixed via #2449.

PR #2449, opened 2019-07-30
Bug #2446: disable record/replay support when tracing is disabled

stwhite91:


matthiasdiener:
this will need changes to `cmake/detect-features.cmake` as well.

stwhite91:
> this will need changes to `cmake/detect-features.cmake` as well.

I don't see any similar features there that depend on other features. Does the cmake system have an equivalent to --with-production?

matthiasdiener:
> > this will need changes to `cmake/detect-features.cmake` as well.
> 
> I don't see any similar features there that depend on other features. Does the cmake system have an equivalent to --with-production?

Yeah, this isn't so well supported (and difficult to replicate in cmake). My suggestion is to add this code to line 108 in the main CMakeLists.txt:

```cmake
set(CMK_REPLAYSYTEM ${REPLAY})

if(${REPLAY} AND NOT ${TRACING})
  set(CMK_REPLAYSYTEM 0)
  message(WARNING "Charm record/replay is disabled because tracing is disabled")
endif()
```

PR #2450, opened 2019-07-30
ROMIO: Remove 'P' from PMPI_File_set_errhandler in adio/common/ad_end.c

evan-charmworks:


evan-charmworks:
Fixes a linking error on Darwin in any program that uses ROMIO.

matthiasdiener:
Is there a `PMPI_File_set_errhandler` call in mpi-io/close.c too?

evan-charmworks:
Looks like there is. I'm not sure why that did not cause a link failure. #2451

PR #2453, opened 2019-08-01
Add CmiPuts wrapper around CmiPrintf for charm4py

evan-charmworks:


juanjgalvez:
Since this is just for Charm4py, maybe can put inside the existing CMK_CHARMPY blocks, if we feel this is not needed for other cases. But I'm fine with this.

PR #2454, opened 2019-08-01
Only run tests with +noCMAForZC if CMK_USE_CMA is true

evan-charmworks:


evan-charmworks:
@nitbhat Since your machine supports CMA, could you test that this patch functions correctly?

nitbhat:
Yes, testing it now.

nitbhat:
On my CMA enabled machine, I do see `checking "whether has support for shm transport using Cross Memory Attach"... "yes"` while building. However, `CMK_USE_CMA` is `0` in `conv-mach-opt.mak`, causing `+noCMAForZC` option tests to not run.

nitbhat:
I tested it on `courage` (a lab machine). If you have access, you can test it on that.

evan-charmworks:
Sorry about that, should be fixed now.

examples/charm++/zerocopy/entry_method_post_api/prereg/nodegroupTest/Makefile
```diff
@@ -21,6 +21,7 @@ else
 	$(call run, +p4 ./nodegroupTest)
 	$(call run, +p4 ./nodegroupTest +noCMAForZC)
 endif
+endif
```
matthiasdiener: this additional `endif` seems spurious.

PR #2455, opened 2019-08-01
Add CmiAbortHook wrapper around CmiAbort for charm4py

juanjgalvez:


src/ck-core/ck.C
```diff
@@ -2203,6 +2203,7 @@ void registerArrayMapProcNumExtCallback(int (*cb)(int, int, const int *)) {
 
 int CkMyPeHook() { return CkMyPe(); }
 int CkNumPesHook() { return CkNumPes(); }
+void CmiAbortHook(const char *msg) { CmiAbort(msg); }
```
matthiasdiener: Might be better to do CmiAbort(“%s”, msg)

evan-charmworks: I agree

matthiasdiener:
Can you comment briefly why this patch is necessary for charm4py?

juanjgalvez:
> Can you comment briefly why this patch is necessary for charm4py?

I don't write C code directly in Charm4py. I support 3 different interfaces to access the shared library, and they all expect to find a C function in libcharm.so with signature `CmiAbort(const char*)`. I don't think function attributes are supported in any of them. If there is some support for it, I have no idea how it works.

juanjgalvez:
> Can you comment briefly why this patch is necessary for charm4py?
> 
> I don't write C code directly in Charm4py. I support 3 different interfaces to access the shared library, and they all expect to find a C function in libcharm.so with signature CmiAbort(const char*). I don't think function attributes are supported in any of them. If there is some support for it, I have no idea how it works.

Also, aren't function attributes a compile time thing? In some cases, like ctypes, there isn't even any compilation.

matthiasdiener:
> > Can you comment briefly why this patch is necessary for charm4py?
> 
> I don't write C code directly in Charm4py. I support 3 different interfaces to access the shared library, and they all expect to find a C function in libcharm.so with signature `CmiAbort(const char*)`. I don't think function attributes are supported in any of them. If there is some support for it, I have no idea how it works.

But can’t you change the signature that the interfaces expect to CmiAbort(const char *fmt, ...) ?

juanjgalvez:
> Can you comment briefly why this patch is necessary for charm4py?
> 
> I don't write C code directly in Charm4py. I support 3 different interfaces to access the shared library, and they all expect to find a C function in libcharm.so with signature CmiAbort(const char*). I don't think function attributes are supported in any of them. If there is some support for it, I have no idea how it works.
> 
> But can’t you change the signature that the interfaces expect to CmiAbort(const char *fmt, ...) ?

Possibly. I'll need some time to test this, since there are 3 different interfaces and they work differently. Will have to test on Windows too. If it works I guess we can just revert these patches.

PR #2457, opened 2019-08-02
CkIO: Use default stripe size when automatic detection errors

### LINKED ISSUE #2456: CkIO Error Getting Lustre Stripe Size on Blue Waters
@trquinn reported the following:

> I'm getting the following abort using the CkIO on a Lustre file system.
> 
> This is for charm version
> Converse/Charm++ Commit ID: v6.9.0-300-g99cda7a11
> 
> Any ideas to fix this?
> 
> CkIO llapi error -61 on h516.cosmo25cmb.3072g1wsbHK1.002080.rung
> ------------- Processor 0 Exiting: Called CmiAbort ------------
> Reason: [CkIO] llapi_file_get_stripe error
> [0] Stack Traceback:
>    [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0x5e  [0x817a83]
>    [0:1]   [0x817af3]
>    [0:2] CkGetFileStripeSize+0x59  [0x722faf]
>    [0:3] _ZN2Ck2IO4impl8Director8openFileENSt7__cxx1112basic_stringIcSt11char_tra
> itsIcESaIcEEE10CkCallbackNS0_7OptionsE+0x4b3  [0x7212fd]
>    [0:4] _ZN2Ck2IO4impl16CkIndex_Director24_call_openFile_marshall2EPvS3_+0x1e6
> [0x717034]

rbuch:
Fixes #2456

### DIFF
--- a/src/libs/ck-libs/io/fs_parameters.c
+++ b/src/libs/ck-libs/io/fs_parameters.c
@@ -43,7 +43,8 @@ size_t CkGetFileStripeSize(const char *filename) {
   }
 
   if (rc != 0) {
-    CkAbort("[CkIO] llapi_file_get_stripe error");
+    CkPrintf("[CkIO] Cannot extract lustre file stripe size for %s, using default of 4 MB\n", filename);
+    return 4 * 1024 * 1024;
   }
 
   return lump->lmm_stripe_size;


src/libs/ck-libs/io/fs_parameters.c
```diff
@@ -43,7 +43,8 @@ size_t CkGetFileStripeSize(const char *filename) {
   }
 
   if (rc != 0) {
-    CkAbort("[CkIO] llapi_file_get_stripe error");
+    CkPrintf("[CkIO] Cannot extract lustre file stripe size, using default of 4 MB\n");
```
stwhite91: Is this function called on every PE, or only PE 0? If the former we need to limit output to PE 0 only.

rbuch: This is called once per `Ck::IO::open` invocation if the write stripe size has not been specified. This is just a raw function call, so it should execute once on whatever PE makes the call (it kicks off some messaging internally, but that's not relevant to this patch).

matthiasdiener: Would it make sense to run this detection only once per execution of the application?

rbuch: No, this detects the stripe settings for the particular file to be written; different files in different parts of the file system can have different stripe settings set, so this should ideally be dynamic.

matthiasdiener: Ok, in that case, it might make sense to also print the filename of the file for which the detection fails.

rbuch: Good idea, I've just done that.

PR #2458, opened 2019-08-02
Add a pup_buffer API with ZC functionality

nitbhat:


src/util/pup.h
```diff
@@ -126,6 +127,37 @@ typedef enum {
   dataType_last //<- for setting table lengths, etc.
 } dataType;
 
+template <class T>
+dataType getXlateDataType(T *a) {
+  if(std::is_same<T, signed char>::value) return Tchar;
```
stwhite91: Maybe better to use a switch/case here

tests/charm++/pup_buffer/pup_buffer.C
```diff
@@ -0,0 +1,82 @@
...
+    arr(CkMigrateMessage *msg) {
+      delete msg;
+    }
+
+    static void *buff_allocate(size_t n) {
+      return new int[n];
+    }
+
+    static void buff_deallocate(void *ptr) {
+      free(ptr);
```
stwhite91: Should use delete here or malloc above, otherwise valgrind will warn about mismatched new/free

tests/charm++/pup_buffer/Makefile
```diff
@@ -0,0 +1,31 @@
...
+	$(CHARMC) -c pup_buffer.C
+
+test: all
+	$(call run, +p1 ./pup_buffer +balancer RotateLB)
+	$(call run, +p4 ./pup_buffer +balancer RotateLB)
+
+smptest: all
+	$(call run, +p2 ./pup_buffer ++ppn 2 +balancer RotateLB)
+	$(call run, +p4 ./pup_buffer ++ppn 2 +balancer RotateLB)
+	$(call run, +p2 ./pup_buffer +noCMAForZC ++ppn 2 +balancer RotateLB)
```
stwhite91: if/when Evan's patch for conditionally running these is merged, should use that

matthiasdiener: It's already merged (#2454 )

src/util/pup.h
nitbhat: I am wondering if I can use `is_same` expression in a switch case because the expression evaluates to boolean and not distinct values that I can compare in the switch statement.

tests/charm++/pup_buffer/pup_buffer.C
nitbhat: Changed free to delete []

tests/charm++/pup_buffer/Makefile
nitbhat: Removed the `+noCMAForZC` run commands because this currently doesn't use ZC internally. Will add back these run statements with the CMA makefile clause, when I internally use ZC.

tests/charm++/pup_buffer/pup_buffer.C
```diff
@@ -0,0 +1,82 @@
...
+  CkCallback cb;
+  int iteration;
+  public:
+    arr() {
+      buffer = new int[SIZE];
+      iteration = 0;
+
+      usesAtSync = true;
+
+      for(int i=0; i < SIZE; i++) buffer[i] = thisIndex;
```
matthiasdiener: Maybe put a line break before `buffer[I]`

tests/charm++/pup_buffer/pup_buffer.C
```diff
@@ -0,0 +1,82 @@
...
+    }
+};
+
+class arr : public CBase_arr {
+  int *buffer;
+  CkCallback cb;
+  int iteration;
+  public:
+    arr() {
+      buffer = new int[SIZE];
```
matthiasdiener: Is this deleted somewhere?

src/util/pup.h
```diff
@@ -208,85 +240,30 @@ class er {
...
-  void operator()(unsigned CMK_PUP_LONG_LONG *a,size_t nItems)
-    {bytes((void *)a,nItems,sizeof(unsigned CMK_PUP_LONG_LONG),Tulonglong);}
-#endif
-#if CMK_HAS_INT16
-  void operator()(CmiInt16 *a,size_t nItems)
-    {bytes((void *)a,nItems,sizeof(CmiInt16),Tint128);}
-  void operator()(CmiUInt16 *a,size_t nItems)
-    {bytes((void *)a,nItems,sizeof(CmiUInt16),Tuint128);}
-#endif
+  template<class T>
```
matthiasdiener: Maybe put a comment here that this is the same as above but with a custom allocator, deallocator

src/util/pup.h
```diff
@@ -126,6 +127,37 @@ typedef enum {
...
+  else if(std::is_same<T, bool>::value) return Tbool;
+#ifdef CMK_PUP_LONG_LONG
+  else if(std::is_same<T, CMK_PUP_LONG_LONG>::value) return Tlonglong;
+  else if(std::is_same<T, unsigned CMK_PUP_LONG_LONG>::value) return Tulonglong;
+#endif
+#if CMK_HAS_INT16
+  else if(std::is_same<T, CmiInt16>::value) return Tint128;
+  else if(std::is_same<T, CmiUInt16>::value) return Tuint128;
+#endif
+  CmiAbort("getXlateDataType: Matching primitive data type not found in dataType enum");
```
matthiasdiener: Maybe also print `typeid(*a).name()` here

nitbhat:
This branch requires a significant amount of cleanup. Currently, it's not ready for review.

nitbhat:
This branch is ready for review.

evan-charmworks:
I tested the pup_buffer API with my Isomalloc redesign and the pointer-swapping aspect of within-node migration works correctly. For inter-node migration on netlrts I am encountering an issue where control flow will be passed back to the AMPI rank after the `allocate` function is called but before the data transfer happens, causing a crash due to the call stack and heap being all zeros.

I had one other review comment here: https://github.com/UIUC-PPL/charm/pull/2458/files#r323943139

src/ck-core/cklocation.C
```diff
@@ -3026,8 +3039,17 @@ void CkLocMgr::immigrate(CkArrayElementMigrateMessage *msg)
...
 	
+	envelope *env = UsrToEnv(msg);
+	CmiAssert(CkpvAccess(newZCPupGets).empty()); // Ensure that vector is empty
 	//Create the new elements as we unpack the message
 	pupElementsFor(p,rec,CkElementCreation_migrate);
+	if(!CkpvAccess(newZCPupGets).empty()) {
+		// newZCPupGets is not empty, rgets need to be launched
+		// newZCPupGets is populated with NcpyOperationInfo during pupElementsFor by pup_buffer calls that require Rgets
+		// Issue Rgets using the populated newZCPupGets vector
+		zcPupIssueRgets(msg->id, this);
```
stwhite91: What happens if the rgets don't complete inline (eg the migration is across the network)? Where do we handle the case where completion is asynchronous?

nitbhat:
> I tested the pup_buffer API with my Isomalloc redesign and the pointer-swapping aspect of within-node migration works correctly. For inter-node migration on netlrts I am encountering an issue where control flow will be passed back to the AMPI rank after the `allocate` function is called but before the data transfer happens, causing a crash due to the call stack and heap being all zeros.
> 
> I had one other review comment here: https://github.com/UIUC-PPL/charm/pull/2458/files#r323943139

Ah, that's interesting. I think there's potentially a bug that is not being captured by the ZC Pup buffer test. Can you send me the steps to replicate this failure? I'll try debugging it.

src/ck-core/cklocation.C
stwhite91: I think this is the issue Evan is seeing

evan-charmworks:
> Ah, that's interesting. I think there's potentially a bug that is not being captured by the ZC Pup buffer test. Can you send me the steps to replicate this failure? I'll try debugging it.

I just pushed a testing branch that combines your ZC PUP work and my Isomalloc redesign: `evan/isomalloc_zc_pup_test`. This command should reproduce the problem: `./build AMPI netlrts-linux-x86_64 -j8 -g3 && pushd netlrts-linux-x86_64/benchmarks/ampi/isomalloc && make OPTS="-g3" os-isomalloc -j8 && ./charmrun ./os-isomalloc +p2 +vp2 +isomalloc_sync +balancer RotateLB ++local`

src/ck-core/cklocation.C
nitbhat: Right, for asynchronous Gets (Rgets, which don't complete inline), in Charm++, I just buffer the messages to the target chares to not allow those messages to execute since the data is not ready. It's only after the completion of the Rgets in `zcPupGetCompleted` that the buffered messages are executed because the data is ready. 

However, I see that for `isomalloc`, it looks like you'd want a capability where the pup (with `isUnpacking` instance) function is not called until the data transfer is complete. Currently, the `pup` function is called during unpacking (just as before) because we'd want the input parameter `buffer`(the pointer) that the user wants the destination region to be assigned to.

evan-charmworks: The `isUnpacking` PUP::er would still need to be called before the data transfer so that the receiver knows what the proper `allocate` function is. The problem is that execution of the TCharm chare resumes before the data transfer completes.

nitbhat:
> > Ah, that's interesting. I think there's potentially a bug that is not being captured by the ZC Pup buffer test. Can you send me the steps to replicate this failure? I'll try debugging it.
> 
> I just pushed a testing branch that combines your ZC PUP work and my Isomalloc redesign: `evan/isomalloc_zc_pup_test`. This command should reproduce the problem: `./build AMPI netlrts-linux-x86_64 -j8 -g3 && pushd netlrts-linux-x86_64/benchmarks/ampi/isomalloc && make OPTS="-g3" os-isomalloc -j8 && ./charmrun ./os-isomalloc +p2 +vp2 +isomalloc_sync +balancer RotateLB ++local`

I was initially thinking that there was a bug in the implementation where one of the messages to the target chares (involved in a ZC pup) is going through even though the Rget is incomplete. However, I think this issue is more fundamental and related to the basic design of ZC pup. Based on what you mentioned, we'd want a scheme to not execute the `pup` function until the data is ready. 

Alternatively, is there a way you can restructure isomalloc to not expect data immediately in the `pup` method, but when a "dummy" entry method is executed? (since the "dummy" entry method will only be executed when the data is ready i.e when the rgets complete). 

Also, can you test `isomalloc` in a CMA enabled machine and verify that it's working correctly? (Do not pass `+noCMAForZC` in the run since that switches to RDMA).

src/ck-core/cklocation.C
nitbhat: Okay, so, you're saying an entry method of TCharm resumes before the data transfer completes?

evan-charmworks:
> However, I think this issue is more fundamental and related to the basic design of ZC pup. Based on what you mentioned, we'd want a scheme to not execute the `pup` function until the data is ready.
> 
> Alternatively, is there a way you can restructure isomalloc to not expect data immediately in the `pup` method, but when a "dummy" entry method is executed? (since the "dummy" entry method will only be executed when the data is ready i.e when the rgets complete).

I agree that we should revisit the ZC PUP API design on the receiver side. One thing that comes to mind is how currently we take a reference to a pointer as the first argument so it can be populated with the result of the allocate function. However if the PUP routine passes a pointer on the local stack as my implementation does and the transfer does not complete inline, it would corrupt the location on the stack at a later point. We should discuss this further.

> Also, can you test `isomalloc` in a CMA enabled machine and verify that it's working correctly? (Do not pass `+noCMAForZC` in the run since that switches to RDMA).

I have been testing multiprocess runs on a single host in the Charmworks office that supports CMA and only has commodity ethernet.

evan-charmworks:
This branch needs an interactive rebase to remove the my Isomalloc commits.

nitbhat:
It looks like I had pushed some commits from Evan's other branches into this (not sure how that happened). But I have removed those commits from this branch.

evan-charmworks:
This PR has merge conflicts that I am unsure how to resolve. @nitbhat if you could resolve those, I can rebase my Isomalloc use case, and then @stwhite91 can do performance analysis.

nitbhat:
> This PR has merge conflicts that I am unsure how to resolve. @nitbhat if you could resolve those, I can rebase my Isomalloc use case, and then @stwhite91 can do performance analysis.

Sounds good, I will resolve the merge conflicts and push the rebased version.

evan-charmworks:
I've refreshed my `evan/isomalloc_zc_pup_test` branch based on Nitin's rebased zc_pup branch.

stwhite91:
I think we only want to enable zero copy transfers for LB, at least at first. I don't think we have any way currently of knowing from inside a PUP routine if you're being pupped for LB or FT or whatever, so we'll need to add something like that.

nitbhat:
@stwhite91: I built charm's master branch with `syncft` using `./build AMPI mpi-linux-x86_64 syncft --with-production --enable-error-checking --suffix=prod -j16` and was able to run `tests/ampi/chkpt/` successfully.

evan-charmworks:
> @stwhite91: I built charm's master branch with `syncft` using `./build AMPI mpi-linux-x86_64 syncft --with-production --enable-error-checking --suffix=prod -j16` and was able to run `tests/ampi/chkpt/` successfully.

I am able to reproduce the preexisting restore from checkpoint issue on Darwin but not on Linux. My hypothesis is that it is something to do with the internals of `setjmp`/`longjmp`. Either way, it's not related to this PR.
(`tests/ampi/chkpt/` should work without `syncft` since that option is only necessary for _online_ fault tolerance.)

I am able to reproduce the ZC PUP issue by checking out my `evan/isomalloc_zc_pup_test` branch and running: `./build AMPI netlrts-linux-x86_64 -j8 -g3 && cd netlrts-linux-x86_64/tests/ampi/chkpt/ && make OPTS="-g3" -j8 && gdb -ex r --args ./hello +p1`

nitbhat:
> > @stwhite91: I built charm's master branch with `syncft` using `./build AMPI mpi-linux-x86_64 syncft --with-production --enable-error-checking --suffix=prod -j16` and was able to run `tests/ampi/chkpt/` successfully.
> 
> I am able to reproduce the preexisting restore from checkpoint issue on Darwin but not on Linux. My hypothesis is that it is something to do with the internals of `setjmp`/`longjmp`. Either way, it's not related to this PR.
> (`tests/ampi/chkpt/` should work without `syncft` since that option is only necessary for _online_ fault tolerance.)
> 
> I am able to reproduce the ZC PUP issue by checking out my `evan/isomalloc_zc_pup_test` branch and running: `./build AMPI netlrts-linux-x86_64 -j8 -g3 && cd netlrts-linux-x86_64/tests/ampi/chkpt/ && make OPTS="-g3" -j8 && gdb -ex r --args ./hello +p1`

Ah yes, that's why I deleted my previous comment. I hadn't checked out the correct branch previously.

nitbhat:
```
nbhat4@courage:/scratch/nitin/charm_1/src/conv-core$ git diff
diff --git a/src/util/pup_util.C b/src/util/pup_util.C
index 81ac25d2f..333167e38 100644
--- a/src/util/pup_util.C
+++ b/src/util/pup_util.C
@@ -381,12 +381,10 @@ void PUP::toDisk::bytes(void *p,size_t n,size_t itemSize,dataType /*t*/)

 void PUP::toDisk::pup_buffer(void *&p,size_t n,size_t itemSize,dataType t) {
   bytes(p, n, itemSize, t);
-  if(isPacking()) free(p);
 }

 void PUP::toDisk::pup_buffer(void *&p,size_t n, size_t itemSize, dataType t, std::function<void *(size_t)> allocate, std::function<void (void *)> deallocate) {
   bytes(p, n, itemSize, t);
-  if(isPacking()) deallocate(p);
 }
```

If I remove the lines for deallocation (either using free or the custom deallocator passed), `/tests/ampi/chkpt/` passes successfully. 


Does AMPI not want the memory region to be deallocated when using `pup_buffer` to write to disk?

evan-charmworks:
Does it work if you change `isPacking` to `isDeleting`?

nitbhat:
> Does it work if you change `isPacking` to `isDeleting`?

Yes, it does.

evan-charmworks:
Nice, that would be my recommended change.

nitbhat:
> Nice, that would be my recommended change.

Yes, sounds good. I'll add that patch.  
Out of curiosity, I'm guessing, for Isomalloc, we never really free the allocated region when pupping to disk in order to reuse it later?

nitbhat:
Let me know if there are more changes required for this PR based on your review/testing.

evan-charmworks:
We should discuss whether this API needs any adjustment on the receiver side. Does this API guarantee that receives complete inline to the pup_buffer call? If so, we should eliminate the allocate function and change the passed by reference pointer to pass by value.

nitbhat:
> We should discuss whether this API needs any adjustment on the receiver side. Does this API guarantee that receives complete inline to the pup_buffer call? If so, we should eliminate the allocate function and change the passed by reference pointer to pass by value.

Only inter-process CMA transfers happen inline. RDMA transfers are async calls that are called on the receiver side from the pup_buffer call. The RDMA transfers complete behind the scenes and is guaranteed to be complete when the following are called by the RTS -
- ckJustMigrated
- ResumeFromSync
- Any entry method invoked on the object that just migrated calling pup_buffer 


^ Basically these calls are stalled until Rgets complete since the data is not yet ready.

nitbhat:
I was thinking of the use-case where the user would like to not have the RTS allocate the buffer but provide a pre-allocated buffer to receive the data sent using `pup_buffer`. Right now, to meet that use-case the user would have to return the pointer in the custom allocation method. However, it would be simpler for the user if we can provide a variant of the pup_buffer API with pass-by-value `void pup_buffer(void *buf, size_t  size);` where we know the user has taken care of allocation and we just transfer the data in that buffer.

evan-charmworks:
If a transfer does not complete inline, then what happens to the reference to a pointer that is passed in as part of the pup_buffer call? Is it written to later? If the pointer is on the stack of the PUP::er, wouldn't that cause a memory corruption issue?

Additionally, if transfers don't complete inline, what semantics are available for performing some operation using the contents of the received buffer before unpacking is considered complete? Is this case worth supporting? I can't think of an example though.

nitbhat:
> If a transfer does not complete inline, then what happens to the reference to a pointer that is passed in as part of the pup_buffer call? Is it written to later? If the pointer is on the stack of the PUP::er, wouldn't that cause a memory corruption issue?

On the receiver side, in the `pup_buffer` call,  the receiver buffer is allocated (either using the user's custom method or a `malloc`) and the reference to the pointer is initialized to this buffer. 
In the case of CMA, it is written to inline. In the case of RDMA, rgets are launched, which asynchronously complete at a later time. I didn't entirely understand the memory corruption issue. 

> 
> Additionally, if transfers don't complete inline, what semantics are available for performing some operation using the contents of the received buffer before unpacking is considered complete? Is this case worth supporting? I can't think of an example though.

I think the user (application) would be interested in using the buffer only after the data is ready i.e. the rget is complete. In the case of the `pup_buffer` API, since completion is not inline, the user cannot do anything with the buffer inside the `pup` method. Data is considered ready only in the following cases:
1. inside `ckJustMigrated`
2. inside `ResumeFromSync`
3. inside any entry method that is invoked after migration. (So, if the user wants to test if data is ready, they'd have to call the entry method through the scheduler) - I'm thinking if there's an easier way to do this - maybe a callback?

evan-charmworks:
> > If a transfer does not complete inline, then what happens to the reference to a pointer that is passed in as part of the pup_buffer call? Is it written to later? If the pointer is on the stack of the PUP::er, wouldn't that cause a memory corruption issue?
> 
> On the receiver side, in the `pup_buffer` call, the receiver buffer is allocated (either using the user's custom method or a `malloc`) and the reference to the pointer is initialized to this buffer.
> In the case of CMA, it is written to inline. In the case of RDMA, rgets are launched, which asynchronously complete at a later time. I didn't entirely understand the memory corruption issue.

Ah, I see. The allocation always completes inline even if the transfer does not. I was thinking that in the case of non-inline transfers, the allocation would take place immediately before the transfer, and the reference passed to pup_buffer may no longer be valid, causing memory corruption.

> > Additionally, if transfers don't complete inline, what semantics are available for performing some operation using the contents of the received buffer before unpacking is considered complete? Is this case worth supporting? I can't think of an example though.
> 
> I think the user (application) would be interested in using the buffer only after the data is ready i.e. the rget is complete. In the case of the `pup_buffer` API, since completion is not inline, the user cannot do anything with the buffer inside the `pup` method. Data is considered ready only in the following cases:
> 
> 1. inside `ckJustMigrated`
> 2. inside `ResumeFromSync`
> 3. inside any entry method that is invoked after migration. (So, if the user wants to test if data is ready, they'd have to call the entry method through the scheduler) - I'm thinking if there's an easier way to do this - maybe a callback?

From testing, `ckJustMigrated` does exactly what we want, for the checkpoint-restart case too. (Technically this method is currently not available for mainchares and singleton chares, but that is not important for most use cases. I've documented this issue in #2678.)

evan-charmworks:
It looks like this PR hangs with UTH.

nitbhat:
> Hangs on UTH

Fixed in the latest commit 'Support non-migration layers in pup_buffer test'

PR #2463, opened 2019-08-05
cleanup: change #!/bin/sh to #!/bin/bash

matthiasdiener:
Several scripts already require /bin/bash, so we shouldn't lose
compatibility.

This allows future improvements to our scripts, e.g.:
- `[[` instead of `[`
- `source` instead of `.`
- `set -o nounset` etc.


ericjbohm:
Is there a reasonable guarantee that bash is installed?  It is common in many desktop distributions, but there is certainly no certainty that it is expected even in linux distributions.

It is **not** in the LSB.

https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Common/LSB-Common/book1.html

matthiasdiener:
> Is there a reasonable guarantee that bash is installed? It is common in many desktop distributions, but there is certainly no certainty that it is expected even in linux distributions.
> 
> It is **not** in the LSB.
> 
> https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Common/LSB-Common/book1.html

Various scripts already require `/bin/bash` to build and test Charm++, so this doesn't add an additional dependency. E.g.,
- src/util/pup_f.f90.sh
- src/scripts/testrun
- src/scripts/commitid.sh
- ssh script generated by src/util/charmrun-src/charmrun.C
- src/arch/win/gather*.local

matthiasdiener:
Bumping this again. Imho, we can rely on the presence of `/bin/bash` on any system. Even though a user might have a different shell, `/bin/bash` only needs to be present, not the user's shell. Debian, somewhat famously, tried and failed to make bash it removable (https://wiki.debian.org/Proposals/RemoveBashFromEssential). macOS has it. Windows WSL and cygwin have it. As mentioned above, any build of Charm++ requires `/bin/bash` currently, even without this patch. 

Putting a requirement on `/bin/bash` avoids hard-to-find bugs such as #2506,  f4d526293c11950081120b18d6f419c04b502885 or d4bed4db01124bc879235139a86e34d9a7531fd7, apart from the other potential improvements mentioned above.

evan-charmworks:
If we switch to bash and replace all `[` and `test` with `[[`, I would expect build times to improve on Windows, where process launch is expensive.

stwhite91:
I think we can merge this after 6.10 and see if anyone complains about the dependence

matthiasdiener:
> I think we can merge this after 6.10 and see if anyone complains about the dependence

Agreed. The idea is to merge this to the master branch, which won't go into 6.10 anymore.

matthiasdiener:
Needs another review.

PR #2464, opened 2019-08-08
AMPI: debug API arguments

matthiasdiener:
When enabled, prints each MPI function call of an application with the PE/VP that made the call, as well as the names and values of the call arguments.

This is disabled by default, and needs to be enabled by setting `AMPI_DO_DEBUG` to 1. This will also enable `AMPI_DEBUG`.

Printing is done using `std::cout`, and not `CmiPrintf` or `ckout` since:
- `CmiPrintf` etc. can't do automatic formatting of arguments based on their types.
- `ckout` can't print (e.g.) function pointers.

Example output:

```
[3:3] AMPI_Barrier(comm=9000000)
[0:0] AMPI_Scatter(sendbuf=0x7fb5b91066c0, sendcount=1, sendtype=1, recvbuf=0x580103a64, recvcount=1, recvtype=1, root=3, comm=9000000)
[0:0] AMPI_Reduce(inbuf=0x58010393c, outbuf=0x580103930, count=1, type=1, op=2, root=3, comm=9000000)
[2:2] AMPI_Isend(buf=0x3fdb80103e78, count=1, type=1, dest=0, tag=12387, comm=1000000, request=0x3fdb80103afc)
[2:2] AMPI_Wait(request=0x3fdb80103afc, sts=0x0)
```

### DIFF
--- a/src/libs/ck-libs/ampi/ampiimpl.h
+++ b/src/libs/ck-libs/ampi/ampiimpl.h
@@ -20,7 +20,7 @@
 
 #define AMPI_DEBUG(...) CkPrintf(__VA_ARGS__)
 
-// Support for variable-argument macros
+// Support for variable-argument macros (up to 16 arguments)
 #define FE_1(WHAT, X) WHAT(X, true) 
 #define FE_2(WHAT, X, ...) WHAT(X, false)FE_1(WHAT, __VA_ARGS__)
 #define FE_3(WHAT, X, ...) WHAT(X, false)FE_2(WHAT, __VA_ARGS__)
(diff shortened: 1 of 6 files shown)


src/libs/ck-libs/ampi/ampiimpl.h
```diff
@@ -7,13 +7,59 @@
...
+// Perform 'action' (PRINT_ARG in this case) on each argument
+#define FOR_EACH(action,...) \
+  GET_MACRO(__VA_ARGS__,FE_16,FE_15,FE_14,FE_13,FE_12,FE_11,FE_10,FE_9,FE_8,FE_7,FE_6,FE_5,FE_4,FE_3,FE_2,FE_1)(action,__VA_ARGS__)
+
+// Print a single argument name and its value (unless the argument name is '""', which indicates a nonexistent argument)
+#define PRINT_ARG(arg, last) if ("\"\""!=#arg) std::cout << #arg << "=" << arg << (last ? "" : ", ");
+
+// Prints the rank, function name, and argument name/value for each function argument
+#define AMPI_DEBUG_ARGS(function_name, ...) \
+  std::cout << "[" << (isAmpiThread() ? getAmpiParent()->thisIndex : -1) << "] "<< function_name <<"("; \
```
stwhite91: It would be good to print the PE and the VP

matthiasdiener: VP is just `thisIndex`, right?

evan-charmworks:
Is there any way this could be optionally disabled at runtime once built in, maybe via `++quiet`? HDF5's build process relies on the output of two binaries, which breaks due to this diagnostic output when enabled.

matthiasdiener:
> Is there any way this could be optionally disabled at runtime once built in, maybe via `++quiet`? HDF5's build process relies on the output of two binaries, which breaks due to this diagnostic output when enabled.

But wouldn't it break in any case if `AMPI_DEBUG` is enabled?

src/libs/ck-libs/ampi/ampiimpl.h
matthiasdiener: added printing of PE.

evan-charmworks:
Yes, I think so.

matthiasdiener:
> Yes, I think so.

Interestingly, `CmiPrintf` in src/conv-core/convcore.C should not print anything when quietMode is active, but the messages still get printed.

matthiasdiener:
> Is there any way this could be optionally disabled at runtime once built in, maybe via `++quiet`? HDF5's build process relies on the output of two binaries, which breaks due to this diagnostic output when enabled.

The most recent update should honor ++quiet.

PR #2460, opened 2019-08-02
AMPI+CUDA: fix build

matthiasdiener:
Fixes #2427. Followup from #2062/#2121. See also #2459. Only lightly tested.

PR #2121, opened 2019-05-08
Bug #2062: Don't force -lcudart to non-CUDA programs

minitu:
*Original date: 2019-05-07 22:06:37*
*Original PR: https://charm.cs.illinois.edu/gerrit/5176*

---

Change-Id: I857c852bd88ef890de1c29ed151a6f269c546f19

PR #2466, opened 2019-08-08
Revert+fix cuda linking

matthiasdiener:
Reverts #2462 #2460 #2121 .

Manually tested on talent.


matthiasdiener:
This should fix the cuds-related auto build issues. Ready for review.

PR #2468, opened 2019-08-12
Remove BigSim testing of ampi/datatype test which uses MPI_Ssend

stwhite91:
BigSim currently doesn't work for AMPI programs that use MPI_Ssend,
because AMPI uses Charm++'s zero copy API to implement it and that
currently lacks BigSim support.


matthiasdiener:
Can we define MPI_Ssend to MPI_Send when compiling for bigsim and avoid this issue?

stwhite91:
That would violate the MPI standard, since Ssend has extra semantics (guaranteeing that the send has been matched to a receive) that plain Send doesn’t have to support (it can return as soon as the buffer may be reused).

matthiasdiener:
Isn't the handling in AMPI (in terms of matching to a recv) the same? (in `waitOnBlockingSend`)

stwhite91:
The difference is that when AMPI does an eager Send, it copies the the user's buffer into an AmpiMsg and sends it, creates a SendReq that is 'complete', and returns from the MPI_Send call immediately. For a rendezvous Send (which is the same as an Ssend), it creates an 'incomplete' SsendReq and an AmpiMsg that is marked as SsendMsg, sends it, and does not return from the MPI_Send (or MPI_Ssend) until the recv'er has matched it, performed an rget on the buffer, and the runtime has invoked a completedSend callback which finally marks the SsendReq 'complete', and resumes the sender's thread.

matthiasdiener:
Ok, thanks for the explanation. We might want to document somewhere which MPI calls aren't supported in BigSim (if that isn't deprecated yet). Are there others apart from `MPI_Ssend`/ `MPI_Issend`?

stwhite91:
> Ok, thanks for the explanation. We might want to document somewhere which MPI calls aren't supported in BigSim (if that isn't deprecated yet). Are there others apart from `MPI_Ssend`/ `MPI_Issend`?

Not that I know of, but I wouldn't be surprised if other parts were broken for it

PR #2469, opened 2019-08-13
CkDDT: Replace vector type with set freeTypes

evan-charmworks:
Solves a performance bug where CkDDT::insertType was linear in
the number of total types.

### DIFF
--- a/src/libs/ck-libs/ampi/ddt.C
+++ b/src/libs/ck-libs/ampi/ddt.C
@@ -5,12 +5,30 @@
 void
 CkDDT::pup(PUP::er &p) noexcept
 {
-  p|types;
+  size_t numTypes = userTypeTable.size();
+  p|numTypes;
+
+  std::vector<signed char> types;
+  if (p.isPacking()) {
+    types.resize(numTypes);
+    for (int i=0; i<userTypeTable.size(); i++) {
+      types[i] = userTypeTable[i] == nullptr ? MPI_DATATYPE_NULL : std::max(AMPI_MAX_PREDEFINED_TYPE, std::min(CkDDT_MAX_PREDEFINED_TYPE, userTypeTable[i]->getType()));
+    }
+    p(types.data(), numTypes);
+    types.clear();
+  }
+  else if (p.isSizing()) {
+    p((signed char *)nullptr, numTypes);
+  }
+
   if (p.isUnpacking()) {
-    userTypeTable.resize(types.size(), nullptr);
+    types.resize(numTypes);
+    p(types.data(), numTypes);
+    userTypeTable.resize(numTypes, nullptr);
     for (int i=0; i<types.size(); i++) {
       switch (types[i]) {
         case MPI_DATATYPE_NULL:
+          freeTypes.emplace(i);
           break;
         case CkDDT_CONTIGUOUS:
           userTypeTable[i] = new CkDDT_Contiguous;
@@ -41,11 +59,19 @@ CkDDT::pup(PUP::er &p) noexcept
           break;
       }
     }
+
+    for (int i=0; i<types.size(); i++) {
+      if (types[i] != MPI_DATATYPE_NULL) {
+        userTypeTable[i]->pupType(p, this);
+      }
+    }
+
+    return;
   }
 
-  for (int i=0; i<types.size(); i++) {
-    if (types[i] != MPI_DATATYPE_NULL) {
-      userTypeTable[i]->pupType(p, this);
+  for (CkDDT_DataType * userType : userTypeTable) {
+    if (userType != nullptr) {
+      userType->pupType(p, this);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/libs/ck-libs/ampi/ddt.h
```diff
@@ -43,6 +44,7 @@
 #define CkDDT_INDEXED             47
 #define CkDDT_HINDEXED            48
 #define CkDDT_STRUCT              49
+#define CkDDT_MAX_PREDEFINED_TYPE 50
```
stwhite91: This name seems confusing to me with AMPI_PMAX_PREDEFINED_TYPE defined above. I think something like CkDDT_MAX_USERDEFINED_TYPE would be better. Could also rename all these CkDDT_* macros to AMPI_*

src/libs/ck-libs/ampi/ddt.h
```diff
@@ -446,15 +448,13 @@ class CkDDT_Struct final : public CkDDT_DataType
  *                       (to minimize per-rank memory fooprint), which holds the CkDDT_DataType
  *                       object pointers for all predefined types.
  * userTypeTable - a vector that holds the CkDDT_DataType object pointers for all user-defined types
- * types - used to identify which CkDDT_DataType derived class a type object really is,
```
matthiasdiener: could add a comment regarding freeTypes here.

evan-charmworks: Done

src/libs/ck-libs/ampi/ddt.h
evan-charmworks: Renamed

src/libs/ck-libs/ampi/ddt.C
```diff
@@ -12,7 +12,9 @@ CkDDT::pup(PUP::er &p) noexcept
   if (p.isPacking()) {
     types.resize(numTypes);
     for (int i=0; i<userTypeTable.size(); i++) {
-      types[i] = userTypeTable[i] == nullptr ? MPI_DATATYPE_NULL : std::max(AMPI_MAX_PREDEFINED_TYPE, std::min(CkDDT_MAX_PREDEFINED_TYPE, userTypeTable[i]->getType()));
+      types[i] = userTypeTable[i] == nullptr
+        ? MPI_DATATYPE_NULL
+        : std::max(AMPI_MAX_PREDEFINED_TYPE, std::min(CkDDT_FIRST_USER_TYPE, userTypeTable[i]->getType()));
```
matthiasdiener: maybe do an if...then...else here

src/libs/ck-libs/ampi/ddt.h
```diff
@@ -443,18 +445,17 @@ class CkDDT_Struct final : public CkDDT_DataType
...
- *         for PUPing the userTypeTable
+ * freeTypes - an ordered set of freed slot indexes in the userTypeTable, available for reuse.
  */
 class CkDDT
 {
  private:
   const std::array<const CkDDT_DataType *, AMPI_MAX_PREDEFINED_TYPE+1>& predefinedTypeTable;
   std::vector<CkDDT_DataType *> userTypeTable;
-  std::vector<int> types;
+  std::set<int> freeTypes; // TODO: replace with a priority queue structure that can be serialized trivially
```
matthiasdiener: Out of curiosity, why is a priority queue the intended data structure here? Could we use an `unordered_map` +  int (as an index)?

src/libs/ck-libs/ampi/ddt.C
stwhite91: Could you explain the logic here too? Why is it not just either MPI_DATATYPE_NULL or userTypeTable[i]->getType()?

matthiasdiener: (maybe add the explanation as a comment)

src/libs/ck-libs/ampi/ddt.h
matthiasdiener: ie, something like 
```c++
std::unordered_map<int, CkDDT_DataType *> userTypeTable;
int lastIndex;

// in insertType()
userTypeTable[++lastIndex] = ptr;
return lastIndex;
```

src/libs/ck-libs/ampi/ddt.C
evan-charmworks: Reformatted and added a comment

src/libs/ck-libs/ampi/ddt.h
evan-charmworks: How would changing the type of userTypeTable from a vector to an unordered_map solve the problem of knowing which indexes are empty and available for reused?

freeTypes is a priority queue so that AMPI can fill in empty slots from least to greatest, so that types freed at the end of the list have the greatest change of being able to shrink userTypeTable as a whole.

matthiasdiener: > How would changing the type of userTypeTable from a vector to an unordered_map solve the problem of knowing which indexes are empty and available for reuse?

Do we need to know which indices can be reused (ie, can't we just increase the indices)?

matthiasdiener:
> The unordered_map + index counter approach would also work, but I don't really have a preference between that and this

Yeah, I think that might be better as future work. I don't think that would affect performance, but the logic in insertType/freeType/pup might become simpler.

PR #2470, opened 2019-08-14
Fix CmiIsomallocBlockList when running with +noisomalloc

evan-charmworks:
For example, this fixes tlsglobals with +noisomalloc.


evan-charmworks:
Good point, it looks like that would still cause a segfault.

evan-charmworks:
This is further improved and will be included with my Isomalloc refactoring

PR #2471, opened 2019-08-14
Re-privatize ROMIO portably

evan-charmworks:
CtvAccess only happens to work for AMPI global variable privatization when tlsglobals is active and Ctv uses __thread underneath. The new way works regardless of any global variable privatization method in use.

(Please rebase and not squash.)


src/libs/ck-libs/ampi/mpio_globals.h
```diff
@@ -0,0 +1,57 @@
...
+    char * token_ptr;
+};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void ADIO_Init_Globals(struct ADIO_GlobalStruct *);
+struct ADIO_GlobalStruct * ADIO_Globals(void);
+#define ADIO_Global(x) (ADIO_Globals()->x)
```
matthiasdiener: Is it possible to do something like `#define ADIOI_Flatlist (ADIO_Globals()-> ADIOI_Flatlist)`? This would remove the need to change the romio source code.

matthiasdiener:
Can you add an explanation (in the commit message or so) on why this is necessary? What are the implications for privatizing other libraries (such as hdf5)?

stwhite91:
Other than Matthias's comments, this looks good

src/libs/ck-libs/ampi/mpio_globals.h
evan-charmworks: Done

evan-charmworks:
> Can you add an explanation (in the commit message or so) on why this is necessary?

Done

> What are the implications for privatizing other libraries (such as hdf5)?

There should be none, this simply improves ROMIO's portability with AMPI.

evan-charmworks:
Any comment on how the file `mpio_globals.h` is handled?

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -1364,6 +1368,50 @@ ampiParent::ampiParent(CkMigrateMessage *msg) noexcept
...
+  /* list of datareps registered by the user */
+
+  /* for f2c and c2f conversion */
+  globals->ADIOI_Ftable = NULL;
+  globals->ADIOI_Ftable_ptr = 0;
+  globals->ADIOI_Ftable_max = 0;
+  globals->ADIOI_Reqtable = NULL;
+  globals->ADIOI_Reqtable_ptr = 0;
+  globals->ADIOI_Reqtable_max = 0;
+#ifndef HAVE_MPI_INFO
```
matthiasdiener: Are these preprocessor symbols available when compiling AMPI itself?

evan-charmworks: Do you mean `HAVE_MPI_INFO`? See the change to ampiimpl.h

src/libs/ck-libs/ampi/mpio_globals.h
```diff
@@ -0,0 +1,56 @@
...
+    int ADIOI_Ftable_ptr;
+    int ADIOI_Ftable_max;
+    MPI_Request * ADIOI_Reqtable;
+    int ADIOI_Reqtable_ptr;
+    int ADIOI_Reqtable_max;
+#ifndef HAVE_MPI_INFO
+    MPI_Info *MPIR_Infotable;
+    int MPIR_Infotable_ptr, MPIR_Infotable_max;
+#endif
+#if defined(ROMIO_XFS) || defined(ROMIO_LUSTRE) || 1
```
stwhite91: Why the "|| 1"?

src/libs/ck-libs/ampi/ampi.C
matthiasdiener: Maybe just add a comment (also ROMIO_XFS, ROMIO_LUSTRE ).

matthiasdiener:
LGTM, but this is for post-6.10, right?

src/libs/ck-libs/ampi/mpio_globals.h
evan-charmworks: See the `// TODO: Sync CMK_HAS_LUSTREFS with ROMIO_LUSTRE` in ampiimpl.h. I don't have access to a Lustre machine to test this. Better to always include the two integers in the struct than fail sometimes.

evan-charmworks:
> LGTM, but this is for post-6.10, right?

Yes

matthiasdiener:
By the way,

> CtvAccess only happens to work when tlsglobals is active and Ctv uses
__thread underneath

is this documented somewhere?

matthiasdiener:
> > What are the implications for privatizing other libraries (such as hdf5)?
> 
> There should be none, this simply improves ROMIO's portability with AMPI.

My question was more related to whether we should do the same style of privatization for (e.g.) HDF5, but if I understood correctly, since we are using `__thread` there, this shouldn't be an issue.

evan-charmworks:
Clarified the commit message

evan-charmworks:
Ready for re-review

evan-charmworks:
Yes. I re-added it to the top post of the PR, and the description can be found in commit 26f454db0996e21eaa2c937d3fea3e3160ea0346.

PR #2475, opened 2019-08-16
Do not run bgtest for ampi datatypes test

stwhite91:


matthiasdiener:
shouldn't we put exit into NOBGDIRS in tests/ampi/Makefile ?

matthiasdiener:
I'll have a patch that implements exit() correctly for BigSim soon, so this PR might not be necessary.

matthiasdiener:
> I'll have a patch that implements exit() correctly for BigSim soon, so this PR might not be necessary.

Nvm, just saw that this is about datatypes, not exit

PR #2479, opened 2019-08-19
Add Zenodo DOI to README.md

rbuch:


matthiasdiener:
This should also be added to Citation.cff, right?

rbuch:
> This should also be added to Citation.cff, right?

Yes, the citation file should also be updated, I'll add that to this PR as well.

rbuch:
The DOI I added to README.md is the umbrella DOI (10.5281/zenodo.3370873, it always resolves to the latest version). The DOI I added to CITATION.cff is for the specific release (10.5281/zenodo.3370874). I did this because the Citation File Format requires a version and release date, so it made more sense to use the specific DOI for it in the citation file.

Anyone have any input on what DOIs to put where? Should README.md have the umbrella DOI or should we update it for every release?

matthiasdiener:
> The DOI I added to README.md is the umbrella DOI (10.5281/zenodo.3370873, it always resolves to the latest version). The DOI I added to CITATION.cff is for the specific release (10.5281/zenodo.3370874). I did this because the Citation File Format requires a version and release date, so it made more sense to use the specific DOI for it in the citation file.
> 
> Anyone have any input on what DOIs to put where? Should README.md have the umbrella DOI or should we update it for every release?

I think what you proposed makes sense, although having to update the citation file for each release seems unfortunate. 

Looking at Github code search for other citation files (https://github.com/search?q=extension%3Acff+zenodo&type=Code), it seems that at least some other projects use the Zenodo DOI without version/release date (e.g. https://github.com/amusecode/amuse/blob/master/CITATION.cff) , so we might want to consider doing that as well.

rbuch:
> > The DOI I added to README.md is the umbrella DOI (10.5281/zenodo.3370873, it always resolves to the latest version). The DOI I added to CITATION.cff is for the specific release (10.5281/zenodo.3370874). I did this because the Citation File Format requires a version and release date, so it made more sense to use the specific DOI for it in the citation file.
> > Anyone have any input on what DOIs to put where? Should README.md have the umbrella DOI or should we update it for every release?
> 
> I think what you proposed makes sense, although having to update the citation file for each release seems unfortunate.
> 
> Looking at Github code search for other citation files (https://github.com/search?q=extension%3Acff+zenodo&type=Code), it seems that at least some other projects use the Zenodo DOI without version/release date (e.g. https://github.com/amusecode/amuse/blob/master/CITATION.cff) , so we might want to consider doing that as well.

Yeah, I think I prefer using the umbrella DOI everywhere, but I'm hesitant to buck the standard.

CITATION.cff
```diff
@@ -1,22 +1,26 @@
 # YAML 1.2
+# Note that we have omitted the release date and version due to using the umbrella DOI
 ---
 cff-version: "1.0.3"
 title: "Charm++"
```
matthiasdiener: Maybe put "The Charm++ parallel programming system"

PR #2481, opened 2019-08-19
Cleanup: ZC Bcast Post API code in CkArray methods

nitbhat:


src/ck-core/ckarray.C
```diff
@@ -1471,21 +1471,20 @@ void CkArray::recvBroadcast(CkMessage *m)
...
-      unsigned int i=0;
-      bool doFree = true;
-      if (stableLocations && i == len-1) doFree = true;
-      broadcaster->deliver(msg, (ArrayElement*)localElemVec[i], doFree);
+    if (zc_msgtype == CMK_ZC_BCAST_RECV_ALL_DONE_MSG) { // message contains pointers to the posted buffer, which contains the data received
+      // All operations done, already consumed by other array elements, now deliver to the first element
 
-    } else if (zc_msgtype == CMK_ZC_BCAST_RECV_MSG) { // deliver to all array elements
+      bool doFree = true; // free it since all ops are done
+      broadcaster->deliver(msg, (ArrayElement*)localElemVec[firstElemIndex], doFree);
```
juanjgalvez: Why not just localElemVec[0]?

src/ck-core/ckarray.C
```diff
@@ -1471,21 +1471,20 @@ void CkArray::recvBroadcast(CkMessage *m)
...
 
+    } else if (zc_msgtype == CMK_ZC_BCAST_RECV_MSG) { // message is used by the receiver to post the receiver buffer
+      // Initial metadata message, send only to the first element, other elements are sent CMK_ZC_BCAST_RECV_DONE_MSG after rget completion
 
-      unsigned int i=0;
-      bool doFree = false;
-      if (stableLocations && i == len-1) doFree = true;
-      broadcaster->deliver(msg, (ArrayElement*)localElemVec[i], doFree);
+      bool doFree = false; // do not free since msg will be reused to send buffers to peers,
```
juanjgalvez: Previously there was a chance here that doFree was set to true (if stableLocations and len==1 right?). I'm not sure if we either need to keep that, or if it was a bug.

juanjgalvez:
BTW, what happens to ZC code path if there are no array elements on a PE? Can the ZC messages be received by CkArray in that case, and will it attempt to do localElemVec[0] even though it is empty or does it take that into account somewhere?

src/ck-core/ckarray.C
nitbhat: Lol yeah, good point.

src/ck-core/ckarray.C
nitbhat: Yeah, it shouldn't be freed since the same message is re-used and freed later. I'll run a test with len=1 and ensure that it crashes with the previous code.

nitbhat: The previous code never crashed because `stableLocations` was false and that caused `doFree` to never be true. But that was a bug nevertheless since the message shouldn't be freed.

nitbhat:
> BTW, what happens to ZC code path if there are no array elements on a PE? Can the ZC messages be received by CkArray in that case, and will it attempt to do localElemVec[0] even though it is empty or does it take that into account somewhere?

This was a case I didn't handle in the previous code. I always had more than 1 array element/PE in my tests, so I overlooked this corner case. I'll add `len > 0` to the if condition to ensure that `localElemVec[0]` is only used when there is a 0th element.

nitbhat:
Thanks @juanjgalvez, I had not considered the case of len being 0 on each PE. There is an issue with ZC Bcast Post API whenever the first PE of the logical node has 0 elements. I should either mark that as a known issue or post a fix before 6.10. 


From slack chat on #core:

```
nbhat4  How likely is it for the number of chare array elements of a chare array to be less than the number of total PEs? (since ideally, the number of chare array elements ~ 10 * CkNumPes())
I am asking this because I realized that ZC Bcast Post API for chare arrays doesn’t handle some corner cases. (Currently, it assumes that there’s an array element on the first PE of a logical node).
For 6.10, I am inclined to mark this as a known issue, since supporting it will change the implementation quite a bit.

swhite  OpenAtom uses numChareArrayElements < Cores for specific chare arrays, though they have multiple different chare arrays so in that way make use of all cores
subcommunicators in AMPI are similar

nbhat4  Ah okay, makes sense. Yeah, that assumption that there’ll be a chare array element on every logical node was fundamentally incorrect.
So, for ZC Bcast Post API, the mapping of chare array elements to PEs makes all the difference. (for eg, there might be chare array elements on the root node and leaves, but none on intermediate nodes. In such a case, the leaves cannot rget from their immediate parents, they should rget from the last parent that has an element).

swhite  This only effects the post api version of ZC Bcast?


nbhat4  Yes, that’s right. (Since the ZC Bcast Send API allocates a buffer on every process (logical node) irrespective of the number of elements present.) (edited) 
```

Issue #2472, opened 2019-08-15
netlrts-linux-x86_64-bigsim fails in megampi due to rput


evan-charmworks:
```
Charm++> Running in non-SMP mode: 2 processes (PEs)
Converse/Charm++ Commit ID: v6.9.0-529-gff4c238f6
Charm++> scheduler running in netpoll mode.
BG info> Simulating 2x1x1 nodes with 1 comm + 1 work threads each.
BG info> Network type: bluegene.
alpha: 1.000000e-07     packetsize: 1024        CYCLE_TIME_FACTOR:1.000000e-03.
CYCLES_PER_HOP: 5       CYCLES_PER_CORNER: 75.
BG info> cpufactor is 1.000000.
BG info> floating point factor is 0.000000.
BG info> Using WallTimer for timing method. 
CharmLB> Load balancer ignores processor background load.
CharmLB> Load balancer assumes all CPUs are same.
Trace: traceroot: /home/evan/charm/netlrts-linux-x86_64-bigsim/tests/ampi/megampi/bg_pgm
CharmLB> RandCentLB created.
[0] Testing: Testing...
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: Calling BgMyRank in the main thread!
[0] Stack Traceback:
  [0:0] _Z14CmiAbortHelperPKcS0_S0_ii+0xb9  [0x55555594d8d7]
  [0:1] CmiGetNonLocal+0  [0x55555594d9b1]
  [0:2] BgMyRank+0x47  [0x55555580543b]
  [0:3] _ZN18CMessage_CkDataMsg5allocEimPii11GroupDepNum+0x44  [0x55555585a626]
  [0:4] _ZN18CMessage_CkDataMsgnwEmPii+0x42  [0x55555585a4dd]
  [0:5] _ZN9CkDataMsg8buildNewEiPKv+0x28  [0x555555859e24]
  [0:6] _ZNK10CkCallback4sendEiPKv+0x34  [0x555555858caa]
  [0:7] _Z14invokeCallbackPviR12CkNcpyBuffer+0x28  [0x5555558de9d3]
  [0:8] _Z20invokeSourceCallbackP10ncpystruct+0xa8  [0x5555558de778]
  [0:9] _Z25handleDirectApiCompletionP10ncpystruct+0x63  [0x5555558de8f5]
  [0:10] _Z22CkRdmaDirectAckHandlerPv+0x54  [0x5555558de96a]
[...]

PR #2482, opened 2019-08-19
AMPI: work around issue #2472 in ampi::requestPut code path

stwhite91:
AMPI bug #2472 occurs after migration in SMP mode runs with
multiple processes. It arises when a right after migration a sender
thinks it is in the same process as the recv'er but is not. By disabling
AMPI_NODE_LOCAL_IMPL we fall back to always using the zero copy
get()-based scheme.


stwhite91:
Real fix incoming

PR #2488, opened 2019-08-22
Remove unneeded locking from CkLocMgr

epmikida:
These locks were added in 2004, and only protect some accesses to
the hash member of CkLocMgr. Lock usage is inconsistent, none of
the places its used are accessed by multiple threads, and TSAN
reports no issues after removing the locks.


evan-charmworks:
Do we want to merge this for the release or hold it back?

epmikida:
My assumption is hold it, since it doesn't address and 6.10 issues and we just said we were going to freeze unnecessary merges in yesterdays core.

Issue #2489, opened 2019-08-22
Remove converse tests from cuda auto build target


matthiasdiener:
These tests currently fail due to #2467.

evan-charmworks:
Fixed by #2490

PR #2490, opened 2019-08-22
Workaround #2489: Don't run Converse tests when CUDA is enabled

evan-charmworks:
Needs to be tested

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -6,7 +6,11 @@ OPTS += $(OPTSATBUILDTIME)
 DIRS = \
   ampi \
   charm++ \
-  converse \
+
+# TODO #2467: eliminate the need for this
+ifneq ($(BUILD_CUDA),1)
+  DIRS += converse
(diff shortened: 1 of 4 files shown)


evan-charmworks:
Testing now

evan-charmworks:
Tested and confirmed on talent, ready for review

benchmarks/Makefile
```diff
@@ -1,11 +1,16 @@
+-include ../include/conv-mach-opt.mak
+
 # Temporary hack that needs to be removed / replaced
```
matthiasdiener: What is the 'temporary hack' referring to now? (also in the other files)

evan-charmworks:
`OPTS += $(OPTSATBUILDTIME)`

benchmarks/Makefile
evan-charmworks: `OPTS += $(OPTSATBUILDTIME)`

PR #2492, opened 2019-08-22
Fix barrier implementation and usage at exit

rbuch:


rbuch:
Fixes #2494

evan-charmworks:
This is more of a workaround than a fix. The fully-fledged solution is for all threads to terminate, potentially by returning from `call_startfn`, before the remaining thread calls `exit`. This patch adds a single-use spinlock barrier after the final use of the main barrier infrastructure to avoid the race condition.

PR #2498, opened 2019-08-27
Fix bug in record/replay with migration

epmikida:
An incorrect if, which only evaluated to true when using
+replay-detail, prevented LBDB pointers from being correctly set
when reading in migrate messages from a replay file. This caused a
crash later on when trying to access LBDB via invalid pointers read
in from the replay file, instead of the correct pointers for the
current execution.


ericjbohm:
Normal record replay only records enough about every message to provide an unambiguous ordering.

Replay-detail actually records the entire message.  It is quite a disk space killer.  Only really useful to do if you’re going to do some kind of transformation, as might come from simulating different hardware, then replay that.

PR #2500, opened 2019-08-29
tlsglobals interface and calculation cleanup

evan-charmworks:


evan-charmworks:
This is further improved and will be included with my Isomalloc refactoring

PR #2501, opened 2019-08-29
Armv8 does not build with UCX

paklui:
Adding the file to src/arch/ucx-linux-arm8 that allow "ucx-linux-arm8" to build.

### DIFF
--- a/src/arch/ucx-linux-arm8/conv-mach.h
+++ b/src/arch/ucx-linux-arm8/conv-mach.h
@@ -1,0 +1,94 @@
+#ifndef _CONV_MACH_H
+#define _CONV_MACH_H
+
+#define CMK_UCX 1
+
+/* define the default linker, together with its options */
+#define CMK_DLL_CC   "g++ -shared -O3 -o "
+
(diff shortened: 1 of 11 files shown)


src/arch/ucx-linux-arm8/conv-mach.h
```diff
@@ -0,0 +1,94 @@
...
+   or not (0) */
+#define CMK_WEB_MODE                                       1
+
+#define CMK_DEBUG_MODE                                     0
+
+/* enables the load balancer framework. set to 1 for almost all the machines */
+#define CMK_LBDB_ON                    1
+
+#define CMK_64BIT                      1
+#define CMK_AMD64                      1
```
stwhite91: Should this be 0 for arm8?

paklui: I see these ones are for enabling the load balancer framework. The Armv8 supports 64 bit, so CMK_64BIT should be 1. The CMK_AMD64 should be 0 (but I see the other non-x86 arch like the ppc has it set to 1) but should be changed.

PR #2502, opened 2019-08-30
Revert charm4py cmiwrappers

juanjgalvez:
I was able to get the different interface layers in Charm4py to work directly with CmiPrintf and CmiAbort


juanjgalvez:
> This could go into 6.10, imho.

Agreed, and it won't break anything. Charm4py master is already using the original functions.

PR #2504, opened 2019-08-30
cleanup: Address sanitizer reports leaks for charmrun

ericjbohm:
This cleans up a few minor leaks in argument parsing so that
address sanitizer will at least run cleanly when ++local is used.

### DIFF
--- a/src/util/charmrun-src/charmrun.C
+++ b/src/util/charmrun-src/charmrun.C
@@ -241,6 +241,20 @@ static char *substr(const char *lo, const char *hi)
   return res;
 }
 
+
+/* get substring from lo to hi, remove quote chars */
+static const char *substr_const(const char *lo, const char *hi)
+{
+  if (is_quote(*lo))
+    lo++;
+  if (is_quote(*(hi - 1)))
+    hi--;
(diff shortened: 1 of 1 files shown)


src/util/charmrun-src/charmrun.C
```diff
@@ -241,6 +241,20 @@ static char *substr(const char *lo, const char *hi)
...
+/* get substring from lo to hi, remove quote chars */
+static const char *substr_const(const char *lo, const char *hi)
+{
+  if (is_quote(*lo))
+    lo++;
+  if (is_quote(*(hi - 1)))
+    hi--;
+  int len = hi - lo;
+  assert(len>0);
+  const char *res = strndup(lo,len);
```
matthiasdiener: Where is this freed?

ericjbohm: These end up in an array of the parsed arguments which is handled cleanly in the ++local case.  The other cases were harder to follow and asan will complain about them in a way that didn't lead me to find a clean solution.  Overall, I would recommend that any further time devoted to this be spent refactoring the entire thing to use String.

PR #2506, opened 2019-09-02
remove '<<<' bashism from configure

### LINKED ISSUE #2503: No rule to make target 'conv-autoconfig.h', needed by 'xi-main.o'
I'm attempting to build on alpine linux, using
```
build charm++ mpi-linux-x86_64 --build-shared --with-production --disable-charmdebug -j36 -O3 -DNDEBUG
```
I get
```
./configure: line 3806: syntax error: unexpected redirection
make[1]: *** No rule to make target 'conv-autoconfig.h', needed by 'xi-main.o'.  Stop.
make[1]: Leaving directory '/home/quinoa/quinoa-tpl/install/gnu-x86_64/charm/mpi-linux-x86_64/tmp'
```
Please let me know if you need more information.

matthiasdiener:
Fixes #2503.

### DIFF
--- a/src/scripts/configure.ac
+++ b/src/scripts/configure.ac
@@ -822,7 +822,7 @@ EOT
         got_opt_whole_archive="true"
         IFS=' ' read LDXX_WHOLE_ARCHIVE_PRE LDXX_WHOLE_ARCHIVE_POST <<EOF
         $i
-        EOF
+EOF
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Imho, this could still go into 6.10 (it only changes the configure test for Charm4py).

@jbakosi Could you please test whether this fixes #2503 for you? (the fix here is slightly different from what I had suggested in #2503)
@juanjgalvez Could you please check whether Charm4py works correctly with this change?

matthiasdiener:
Juan confirmed Charm4py still works with this patch.

PR #2509, opened 2019-09-03
citation.cff: add BibTeX entries as comments

matthiasdiener:
Author listing in the Zenodo entry follows Zenodo's own list (minus putting Sanjay first).

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -10,6 +10,22 @@ authors:
 message: "If you use this software, please cite both the software and the given book chapter."
 repository-code: "https://github.com/UIUC-PPL/charm"
 doi: "10.5281/zenodo.3370873"
+notes: |
+  @misc{charm-zenodo,
+    author = {Kale, Laxmikant and White, Sam and Ramos, Evan and Jain, Nikhil
+    and Bhatele, Abhinav and Bohm, Eric and Venkataraman, Ramprasad and
+    Chakravorty, Sayantan and Becker, Aaron and Bhat, Nitin and Diener,
+    Matthias and Lifflander, Jonathan and Menon, Harshitha and Buch, Ronak and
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Nota bene that Zenodo's BibTeX generator creates illegal bibtex entries...

matthiasdiener:
> Where is that order visible to others besides this file? In our zenodo webpage?
> 
> Also, is the order arbitrary? Not sure if there was a discussion about this in core, but maybe we want a non-arbitrary order, like chronological or something else.

Yes, it is the same order as here: https://zenodo.org/record/3370874#.XW65yS2ZNTY

We discussed this in core, but we didn't really figure out how the order was generated (apart from some general notion of "number of contributions").

rbuch:
Note that we can also modify the listing (in terms of ordering and in terms of who is included) that Zenodo uses to whatever we want it to be.

It may make more sense to put the BibTeX entries under the `notes` key of a CFF file so that it has some semantic meaning rather than putting them in as comments.

matthiasdiener:
> It may make more sense to put the BibTeX entries under the `notes` key of a CFF file so that it has some semantic meaning rather than putting them in as comments.

Converted to notes.

juanjgalvez:
This is fine by me, but maybe we should check that there is consensus in core?

rbuch:
> Nota bene that Zenodo's BibTeX generator creates illegal bibtex entries...

Just to check, I assume the entries you've added to the CITATION.cff file are not the same as these illegal bibtex entries, right?

matthiasdiener:
> > Nota bene that Zenodo's BibTeX generator creates illegal bibtex entries...
> 
> Just to check, I assume the entries you've added to the CITATION.cff file are not the same as these illegal bibtex entries, right?

Yes, that's correct.

matthiasdiener:
Author order changed to Zenodo's (alphabetically with Sanjay first).

PR #2510, opened 2019-09-03
Preview of new LB Framework

rbuch:
This PR isn't intended to ever be merged, I'm just making it so that others can look at and "review" the changes of the new LB Framework using the PR tooling Github provides.

When it comes time to actually merge this work, new PRs will be made in a more well separated way for real review.


src/ck-ldb/tree_level.ci
```diff
@@ -0,0 +1,35 @@
...
+    int num_incoming[];
+    int obj_start[];
+    int to_pes[];
+  };
+
+  message SubtreeLoadMsg {
+  };
+
+  message SubtreeMigrateDecisionMsg {
+    // FIXME? I don't think this layout is ideal, but I don't know if ci files
```
stwhite91: We should support arrays of custom structs in messages, particularly if those types are declared as PUPbytes

stwhite91: By that I mean I think we already do support it

juanjgalvez: Thanks, I will have to revisit that. But I think we are going to split this into multiple PRs so you should remember to move this comment to the TreeLB PR.

src/ck-ldb/lb_strategy.h
```diff
@@ -0,0 +1,235 @@
...
+
+  struct obj_1_data { float load; };
+  template <int N> struct obj_N_data { float load[N]; float maxload; };
+
+  template <int N, bool multi = (N > 1)>
+  class Obj : public std::conditional<multi, obj_N_data<N>, obj_1_data>::type
+  {
+  public:
+
+    unsigned int id;
```
matthiasdiener: Nitpick: `unsigned` is often not a good idea. Is there a reason why `id` is unsigned here (in contrast to Proc's id, for example)?

src/ck-ldb/lb_strategy.h
```diff
@@ -0,0 +1,235 @@
...
+
+    inline float getLoad() const { return this->maxload; }
+
+    inline void assign(const Obj<N> *o) {
+      for (int i=0; i < N; i++) {
+        this->load[i] += o->load[i];
+        if (this->load[i] > this->maxload) this->maxload = this->load[i];
+      }
+    }
+    inline void assign(const Obj<N> &o) { assign(&o); }
```
matthiasdiener: No need for `inline` in header files.

src/ck-ldb/pheap.h
```diff
@@ -0,0 +1,131 @@
+#ifndef PROC_HEAP_H
+#define PROC_HEAP_H
+
+#include <vector>
+
+// custom heap to allow removal of processors from any position
+template <typename P>
+class ProcHeap
```
matthiasdiener: Instead of this custom class, could you use a std::multiset?

rbuch:
This has been superseded by #2527, #2528, #2529, #2530, and #2531.

PR #2513, opened 2019-09-06
Eliminate calls to strndup from charmrun.C

evan-charmworks:
This retains the clean output of ASan when running Charmrun with
++local.


matthiasdiener:
> This retains the clean output of ASan when running Charmrun with
> ++local.

This mostly reverts #2504. Was ASan unclean just do to the missing `ChMessage_free`?

evan-charmworks:
I think yes. There is no functional difference between `substr` and `substr_const`.

PR #2514, opened 2019-09-09
Fix OPTS when rebuilding libmemory and libthreads from tmp

evan-charmworks:
Could merge this for 6.10

### DIFF
--- a/src/scripts/Makefile
+++ b/src/scripts/Makefile
@@ -661,9 +661,9 @@ MEM_DEPS=memory-gnu.C memory-gnu-internal.C converse.h conv-mach.h $(CVHEADERS)
 
 # $1 is the name of the module, $2 is any additional dependency beyond memory.C and $(MEM_DEPS), $3 is extra arguments to charmc
 define libmem
-$(L)/libmemory-$1.a: memory.C $(MEM_DEPS) $2
-	$(CHARMC) -I. -o libmemory-$1.o $3 $$<
-	$(CHARMC) -I. -o $$@ $3 libmemory-$1.o
+$(L)/libmemory-$1.a: memory.C $$(MEM_DEPS) $2
+	$$(CHARMC) -I. -o libmemory-$1.o $3 $$<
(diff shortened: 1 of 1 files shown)


matthiasdiener:
I don't understand this patch. According to the make docs (https://ftp.gnu.org/old-gnu/Manuals/make-3.80/html_node/make_59.html#SEC63)

> The define directive does not expand variable references and function calls in the canned sequence; the `$' characters, parentheses, variable names, and so on, all become part of the value of the variable you are defining.

Isn't that the correct behavior here?

evan-charmworks:
A single `$` will cause the variable to expand at the time of the `define`, while a double `$$` will cause it to expand when the define is `call`ed.

My guess is something to do with this at the bottom of the Makefile:

```
# If OPTS was not defined, use the build time options
export OPTS ?= $(OPTSATBUILDTIME)
ifeq ($(strip $(OPTS)),)
$(warning Variable OPTS is defined to an empty string. Are you sure this is what you want?)
endif
```

matthiasdiener:
> A single `$` will cause the variable to expand at the time of the `define`, while a double `$$` will cause it to expand when the define is `call`ed.

But doesn't this contradict the manual that I quoted above?

I tested this with the following Makefile:

``` makefile
MEM_DEPS=foo_mem_deps
CHARMC=foo_charmc


define libmem
mytarget: $$(MEM_DEPS) $2
	@echo $$(CHARMC) -I. -o libmemory-$1.o $3
	@echo $$(CHARMC) -I. -o $$@ $3 libmemory-$1.o
endef

all: mytarget

foo_mem_deps:
first_arg:
second_arg:
third_arg:

$(eval $(call libmem,first_arg,second_arg,third_arg))
```

Writing `$(MEM_DEPS)` or `$$(MEM_DEPS)` doesn't seem to make a difference. This is with GNU make 3.81.

evan-charmworks:
Needs one more review

PR #2517, opened 2019-09-11
Revert synchronous memory copy operations in GPU Manager to be truly asynchronous

mprobson:
Our previous release of GPU manager changed previously asynchronous memory copy operations from asynchronous to synchronous. This was necessary for correctness but severely deteriorated performance. This PR restores their previous asynchronous state and also adds automatic and robust per-PE stream creation to ensure that the copies happen truly asynchronously.


src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -249,43 +252,74 @@ void GPUManager::init() {
...
-      n_streams_ = 16;
+      new_n_streams = 16;
     else // 6.0 or unknown 6.x
-      n_streams_ = 128;
+      new_n_streams = 128;
   }
   else // unknown (future) compute capability
-    n_streams_ = 128;
+    new_n_streams = 128;
+#if CMK_SMP || CMK_MULTICORE
```
rbuch: I think this empty `#if` should be inverted so that the `#else` can be removed.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -249,43 +252,74 @@ void GPUManager::init() {
...
 
-  streams_ = new cudaStream_t[n_streams_];
-  for (int i = 0; i < n_streams_; i++) {
+  if (new_n_streams <= n_streams_)
+    return n_streams_;
+
+  return createNStreams(new_n_streams);
+}
+
+int GPUManager::createNStreams(int new_n_streams) {
```
rbuch: Does this function need a lock?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -249,43 +252,74 @@ void GPUManager::init() {
...
+  hapiCheck(cudaGetDeviceCount(&device_count));
+  int nodes_per_device = CmiNumPesOnPhysicalNode(0) / device_count;
+  nodes_per_device = nodes_per_device > 0 ? nodes_per_device : 1;
+  new_n_streams =  (new_n_streams + nodes_per_device - 1) / nodes_per_device;
 #endif
 
-  streams_ = new cudaStream_t[n_streams_];
-  for (int i = 0; i < n_streams_; i++) {
+  if (new_n_streams <= n_streams_)
+    return n_streams_;
```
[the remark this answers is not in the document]
mprobson: It is, however this function may be called elsewhere in the future.

mprobson:
All issues should be addressed now. As a general note, my approach has been that functions prefixed with hapi are exposed to the user and meant to be thread safe, thus they require correct locking. Other functions are internal and not necessarily thread safe and are locked when they are called for performance.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -249,43 +252,78 @@ void GPUManager::init() {
...
-  n_streams_ = 4; // FIXME per PE in non-SMP mode
+    new_n_streams = 128;
+#if !CMK_SMP && !CMK_MULTICORE
+  // Allocate total physical streams between GPU managers sharing a device...
+  // i.e. PEs / num devices
+  int device_count;
+  hapiCheck(cudaGetDeviceCount(&device_count));
+  int pes_per_device = CmiNumPesOnPhysicalNode(0) / device_count;
+  pes_per_device = pes_per_device > 0 ? pes_per_device : 1;
+  new_n_streams =  (new_n_streams + pes_per_device - 1) / pes_per_device;
```
rbuch: This splits the pool of streams to different in PEs in non-SMP mode, but what's the expected behavior for SMP mode if the user has more than one process per GPU?

PR #2524, opened 2019-09-18
Redesign Isomalloc to better meet AMPI's needs

evan-charmworks:
Originally, each PE was assigned a section of the global address space
and was responsible for managing it. The new paradigm divides the
available address space by logical units of work, such as AMPI virtual
ranks. This way, each rank can manage itself independently of all others
once memory ranges are assigned, and a rank's underlying memory will
always be contiguous in ownership allowing it to be migrated as a single
buffer.

The nature of Isomalloc requires it to interface with the system using
mmap, directly affecting virtual memory at the level of pages. This
means some form of pooling is required to avoid wasting memory when
servicing malloc calls smaller than the system page size, as well as to
minimize the number of expensive system calls made during runtime. The
new design includes the Isomempool to achieve these goals. Previously
each allocation made using Isomalloc would result in its own call to
mmap, which wasted memory space and computing time. The GNI machine
layer's mempool had also been available to Isomalloc, but it suffered
from correctness and efficiency issues, and unclear code. These two
modes of operation have been removed entirely in favor of the
Isomempool.

The Isomempool uses a red-black tree to store the locations and sizes of
free space within the region provided by the system through mmap. This
tree is used to determine the best placement for user allocations. Each
user allocation is prefixed with a header that points to its immediate
neighbors, making a doubly linked list of all regions. Deallocations
mark a region available but leave its header in place so that an
adjacent deallocation can merge the two empty regions.

This patch also cleans up tlsglobals and swapglobals due to the close
ties between migratable threads and global variable privatization. This
includes their API, how they are initialized, and when they are
(de)activated.


matthiasdiener:
Error from NetLRTS:

```
make[2]: Entering directory '/home/circleci/project/netlrts-linux-x86_64/tests/util'
../../bin/charmc -g -c check.C
../../bin/charmc -g -o check check.o -language converse++
../../bin/../lib/libthreads-default.a(libthreads-default.o): In function `CthInterceptionsDeactivate':
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:578: undefined reference to `CtgUninstall'
../../bin/../lib/libthreads-default.a(libthreads-default.o): In function `CthInterceptionsCreate':
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:594: undefined reference to `CtgGetSize'
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:602: undefined reference to `CtgCreate'
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:594: undefined reference to `CtgGetSize'
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:602: undefined reference to `CtgCreate'
../../bin/../lib/libthreads-default.a(libthreads-default.o): In function `CthInterceptionsActivate':
/home/circleci/project/netlrts-linux-x86_64/tmp/threads.C:562: undefined reference to `CtgInstall'
collect2: error: ld returned 1 exit status
Fatal Error by charmc in directory /home/circleci/project/netlrts-linux-x86_64/tests/util
   Command g++ -rdynamic -g check.o moduleinit5189.o -L../../bin/../lib -lmemory-default -lthreads-default -lconv-machine -lconv-core -ltmgr -lconv-util -lconv-partition -lhwloc_embedded -lm -ltrace-converse -lmemory-default -lthreads-default -lldb-rand -lconv-ldb -lckqt -ldl -lz -lm ../../bin/../lib/conv-static.o -o check returned error code 1
charmc exiting...
Makefile:9: recipe for target 'check' failed
```

stwhite91:
Does this run cleanly with migration under Valgrind? Also worth sanity checking syncft too

evan-charmworks:
> Does this run cleanly with migration under Valgrind? Also worth sanity checking syncft too

Valgrind seems to conflict with `-memory` options in some way. I ran with ubsan and it did not detect anything in Isomalloc code.

I checked `netlrts-linux-x86_64-syncft` by changing `AMPI_INFO_LB_SYNC` to `AMPI_INFO_CHKPT_IN_MEMORY` in the stress test and it was able to successfully recover from terminating a process during execution.

stwhite91:
> Valgrind seems to conflict with `-memory` options in some way. I ran with ubsan and it did not detect anything in Isomalloc code.

I thought I had run with Isomalloc and Valgrind before in the past, but perhaps not

evan-charmworks:
Ready for review

src/conv-core/isomalloc.C
```diff
@@ -83,2113 +73,2072 @@ static int _mmap_probe = 0;
...
+    size_t sizefield;
+    BitCarrier occupiedfield;
+  };
 
-static void delete_list_array(slotset *ss)
-{
-  for (int i = 0; i < LIST_ARRAY_SIZE; i++)
+  using Occupied = RegionHeader::Occupied;
+
+  struct RBTree
```
matthiasdiener: std::map is almost always implemented as an rbtree. Could this be a simpler alternative?

evan-charmworks: Using `std::map` or other STL containers would introduce calls to system malloc/free, which would have a performance impact. Is it semantically appropriate for an allocator to call another allocator on the critical path? Implementing our own structure has allowed me to tweak it to the needs of the mempool. For example, moving the RBTree code into the class definition so it could be inlined by the compiler gave a 9% speedup. Additionally, migration does not need to reconstruct the tree structure, since the tree is embedded in the mmapped region.

matthiasdiener: Ok. It might be useful to add this as a comment somewhere.

evan-charmworks: Added as a comment

src/conv-core/memory-isomalloc.C
```diff
@@ -85,21 +80,21 @@ static void *meta_malloc(size_t size)
         int _isomalloc_thread = isomalloc_thread;
         if (CmiThreadIs(CMI_THREAD_IS_TLS)) _isomalloc_thread = 1;
 #endif
-	if (meta_inited && CpvInitialized(isomalloc_blocklist) && CpvAccess(isomalloc_blocklist)
+	if (meta_inited && CpvInitialized(isomalloc_state) && CpvAccess(isomalloc_state).context && !CpvAccess(isomalloc_state).disabled
```
stwhite91: Could pull this long if-statement out into a function returning a bool. It's repeated below

evan-charmworks: Done

doc/converse/manual.rst
```diff
@@ -1459,77 +1459,6 @@ Returns the number of children of nodeNum in the spanning tree.
 This function fills the array children with node numbers of children of
 nodeNum in the spanning tree.
 
-Isomalloc
```
stwhite91: Why are we removing this entirely? Shouldn't it be updated?

evan-charmworks: Readded the documentation and updated it based on the changes in the PR.

PR #2527, opened 2019-09-25
[lbMerge - 1] Remove old, extraneous load balancers

rbuch:


epmikida:
Looks good to me, other than the build error, but would this also be a good patch to rename DistributedLB to GrapevineLB?

rbuch:
> Looks good to me, other than the build error, but would this also be a good patch to rename DistributedLB to GrapevineLB?
 
Yeah, could be, but I think I'm more pro having another commit/PR that is responsible for renaming DistributedLB.

src/scripts/Makefile
```diff
@@ -7,6 +7,7 @@
 #     * Charm++ runtime
 #     * ITC++ Translator
 #     * Languages
+#     * Documentation
```
evan-charmworks: Should this line be here?

rbuch: No, good catch, I'll take it out.

PR #2531, opened 2019-09-25
[lbMerge - 5] Add TreeLB and several lb strategies dependent on it

rbuch:
TreeLB is meant to replace CentralLB and HybridBaseLB. It does
centralized load balancing when configured as a 2-level tree.
It can work with multiple load balancing strategies, which can be
implemented in a separate source file and then added to
strategy_factory.h

This replaces the old common lb strategies like Greedy,
GreedyRefine and Refine with their new TreeLB-based
implementations.


src/ck-ldb/Make.lb
```diff
@@ -1,42 +1,17 @@
 # Automatically generated by script Makefile_lb.sh
 COMMON_LDBS=\
-   DummyLB \
-   GreedyLB \
-   GreedyRefineLB \
-   RandCentLB \
-   RefineLB \
-   RotateLB \
+   TreeLB \
```
matthiasdiener: These changes will need to be reflected in the make build system as well

src/ck-ldb/lb_strategy.h
```diff
@@ -0,0 +1,235 @@
...
+
+  struct obj_1_data { float load; };
+  template <int N> struct obj_N_data { float load[N]; float maxload; };
+
+  template <int N, bool multi = (N > 1)>
+  class Obj : public std::conditional<multi, obj_N_data<N>, obj_1_data>::type
+  {
+  public:
+
+    unsigned int id;
```
matthiasdiener: `unsigned` is often not a good idea. Is there a reason why `id` is unsigned here (in contrast to Proc's id, for example)?

src/ck-ldb/lb_strategy.h
```diff
@@ -0,0 +1,235 @@
...
+
+  // --------- Proc rateAware=false specializations ---------
+
+  template <int N, bool multi>
+  class Proc<N,false,multi> : public std::conditional<multi, proc_N_data<N>, proc_1_data>::type
+  {
+  public:
+    int id = -1;
+
+    inline void populate(int _id, float *_bgload, float *_speed) {
```
matthiasdiener: No need for `inline` in header files.

src/ck-ldb/pheap.h
```diff
@@ -0,0 +1,131 @@
+#ifndef PROC_HEAP_H
+#define PROC_HEAP_H
+
+#include <vector>
+
+// custom heap to allow removal of processors from any position
```
matthiasdiener: Instead of this custom class, could you use a `std::multiset`?

juanjgalvez: I need a heap, not a fully sorted container. But I guess it really comes down to which one is faster in practice (I haven't tried multiset).

src/ck-ldb/tree_level.ci
```diff
@@ -0,0 +1,35 @@
...
+    int num_incoming[];
+    int obj_start[];
+    int to_pes[];
+  };
+
+  message SubtreeLoadMsg {
+  };
+
+  message SubtreeMigrateDecisionMsg {
+    // FIXME? I don't think this layout is ideal, but I don't know if ci files
```
stwhite91: I think we do support arrays of any PUPable type inside messages, and it should be just as efficient as integral types so long as the type is declared as PUPbytes

juanjgalvez:
The content of this PR might need another cleanup pass from me. But I'm currently focused on getting all the tests/examples to pass.

src/ck-ldb/idm.h
```diff
@@ -0,0 +1,23 @@
+#include <vector>
+#include <unordered_map>
+#include <pup_stl.h>
+
+class IDM
```
epmikida: What is IDM?

juanjgalvez: Have to either rename or add comments

juanjgalvez:
> Did I miss strategies like Rotate and Random, or are those just coming in future patches?

Random, Rotate and Dummy are defined at the bottom of `lb_strategy.h`. They are so small and presumably not relevant for most users that I just put them in there instead of making separate files for each, but we could also move them out.

src/ck-ldb/lb_strategy.h
juanjgalvez: For objects, this ID is basically the index in the vector of objects passed to lb strategies. It is not related to 64 bit ID of chares. So, it is unsigned because it goes from 0 to numobjs-1. But I agree that `unsigned int` is not ideal. It should probably need to be typedef'd to some unsigned type though. The size of the data type will determine how many objects a load balancing strategy can accept. I'm not sure if we need a 64-bit type. It could make running a strategy like greedy with millions of objects slightly slower compared to 32-bit. Also, very big jobs would presumably be using a hierarchical mode of TreeLB so that top-level strategies would not deal with many millions of objects.

juanjgalvez: Proc has a signed ID because it coincides with the PE number, which is `int`, and also if I remember correctly the PE number of some objects can be unknown if the migrate from one subtree to another.

src/ck-ldb/LBManager.C
```diff
@@ -694,9 +715,24 @@ void LBManager::nextLoadbalancer(int seq)
...
+      config["Root"]["strategies"] = {"GreedyRefine"};
+      config["Process"]["strategies"] = {"GreedyRefine"};
+    } else {
+      config["tree"] = "PE_Root";
+      config["Root"]["pe"] = 0;
+      config["Root"]["strategies"] = {lbNames[switchTo]};
+    }
+    configureTreeLB(config);
+  } else {
+    //TODO: Implement turn off / on for Distributed
```
stwhite91, line 734: Is this in the pipeline of LB patches or should an issue be made for it?

examples/armci/putTest/Makefile
```diff
@@ -4,10 +4,10 @@ CHARMC=../../../bin/charmc $(OPTS)
 all: put
 
 put: put.c
-	$(CHARMC) -o put put.c -language armci -balancer RotateLB 
+	$(CHARMC) -o put put.c -language armci -balancer TreeLB
```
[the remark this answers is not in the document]
rbuch: Technically yes, but many of the load balancers are no longer what they were before. GreedyLB, GreedyRefineLB, RefineLB, RandCentLB, DummyLB, and RotateLB are now substrategies of TreeLB. You can't link with those specifically, you instead have to link with `-balancer TreeLB` and pass in the one you want as a `-balancer` on the runtime line.

TreeLB is also included in `EveryLB` and `CommonLBs`.

src/ck-core/init.C
```diff
@@ -1483,6 +1483,7 @@ void _initCharm(int unused_argc, char **argv)
 		_registerCkFutures();
 		_registerCkArray();
 		_registerLBManager();
+    _registertree_level();
```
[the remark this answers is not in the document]
rbuch: Fixed.

src/ck-ldb/LBManager.C
```diff
@@ -211,10 +214,32 @@ void _loadbalancerInit()
...
-        lbRegistry.addRuntimeBalancer(balancer); /* lbRegistry is a static */
+    if (CkMyRank() == 0) {
+      bool TreeLB_registered = false;
+      while (CmiGetArgStringDesc(argv, "+balancer", &balancer, "Use this load balancer")) {
+        if (strcmp(balancer, "GreedyLB") == 0)
+          _lb_args.legacyCentralizedStrategies().push_back("Greedy");
+        else if (strcmp(balancer, "GreedyRefineLB") == 0)
+          _lb_args.legacyCentralizedStrategies().push_back("GreedyRefine");
+        else if (strcmp(balancer, "RefineLB") == 0)
+          _lb_args.legacyCentralizedStrategies().push_back("RefineA");
```
epmikida, line 225: RefineA? Typo?

src/ck-ldb/LBManager.C
```diff
@@ -694,9 +715,24 @@ void LBManager::nextLoadbalancer(int seq)
 }
 
 // switch strategy
-void LBManager::switchLoadbalancer(int switchFrom, int switchTo)
-{
-  // TODO: Implement turn off / on
+void LBManager::switchLoadbalancer(int switchFrom, int switchTo) {
```
epmikida, line 718: Why do we need switchFrom? Is this meant to be exposed to the application? If so, how would the application know the ints to pass? Could it use strings instead?

src/ck-ldb/LBManager.C
```diff
@@ -764,6 +800,28 @@ void LBManager::pup(PUP::er& p)
   }
 }
 
+extern "C" void configureTreeLB(const char *json_str) {
```
epmikida, line 803: Didn't we remove all usage of extern "C" a while back?

src/ck-ldb/LBManager.h
```diff
@@ -34,14 +36,13 @@ class CkLBArgs
...
   int _lb_traceComm;            // stats collection for comm
   int _lb_central_pe;           // processor number for centralized startegy
-  int _lb_percentMovesAllowed;  // Specifies restriction on num of chares to be moved(as a
-                                // percentage of total number of chares). Used by RefineKLB
-  int _lb_teamSize;        // specifies the team size for TeamLB
-  int _lb_maxDistPhases;   // Specifies the max number of LB phases in DistributedLB
-  double _lb_targetRatio;  // Specifies the target load ratio for LBs that aim for a
-                           // particular load ratio
+  int _lb_teamSize;		// specifies the team size for TeamLB
+  int _lb_maxDistPhases;  // Specifies the max number of LB phases in DistributedLB
```
epmikida, line 40: Maybe not relevant for this patch, but it might be better to have each LB define it's own options. Otherwise this list is just going to get bloated.

src/ck-ldb/LBManager.h
```diff
@@ -59,28 +59,29 @@ class CkLBArgs
...
-  inline int& central_pe() { return _lb_central_pe; }
-  inline double& alpha() { return _lb_alpha; }
-  inline double& beta() { return _lb_beta; }
-  inline int& percentMovesAllowed() { return _lb_percentMovesAllowed; }
-  inline int& maxDistPhases() { return _lb_maxDistPhases; }
-  inline double& targetRatio() { return _lb_targetRatio; }
-  inline int& metaLbOn() { return _lb_metaLbOn; }
+  inline char*& treeLBFile() { return _lb_treeLBFile; }
+  inline double & lbperiod() { return _autoLbPeriod; }
+  inline int & debug() { return _lb_debug; }
```
epmikida, line 64: Is this (space before &) the style we use? I assume you'll be doing a clangformat on this stuff anyways, so probably not relevant.

src/ck-ldb/LBManager.C
epmikida: I see now there is refine A and refine B. Do we need two? Shouldn't their names be more useful, or possibly merged into one strategy with different configuration options?

examples/armci/putTest/Makefile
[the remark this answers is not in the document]
stwhite91: We could potentially add aliases for users to link against where the old names alias to TreeLB. But you would still lose the feature of not having to pass +balancer on the command line after linking with a specific balancer

src/ck-ldb/LBManager.C
stwhite91: If it's needed at all if should probably be CMIEXTERNC or whatever our macro for that is

src/ck-ldb/LBManager.h
rbuch: Yes.

src/ck-ldb/LBManager.C
rbuch: It's needed to support non-TreeLBs. This function is used by metabalancer to switch to newly predicted balancers. When switching between two different LBs (note that switching between different strategies of TreeLB, it's not needed since they're both part of the same LB), it has to call `turnOff` and `turnOn`to do setup/cleanup work. So this is lying in wait for a more complete implementation of the metabalancer stuff. Metabalancer should know the ints, it could use strings, but it doesn't need to since this is internal stuff, anyway.

src/ck-ldb/LBManager.C
rbuch: #2818

src/ck-ldb/pheap.h
rbuch: I think it's worth keeping @juanjgalvez's implementation because of the constant time element lookup (at the cost of extra storage), at least for now. I haven't benchmarked it, but I will in the future (there are other places where non-standard data structures are used, I think this should be lumped in with the effort of checking if those should be replaced).

src/ck-ldb/lb_strategy.h
rbuch: I've updated this to use regular `int`, since that's what the type of the incoming value that set this is.

src/ck-ldb/LBManager.C
rbuch: #2818

src/ck-ldb/LBManager.C
rbuch: Fixed.

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,369 @@
...
+
+  virtual uint8_t build(std::vector<LevelLogic*>& logic, std::vector<int>& comm_parent,
+                        std::vector<std::vector<int>>& comm_children,
+                        std::vector<LevelLogic*>& comm_logic, json& config)
+  {
+    uint8_t L = 2;  // num levels
+    reset(L, logic, comm_parent, comm_children, comm_logic);
+
+    int rootPE = 0;
+    try
```
stwhite91, line 46: Can we avoid using exceptions? We have generally avoided using them entirely in the runtime

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,552 @@
...
+  std::vector<TreeLBMessage*> decisions;
+  /// CkMessage *inter_subtree_migrations = nullptr;
+  IDM idm;
+  logic[level]->loadBalance(decisions, idm);
+  if (idm.numDests() > 0)
+  {
+    // this can happen when final destinations of chares has been decided,
+    // and chares from a subtree need to migrate to a PE in a different subtree
+    std::vector<int> idm_dests;
+    idm_dests.reserve(1 + idm.numDests());
```
stwhite91, line 306: Better to pass the size to std::vector's constructor above and to assign directly to idm_dests[i] below

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,552 @@
...
+    CkPrintf("--------- Finished LB step %d ---------\n", lbmgr->step());
+  }
+
+  // Increment to next step
+  lbmgr->incStep();
+
+  LBManager::Object()
+      ->MigrationDone();  // call registered callbacks. not sure what this is for
+
+  // again, TreeLB shouldn't be doing these things. it should just notify LBManager
```
stwhite91, line 507: Do we have an open issue or issues for the various todo's mentioned in comments here?

src/ck-ldb/TreeLevel.h
```diff
@@ -0,0 +1,1236 @@
...
+#include "TopoManager.h"
+#include "TreeLB.h"
+#include "TreeStrategyBase.h"
+#include "TreeStrategyFactory.h"
+#include <cmath>
+#include <limits>  // std::numeric_limits
+
+#define FLOAT_TO_INT_MULT 10000
+
+// TODO rateAware is not implemented, that is: collecting processor speeds
```
stwhite91, line 13: Is this true? Isn't that a step backwards?

src/ck-ldb/TreeLevel.h
```diff
@@ -0,0 +1,1236 @@
...
+      if (repeat_strategies) current_strategy = 0;
+    }
+    else
+    {
+      current_strategy++;
+    }
+    decisions.resize(num_children);
+    decisions[0] = migMsg;
+    for (int i = 1; i < num_children; i++)
+      decisions[i] = (TreeLBMessage*)CkCopyMsg((void**)&migMsg);
```
stwhite91, line 884: Could this be CmiReference instead of CkCopyMsg?

src/ck-ldb/idm.h
```diff
@@ -0,0 +1,18 @@
...
+  ~IDM() {}
+
+  void pup(PUP::er& p) { p | data; }
+
+  size_t numDests() { return data.size(); }
+
+  // pe -> vector of (obj_local_id, dest)
+  // this map stores, for a pe, the list of objects that need to migrate to a destination
+  // PE in a different subtree
+  std::unordered_map<int, std::vector<std::pair<int, int>>> data;
```
stwhite91, line 17: 'dests' might be a better name for this?

src/ck-ldb/pheap.h
```diff
@@ -0,0 +1,145 @@
...
+template <typename P>
+class ProcHeap
+{
+ public:
+  ProcHeap(std::vector<P>& procs)
+  {
+    Q.reserve(procs.size() + 1);
```
stwhite91, line 13: resize() and direct assignment below would be better than reserve and push_back

src/ck-ldb/LBManager.h
```diff
@@ -28,20 +30,21 @@ class CkLBArgs
...
-  int _lb_useCpuTime;           // use cpu instead of wallclock time
-  int _lb_statson;              // stats collection
-  int _lb_traceComm;            // stats collection for comm
-  int _lb_central_pe;           // processor number for centralized startegy
-  int _lb_percentMovesAllowed;  // Specifies restriction on num of chares to be moved(as a
-                                // percentage of total number of chares). Used by RefineKLB
-  int _lb_teamSize;        // specifies the team size for TeamLB
+  int _lb_samePeSpeed;     // ignore cpu speed
+  int _lb_testPeSpeed;     // test cpu speed
+  int _lb_useCpuTime;      // use cpu instead of wallclock time
```
stwhite91, line 35: Some of these can/should be 'bool'

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,369 @@
...
+    {
+      step_freq_lvl2 = 1;
+    }
+
+    // node level (level 1)
+    lvl = 1;
+    if (mype == level1root)
+    {
+      std::vector<int> pes_in_node;
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 159: This loop can be replaced with a call to std::iota

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+
+    // PE level (level 0)
+    int lvl = 0;
+    logic[lvl] = new PELevel(lbmgr);
+
+    // set up comm-tree between levels 0 and 1
+    if (mype == level1root)
+    {
+      // pes in my node excluding me
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 128: Can be replaced with two calls to std::iota() or a call to std::generate() or something

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      std::iota(pes_in_node.begin(), pes_in_node.end(), CkNodeFirst(mynode));
+
+      NodeLevel* level = new NodeLevel(lbmgr, pes_in_node);
+      level->configure(false, config["Process"], step_freq_lvl2);
+      logic[lvl] = level;
+
+      // set up comm-tree between levels 1 and 2
+      std::vector<int> level1_roots;
+      level1_roots.push_back(level2root);  // need this for getPETopoTreeEdges
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 158: I think we can replace this with a call to std::generate() to avoid a loop over all PEs

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+
+    // PE level (level 0)
+    int lvl = 0;
+    logic[lvl] = new PELevel(lbmgr);
+
+    // set up comm-tree between levels 0 and 1
+    if (mype == level1root)
+    {
+      // pes in my node excluding me
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 235: Same here

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      if (step_freq_lvl3 % step_freq_lvl2 != 0)
+        CkAbort("step_freq of Root level is not multiple of previous level\n");
+    }
+
+    // node level (level 1)
+    lvl = 1;
+    if (mype == level1root)
+    {
+      std::vector<int> pes_in_node;
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 264: And here

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      {
+        if (CkNodeOf(pe) == mynode) pes_in_node.push_back(pe);
+      }
+      NodeLevel* level = new NodeLevel(lbmgr, pes_in_node);
+      level->configure(false, config["Process"], step_freq_lvl2);
+      logic[lvl] = level;
+
+      // set up comm-tree between levels 1 and 2
+      std::vector<int> level1_roots;
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 274: And here

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      }
+      comm_logic[lvl] = new MsgAggregator(num_children);
+    }
+
+    // nodeset level (level 2)
+    lvl = 2;
+    if (mype == level2root)
+    {
+      std::vector<int> pes_in_group;
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 297: And here

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      {
+        if (GroupOf(pe) == mygroup) pes_in_group.push_back(pe);
+      }
+      NodeSetLevel* level = new NodeSetLevel(lbmgr, pes_in_group);
+      level->configure(false, config["ProcessGroup"], step_freq_lvl3);
+      logic[lvl] = level;
+
+      /*if (mype == level3root) {
+        for (int pe=0; pe < CkNumPes(); pe++) {
```
stwhite91, line 306: And here?

src/ck-ldb/TreeBuilder.h
```diff
@@ -0,0 +1,348 @@
...
+      }*/
+      if (mype != level3root)  // NEW
+        comm_parent[lvl] = level3root;
+    }
+
+    // root of tree (level 3)
+    lvl = 3;
+    if (mype == level3root)
+    {
+      for (int pe = 0; pe < CkNumPes(); pe++)
```
stwhite91, line 321: And here

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,549 @@
...
+}
+
+void TreeLB::staticObjMovedIn(void* data, LDObjHandle h, bool waitBarrier)
+{
+  //  ((TreeLB*)data)->objMovedIn(h, waitBarrier);
+}
+
+void TreeLB::Migrated(int waitBarrier)
+{
+  bool barrier = false;
```
stwhite91, line 22: This function can just be a call to 'objMovedIn((bool)waitBarrier);'

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,549 @@
...
+
+  std::vector<TreeLBMessage*> decisions;
+  /// CkMessage *inter_subtree_migrations = nullptr;
+  IDM idm;
+  logic[level]->loadBalance(decisions, idm);
+  if (idm.size() > 0)
+  {
+    // this can happen when final destinations of chares has been decided,
+    // and chares from a subtree need to migrate to a PE in a different subtree
+    std::vector<int> idm_dests;
```
stwhite91, line 302: It would be better to construct the vector to size, then assign directly to it in the for loop

src/ck-ldb/TreeLevel.h
rbuch: I don't think there's a good way to do this with Charm messages. Things get wonky if you try to send them more than once.

Do you know if there's a way to get that to work with `CmiReference`? Also, note that these are variable sized messages, so they need to be allocated in a specific way.

stwhite91: I think you can do something like 'CmiReference(UsrToEnv(mig));' but it might not work in this case.

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,537 @@
...
+    if (logic->numStatsReceived() == comm_children[level].size() + 1)
+    {
+      TreeLBMessage* newMsg = logic->mergeStats();
+      newMsg->level = level;
+      thisProxy[comm_parent_pe].sendStatsUp((CkMessage*)newMsg);
+    }
+  }
+}
+
+void TreeLB::receiveStats(TreeLBMessage* stats, int level)
```
stwhite91, line 248: It's kind of weird and I think inefficient to send a message as a part of a parameter marshalled entry method. I assume the runtime has to copy that message into another message. If you made level the refNum or added another member to the message you could get around that.

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,537 @@
...
+    receiveDecision(decisions[0], level);
+    for (int i = 0; i < children.size(); i++)
+    {
+      decisions[i + 1]->level = level;
+      thisProxy[children[i]].sendDecisionDown((CkMessage*)decisions[i + 1]);
+    }
+  }
+}
+
+void TreeLB::receiveDecision(TreeLBMessage* decision, int level)
```
stwhite91, line 367: Same here

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    // if outgoing > 0 and we are in last level (0), concrete objects are moved to
+    // concrete PEs. for chares whose destination is known, the logic object would have
+    // made call to migrate them inside processDecision (above) NOTE: I might be moving
+    // some objects to other subtrees, might need to wait to receive messages from those
+    // subtrees telling me to which PEs to migrate them
+  }
+  // fprintf(stderr, "[%d] TreeLB::receiveDecision, deleting decision msg\n", CkMyPe());
+  delete (CkMessage*)decision;  // FIXME this won't call the subclass' delete method
```
stwhite91, line 403: Does this leak memory?

src/ck-ldb/TreeLB.C
```diff
@@ -0,0 +1,537 @@
...
+  // a check here is needed because PE might have been sending chares/tokens or
+  // receiving them even before receiving the decision msg (with the final tally)
+  // from the parent. So, exchange might have already completed
+  checkLoadExchanged(level);
+}
+
+// order transfer of load token to dest
+void TreeLB::transferLoadToken(TreeLBMessage* transferMsg, int level)
```
stwhite91, line 411: Same issue with parameter marshalling with a message argument

src/ck-ldb/TreeLevel.h
rbuch: I'll leave this as future work, I've been trying to implement it over the past week and it's been full of thorns, it's a bit complicated because when the messages are sent, we can't assume that they're all the same message when they're sent because (while it's not implemented yet), some parts of the infrastructure may create different messages for each of the children, so using section multicast or multicast converse sends won't work. That coupled with actually making sure the messages get freed has complicated things, causing ref counting errors in CmiFree. I think some scheme with `CmiReference` can probably work, but I don't think it should stop get in the way of merging this. I've made #2857 for this.

src/ck-ldb/TreeLB.C
rbuch: This (and the other ones you've noted below) should be fine, they aren't entry methods, just regular methods, so there isn't any of this extra copying.

src/ck-ldb/LBManager.h
rbuch: Fixed.

src/scripts/configure.ac
```diff
@@ -1441,6 +1441,20 @@ test_linkc "whether synchronization primitives (__sync_synchronize) works in C"
 AC_DEFINE_UNQUOTED(CMK_C_SYNC_SYNCHRONIZE_PRIMITIVE, $pass, [whether sync_synchronize primitives works in C])
 
 
+### test for _IO_file_overflow ###
```
matthiasdiener, line 1444: Is this test also include in the cmake build system?

evan-charmworks: It was in this commit: https://github.com/UIUC-PPL/charm/pull/2531/commits/2121c83aa4f1a6996f70bcdd7355d6ff2cff258d It looks like it needs to be re-added.

src/ck-ldb/TreeLB.C
rbuch: No, it just wouldn't call a chain of destructors, which don't even exist in any of the message types as far as I can tell. I'm not sure what @juanjgalvez's concern here was.

src/scripts/configure.ac
rbuch: Sorry, I accidentally removed it when I was resolving a merge conflict. I've added it back in now in d8f8cfd.

src/ck-ldb/TreeLevel.h
rbuch: Fixed in #2871.

examples/armci/putTest/Makefile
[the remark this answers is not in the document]
stwhite91: I think we should support this backwards compatibility of linking with the old names that are still supported by TreeLB. It should be easy to support, and will avoid breaking a bunch of old Makefiles needlessly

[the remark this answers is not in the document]
rbuch: Fixed in ab7f437c2bb82361207765a99129ccb4be11fe6d.

matthiasdiener:
Changa doesn't build:

```
HierarchOrbLB.cpp:10:10: fatal error: GreedyLB.h: No such file or directory
 #include "GreedyLB.h"
          ^~~~~~~~~~~~
````

**Comment by rbuch** [comment]
> Changa doesn't build:
> 
> ```
> HierarchOrbLB.cpp:10:10: fatal error: GreedyLB.h: No such file or directory
>  #include "GreedyLB.h"
>           ^~~~~~~~~~~~
> ```

Fixed in https://github.com/N-BodyShop/changa/pull/72.

File: src/ck-ldb/LBManager.C [hunk 39]
```diff
@@ -445,7 +474,8 @@ void LBManager::initnodeFn()
 void LBManager::callAt()
 {
   localBarrier.CallReceivers();
-  if (nloadbalancers > 0) loadbalancers[0]->InvokeLB();
+  //TODO: Add support for sequencing multiple LBs
+  if (loadbalancers.size() > 0) loadbalancers[0]->InvokeLB();
```
kavithachandrasekar, line 478: Should we fix this in this PR to support calling multiple LBs or make a note in documentation to say that using multiple +balancer option is temporarily not supported?

rbuch: We should add support for multiple LBs, but I think that can go into a subsequent PR, if that's alright; I don't want to delay this one anymore.

PR #2534, opened 2019-09-25
Fix -charm-shared linking on macOS

evan-charmworks:
This fixes the install_name of libckqt.dylib and libampiromio.dylib.

Can be tested with `./build AMPI netlrts-darwin-x86_64 -j8 --build-shared -charm-shared -g3 && cd netlrts-darwin-x86_64/tests/ampi/privatization && make OPTS="-charm-shared -g3" -j8 control && ./cxx-control +p1`


src/scripts/charmc
```diff
@@ -1143,10 +1143,18 @@ if [ "$BUILD_SHARE" = "1" ]
 then
 	# ignore BUILD_SHARE if the target is not .a or .so
 	case "$OBJECT" in
-	*.a|*.so|*.so.*|*.sl|*.dylib)
+	*.a)
 		OPTS_LD="$OPTS_LD $CMK_LD_SHARED $OPTS_LDRO"
 		[ -n "$CHARMLIBSO" ] && OPTS_LD="$OPTS_LD -L$CHARMLIBSO"
```
matthiasdiener: this line seems unnecessary now.

evan-charmworks: It still is because of how `-build-shared` creates both .so and .a files when building .a files.

CMakeLists.txt
```diff
@@ -675,7 +675,7 @@ endif()
 set(optfile_mak ${CMAKE_BINARY_DIR}/include/conv-mach-opt.mak)
 
 file(WRITE  ${optfile_mak} "# Build-time options header, automatically generated by cmake\n")
-foreach(l BUILD_CUDA CMK_AMPI_WITH_ROMIO CMK_BLUEGENEQ CMK_BUILD_PYTHON CMK_CHARMDEBUG CMK_COMPILER CMK_GDIR CMK_HAS_MALLOC_HOOK CMK_HAS_MMAP CMK_LIBJPEG CMK_LUSTREAPI CMK_MULTICORE CMK_UTH CMK_NO_BUILD_SHARED CMK_NO_PARTITIONS CMK_SHARED_SUF CMK_SMP CMK_SUPPORTS_FSGLOBALS CMK_SUPPORTS_PIPGLOBALS CMK_TRACE_ENABLED CMK_USE_LRTS CMK_VDIR OPTSATBUILDTIME)
+foreach(l BUILD_CUDA CMK_AMPI_WITH_ROMIO CMK_MACOSX CMK_BLUEGENEQ CMK_BUILD_PYTHON CMK_CHARMDEBUG CMK_COMPILER CMK_GDIR CMK_HAS_MALLOC_HOOK CMK_HAS_MMAP CMK_LIBJPEG CMK_LUSTREAPI CMK_MULTICORE CMK_UTH CMK_NO_BUILD_SHARED CMK_NO_PARTITIONS CMK_SHARED_SUF CMK_SMP CMK_SUPPORTS_FSGLOBALS CMK_SUPPORTS_PIPGLOBALS CMK_TRACE_ENABLED CMK_USE_LRTS CMK_VDIR OPTSATBUILDTIME)
```
matthiasdiener: This probably won't work, as CMK_MACOSX isn't defined by the cmake build system. Imho, a better solution would be to check uname in the AMPI Makefile instead.

evan-charmworks: This change only serves to keep the CMake and autotools build systems in sync. Shared linking is broken for CMake anyway due to hwloc, and this patch does not touch AMPI's CMakeLists.txt to fix ROMIO.

PR #2539, opened 2019-09-26
Add .zenodo.json

rbuch:
This file is read by Zenodo and allows us to explicitly specify metadata like the author list and ordering rather than having it just try to figure it out.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -4,7 +4,7 @@
     "creators": [
         {
             "orcid": "0000-0001-9673-8445", 
-            "affiliation": "University of Illiinois", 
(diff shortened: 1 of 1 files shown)


.zenodo.json
```diff
@@ -0,0 +1,242 @@
+{
+    "description": "The Charm++ parallel programming system.", 
+    "upload_type": "software", 
+    "creators": [
+        {
+            "orcid": "0000-0001-9673-8445", 
+            "affiliation": "University of Illiinois",
```
evan-charmworks: typo on this line

rbuch: Good catch, fixed.

Issue #1485, opened 2017-04-06
CharmDebug in SMP mode does not work


bilgeacun:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1485*

---
It has been broken since at least Charm 6.6.0.

bilgeacun:
*Original date: 2017-04-06 15:13:03*

---
It simply freezes when you press "Start", nothing else is printed.

ericjbohm:
*Original date: 2017-04-06 22:24:40*

---
See the bohm/charmdebugpatch branch for a hacky workaround to that hang.

bilgeacun:
*Original date: 2017-04-13 18:32:02*

---
bohm/charmdebugpatch branch still freezes with some ppn values, for example this one for tests/charm++/simplearrayhello: ./charmdebug  ./hello +p8 10 ++ppn 2

I have found this comment from Abhishek on TMS:
```

Tried to see what are the issues when using CharmDebug with SMP. Eventually emailed Filippo.
Response from Filippo regarding CharmDebug for SMP mode:
"I don't recall having made documentation for SMP version of CharmDebug, but I also never really explored that direction. This project will be quite an extensive one, I believe. There are quite a few things to fix and design decisions to make. Let me just give you a brainstorming of the items I see missing/problematic:
1) Many variables are not thread safe (need to use more Ckpv/Cpv) primarily in ck-core/debug-charm.C and conv-core/memory-charmdebug.c
2) You need to come up with how the UI should display SMP nodes (this abstraction is totally missing from the java app)
3) You need to design how you plan to track memory allocation in SMP (note that I don't think charmdebug can run without -memory charmdebug...): per core? per node? I thought that memory charmdebug was built uniquely on top of memory-gnu, but apparently I already ported it to work on memory-os, so that is good since memory-gnu should be probably avoided. But remember that nowadays the allocation is done in quite a fancy manner so you need to take that into consideration when you redesign the memory-related facilities (leak detection/scanning/etc.)
4) Most of the Cpd getter utilities should not pose too much of a problem (just need to be extended a bit)
5) Tools like the provisional delivery may need to be disabled. This tool makes use of fork, which would not work in an SMP scenario (at least not easily). This tool does not work under bigsim as well. "


```

bilgeacun:
*Original date: 2017-08-01 21:16:20*

---
Changing this into a Feature since CharmDebug never worked/implemented for SMP.

ericjbohm:
[...]

PR #2540, opened 2019-09-26
bugfix (#1485): fix charmdebug breakpoints in SMP mode

ericjbohm:
Several places in cpd startup constrained initialization to
CmiMyRank==0 based on defunct non-smp assumptions. The
restrictions for CmiRank 0 in cpd and ccs have been cut down
to the minimum.

Likewise the cmiArgDebugFlag has been refined to only
indicate that the debugger is attached. Code that formerly relied
on the (now false) assumption that it was only set on CmiRank 0
now tests for that condition explicitly. +cpd is now parsed
earlier to set cmiArgDebugFlag early enough for CmiTimerInit
to use it.

The barriers in CmiTimerInit caused a startup hang when the
debugger is attached.  Those are now disabled by cmiArgFlagDebug.
This does mean that timers may slip slightly when charmdebug
is attached, however, that is unlikely to create trouble as the
perturbations from charmdebug itself will be many orders of magnitude
greater than that slip.

atomic<int> breakPointSet has been added to the EntryInfo struct
(when charmdebug is enabled) to safely handle entry table
modifications for entry point breakpoints.

This also includes a bunch of commented out print statements,
because debugging a debugger leaves you with a dependency on
bare bones tools.


src/ck-core/register.h
```diff
@@ -75,6 +75,7 @@ class EntryInfo {
       message type but store different data in it, like parameter
       marshalled messages.
```
stwhite91: Maybe add an explanation of this variable to the comment here

src/ck-core/init.C
```diff
@@ -1558,9 +1558,9 @@ void _initCharm(int unused_argc, char **argv)
 	}
 
 	/* The following will happen on every virtual processor in BigEmulator, not just on once per real processor */
-	if (CkMyRank() == 0) {
+	//	if (CkMyRank() == 0) {
```
stwhite91: What's the use of this? We should avoid adding commented-out code that is not debug print statements, and ideally all debug print statements would be controlled by macros rather than commented out

ericjbohm: Good catch, i had commented it out while determining the necessary extent of Rank 0 restrictions.  I'll add a fixup to remove that line.

src/ck-core/register.h
ericjbohm: added in fixup

src/ck-core/init.C
ericjbohm: added in fixup

src/ck-core/init.C
```diff
@@ -1569,7 +1568,7 @@ void _initCharm(int unused_argc, char **argv)
 #if CMK_CHARMDEBUG
 	CpdFinishInitialization();
 #endif
-	if (CkMyRank() == 0)
+	//	if (CkMyRank() == 0)
```
stwhite91: Could remove this?

ericjbohm: done

src/ck-core/debug-charm.C
```diff
@@ -866,6 +874,12 @@ void CpdContinueFromBreakPoint ()
           
         }
     }
+    /*        else
```
evan-charmworks: This could be `#if 0` instead of a block comment

src/ck-core/debug-charm.C
```diff
@@ -930,10 +950,17 @@ void CpdRemoveBreakPoint (char *msg)
       if (CkpvAccess(_debugEntryTable)[idx].isBreakpoint) {
         EntryInfo * breakPointEntryInfo = CpvAccess(breakPointEntryTable)->get(idx);
         if (breakPointEntryInfo != NULL) {
+
```
evan-charmworks: Extraneous blank line?

src/ck-core/register.C
```diff
@@ -316,7 +316,8 @@ void _registerDone(void)
 {
   __registerDone = 1;
 #if CMK_CHARMDEBUG
-  if (CkMyRank() == 0) {
+  //  if (CkMyRank() == 0)
```
evan-charmworks: Should this comment be removed too?

src/conv-ccs/conv-ccs.C
```diff
@@ -545,6 +546,7 @@ CpvDeclare(int, cpdSuspendStartup);
 
 void CcsInit(char **argv)
 {
+
```
evan-charmworks: extraneous blank line

src/conv-ccs/conv-ccs.C
```diff
@@ -598,7 +599,12 @@ void CcsInit(char **argv)
      }
 
      if (CmiGetArgFlagDesc(argv, "+DebugSuspend", "Suspend execution at beginning of program")) {
+#if CMK_SMP       
+       if(CmiMyRank() != CmiMyNodeSize()) CpvAccess(cpdSuspendStartup) = 1;
```
evan-charmworks: This block could be:

```
#if CMK_SMP
      if (!CmiInCommThread())
#endif
        CpvAccess(cpdSuspendStartup) = 1;
```

At the least, testing `CmiMyRank()` in non-SMP mode is extraneous.

src/conv-ccs/middle-ccs.C
```diff
@@ -48,8 +50,23 @@ void req_fw_handler(char *msg)
...
+  
+#if CMK_SMP
+  // Charmrun gave us a message that is being handled on the rank 0 CCS PE
+  // i.e., PE 0, but needs to get forwarded to the real PE destination
+  if(destPE>0 && destPE!=CmiMyPe())
+    {
+      int len=CmiReservedHeaderSize+sizeof(CcsImplHeader)+ChMessageInt(hdr->len);
+      CmiSyncSend(destPE, len, msg);      
+    }
+  else{
```
evan-charmworks: It's probably cleaner to make this:
```
  else
#endif
  {    
    CcsHandleRequest(hdr, msg+offset);
  }
```

src/conv-core/convcore.C
```diff
@@ -3798,19 +3809,21 @@ void ConverseCommonInit(char **argv)
   CmiIOInit(argv);
 #endif
   if (CmiMyPe() == 0)
-      CmiPrintf("Converse/Charm++ Commit ID: %s\n", CmiCommitID);
+    CmiPrintf("[%d] Converse/Charm++ Commit ID: %s\n", CmiMyPe(), CmiCommitID);
```
evan-charmworks: Is it necessary to print the PE if we test `if (CmiMyPe() == 0)` first?

src/ck-core/debug-charm.C
ericjbohm: done in fixup

src/ck-core/debug-charm.C
ericjbohm: done in fixup

src/conv-ccs/conv-ccs.C
ericjbohm: done in fixup

src/conv-ccs/conv-ccs.C
ericjbohm: done in fixup

src/conv-core/convcore.C
ericjbohm: done in fixup

src/ck-core/register.C
ericjbohm: done in fixup

src/conv-ccs/middle-ccs.C
ericjbohm: done in fixup

src/ck-core/debug-charm.C
```diff
@@ -932,8 +953,14 @@ void CpdRemoveBreakPoint (char *msg)
         if (breakPointEntryInfo != NULL) {
           if (--breakPointEntryInfo->msgIdx == 0) {
             // If we are the last to delete the breakpoint, then restore the original name and call function pointer
-            _entryTable[idx]->name =  breakPointEntryInfo->name;
-            _entryTable[idx]->call = (CkCallFnPtr)breakPointEntryInfo->call;
+#if CMK_SMP	    
+	    if(_entryTable[idx]->breakPointSet--==0)
```
evan-charmworks: Is postfix decrement here and and on line 987 correct?

ericjbohm:
I can't reproduce that BigSim error.  Is there something weird about that build environment I should know about?

src/ck-core/debug-charm.C
ericjbohm: Good catch.  Didn't really test removing breakpoints until you pointed that out.  in !fixup

PR #2541, opened 2019-09-27
Charmrun: Refactor FT process restart to share code with normal launch

evan-charmworks:
This allows ++local and ++debug to apply to restarted processes.

Diff best viewed with --ignore-all-space --color-moved.


src/util/charmrun-src/charmrun.C
```diff
@@ -4445,27 +4446,38 @@ static void envCat(char *dest, LPTSTR oldEnv)
 
 /* simple version of charmrun that avoids the sshd or charmd,   */
 /* it spawn the node program just on local machine using exec. */
-static void start_nodes_local(std::vector<nodetab_process> & process_table)
+struct local_nodestart
 {
   char cmdLine[10000];     /*Program command line, including executable name*/
```
matthiasdiener: Could change this into a std::string.

src/util/charmrun-src/charmrun.C
```diff
@@ -5399,6 +5417,41 @@ static void start_nodes_mpiexec()
   /* all ssh_pid remain zero: skip finish_nodes */
 }
 
+static int finish_one_node(nodetab_process & p, int & retries, const char ** argv = arg_argv)
```
matthiasdiener: What's the meaning of the return value?

src/util/charmrun-src/charmrun.C
```diff
@@ -5496,148 +5526,174 @@ static char *find_abs_path(const char *target)
...
-  /* copy environ and expanded to hold NETSTART and CmiNumNodes */
   int envc;
-  for (envc = 0; env[envc]; envc++)
-    ;
-  int extra = 0;
+  char **envp;
+  int n;
+
+  char **dparamp;
+  std::vector<char *> dparamv;
```
matthiasdiener: Could be a std::string as well.

src/util/charmrun-src/charmrun.C
```diff
@@ -4445,27 +4446,38 @@ static void envCat(char *dest, LPTSTR oldEnv)
 
 /* simple version of charmrun that avoids the sshd or charmd,   */
 /* it spawn the node program just on local machine using exec. */
-static void start_nodes_local(std::vector<nodetab_process> & process_table)
+struct local_nodestart
```
matthiasdiener: maybe change to class

src/util/charmrun-src/charmrun.C
evan-charmworks: Done

src/util/charmrun-src/charmrun.C
evan-charmworks: It's an idiom where each non-success return site uses a different non-zero value. The return value is only used in a loop checking that it non-zero.

src/util/charmrun-src/charmrun.C
evan-charmworks: Not the way it is used.

src/util/charmrun-src/charmrun.C
evan-charmworks: If everything inside is public I prefer using struct.

evan-charmworks:
> Is this meant for 6.10 still?

It could be.

src/util/charmrun-src/charmrun.C
matthiasdiener: cmdLine doesn't even need to be public.

src/util/charmrun-src/charmrun.C
matthiasdiener: Yeah, this would require more refactoring.

evan-charmworks:
Since the minor fixes from this were merged in #2555, let's save the change that would require testing for 6.11.

evan-charmworks:
Needs re-review

PR #2546, opened 2019-10-01
Remove duplicate lock in HAPI stream creation

### LINKED ISSUE #2544: hapiCreateStreams() does a double lock
This is with the v6.10.0-rc tag of charm and the trq/gpu-memrace branch of changa compiled with mpi-linux-x86_64-cuda-smp:
hapiCreateStreams() does a
CmiLock(CsvAccess(gpu_manager).stream_lock_);
before calling GPUManager::createStreams().  The CmiLock() call in createStreams() therefore hangs.

minitu:
Fixes #2544


minitu:
Callers of functions in GPUManager scope (e.g. GPUManager::createStreams()) are expected acquire locks when needed, and thus the callee should not invoke locking functions.

Issue #2523, opened 2019-09-16
Spurious segfaults with BigSim + AMPI migration


matthiasdiener:
Happens sometimes in GitHub CI:

``` console
$ ./build LIBS netlrts-linux-x86_64 bigsim -j4
[...]

$ make -C migration bgtest OPTS='' TESTOPTS='++local ++no-va-randomization'
make[3]: Entering directory '/home/runner/work/charm/charm/netlrts-linux-x86_64-bigsim/tests/ampi/migration'
../../../bin/ampicxx  -c test.C
../../../bin/ampicxx  -o pgm test.o

[...]

../../../bin/testrun  ./pgm +p2 +vp2 +x2 +y1 +z1  ++local ++no-va-randomization
Charmrun> scalable start enabled. 
Charmrun> started all node programs in 0.007 seconds.
Charm++> Running in non-SMP mode: 2 processes (PEs)
Converse/Charm++ Commit ID: v6.10.0-beta1-24-gfb5620bcc
Warning> Randomization of virtual memory (ASLR) is turned on in the kernel, thread migration may not work! Run 'echo 0 > /proc/sys/kernel/randomize_va_space' as root to disable it, or try running with '+isomalloc_sync'.
Charm++> scheduler running in netpoll mode.
BG info> Simulating 2x1x1 nodes with 1 comm + 1 work threads each.
BG info> Network type: bluegene.
alpha: 1.000000e-07    packetsize: 1024    CYCLE_TIME_FACTOR:1.000000e-03.
CYCLES_PER_HOP: 5  CYCLES_PER_CORNER: 75.
BG info> cpufactor is 1.000000.
BG info> floating point factor is 0.000000.
BG info> Using WallTimer for timing method. 
CharmLB> Load balancer ignores processor background load.
CharmLB> Load balancer assumes all CPUs are same.
Trace: traceroot: ./pgm

begin migrating

begin migrating
Trying to migrate partition 1 from pe 1 to 0
Entering TCHARM_Migrate_to, FEM_My_partition is 1, CkMyPe() is 1, migrate_test is 1
Leaving TCHARM_Migrate_to, FEM_My_partition is 1, CkMyPe() is 0, migrate_test is 1
Done with step 0
Done with step 0
Trying to migrate partition 1 from pe 0 to 1
Entering TCHARM_Migrate_to, FEM_My_partition is 1, CkMyPe() is 0, migrate_test is 0
[...]

PR #2547, opened 2019-10-01
Fixed uninitialized pointer in LBObjUserData copy constructor

juanjgalvez:
If d.data is NULL, the 'data' pointer of the new instance remains
unintialized, which can cause segfaults if the instance is deleted
afterwards.

### DIFF
--- a/src/ck-ldb/lbdb.h
+++ b/src/ck-ldb/lbdb.h
@@ -141,12 +141,16 @@ public:
     if (d.data != NULL) {
       if (data==NULL) init();
       memcpy(data, d.data, CkpvAccess(lbobjdatalayout).size());
+    } else if (data != NULL) {
+      delete [] data;
+      data = NULL;
     }
     return *this;
   }
-  inline void init() { data = new char[CkpvAccess(lbobjdatalayout).size()]; }
   inline void pup(PUP::er &p);
   void *getData(int idx) { if (data==NULL) init(); return (void*)(data+idx); }
+private:
+  inline void init() { data = new char[CkpvAccess(lbobjdatalayout).size()]; }
 };
 
 typedef struct {
--- a/src/ck-ldb/lbdb.h
+++ b/src/ck-ldb/lbdb.h
@@ -131,6 +131,8 @@ public:
     if (d.data != NULL) {
       init();
       memcpy(data, d.data, CkpvAccess(lbobjdatalayout).size());
+    } else {
+      data = NULL;
     }
   }
 


src/ck-ldb/lbdb.h
```diff
@@ -131,6 +131,8 @@ class LBObjUserData {
     if (d.data != NULL) {
       init();
       memcpy(data, d.data, CkpvAccess(lbobjdatalayout).size());
+    } else {
+      data = NULL;
```
matthiasdiener: Should this also be done for the copy constructor below?

juanjgalvez: You mean the assignment operator? It's not needed there because the operator should imply that `this` has already been initialized.

matthiasdiener: I mean, if the (copied) `d.data == NULL`, `this->data` isn't updated. Is this intentional (ie, shouldn't it also be NULL)?

rbuch: I agree with @matthiasdiener, it seems like if `d.data` is `NULL`, then that should be propagated by `operator=`.

juanjgalvez: I'm not sure what should happen if `d.data` is `NULL`. I'm not familiar with this class. I don't even know if the = operator is used in practice or under what circumstances. A LBObjUserData which has been init'ed (via init call) dynamically allocates an array with garbage (for user to populate later). If you are assigning a LBObjUserData with data == NULL but this.data != NULL, presumably you would want to destroy the array? I'm not sure if this situation would even happen in practice. There could also be a danger of exposing other bugs if you destroy the array without destroying the object. The user might already have a pointer to it.

epmikida: If assigning when data == NULL is not valid, then it should at least error I'd imagine. Either this same check should appear in the operator= or it should more specifically deal with the possibility, not ignore it. If there is a copy ctor, there should be an equivalent operator=, and if they don't behave the same, there should be a good reason for it.

rbuch: As a user, I'd expect the semantics of `operator=` to produce an lhs where every field matches the rhs. I'd expect the array to be destroyed if you are assigning an object with `NULL` `data`, I think.

juanjgalvez: Or maybe the semantics of `operator=` in LBObjUserData when d.data is NULL are that it is an error and it should abort. But I'll make the change anyway since it seems most people agree on this.

epmikida: It sounds like the usage of this class is not well documented, if its not clear what is allowed, so maybe that's something that should be looked at at some point.

src/ck-ldb/lbdb.h
```diff
@@ -139,12 +141,16 @@ class LBObjUserData {
...
+      delete [] data;
+      data = NULL;
     }
     return *this;
   }
-  inline void init() { data = new char[CkpvAccess(lbobjdatalayout).size()]; }
   inline void pup(PUP::er &p);
   void *getData(int idx) { if (data==NULL) init(); return (void*)(data+idx); }
+private:
+  inline void init() { data = new char[CkpvAccess(lbobjdatalayout).size()]; }
```
rbuch: I think this needs to be public, otherwise there's seemingly no way for a user to ever allocate space for data because the only non-copy constructor only sets `data` to NULL.

juanjgalvez:
NetLRTS Linux BigSim failed, but it didn't fail on the previous commit. Not sure if it is related to this PR or not. Does someone know if NetLRTS Linux BigSim is behaving erratically recently, or should it always pass?

matthiasdiener:
> NetLRTS Linux BigSim failed, but it didn't fail on the previous commit. Not sure if it is related to this PR or not. Does someone know if NetLRTS Linux BigSim is behaving erratically recently, or should it always pass?

The BigSim failure is a known issue, we can ignore it:  https://github.com/UIUC-PPL/charm/issues/2523

src/ck-ldb/lbdb.h
juanjgalvez: init() is called from getData()

rbuch: Ah, I missed that, looks fine.

PR #2551, opened 2019-10-04
CkLocMgr: buffer immigrate messages if array managers don't exist

juanjgalvez:
This replaces the buggy busy-wait that reschedules the messages
until the managers are created (which doesn't work because
immigrate messages are expedited).

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2194,9 +2194,19 @@ void CkLocMgr::pup(PUP::er &p){
 /// Add a new local array manager to our list.
 void CkLocMgr::addManager(CkArrayID id,CkArray *mgr)
 {
-	CK_MAGICNUMBER_CHECK
-	DEBC((AA "Adding new array manager\n" AB));
-	managers[id] = mgr;
+  CK_MAGICNUMBER_CHECK
+  DEBC((AA "Adding new array manager\n" AB));
+  managers[id] = mgr;
+  auto i = pendingImmigrate.begin();
+  while (i != pendingImmigrate.end()) {
+    auto msg = *i;
+    if (msg->nManagers <= managers.size()) {
+      i = pendingImmigrate.erase(i);
+      immigrate(msg);
+    } else {
+      i++;
+    }
+  }
 }
 
 void CkLocMgr::deleteManager(CkArrayID id, CkArray *mgr) {
@@ -3015,9 +3025,9 @@ void CkLocMgr::immigrate(CkArrayElementMigrateMessage *msg)
 	if (msg->nManagers < managers.size())
 		CkAbort("Array element arrived from location with fewer managers!\n");
 	if (msg->nManagers > managers.size()) {
-		//Some array managers haven't registered yet-- throw it back
-		DEBM((AA "Busy-waiting for array registration on migrating %s\n" AB,idx2str(idx)));
-		thisProxy[CkMyPe()].immigrate(msg);
+		//Some array managers haven't registered yet -- buffer the message
+		DEBM((AA "Buffering %s immigrate msg waiting for array registration\n" AB,idx2str(idx)));
+		pendingImmigrate.push_back(msg);
 		return;
 	}
 
--- a/src/ck-core/cklocation.h
+++ b/src/ck-core/cklocation.h
@@ -538,6 +538,8 @@ private:
 	CkGroupID lbdbID;
 	CkGroupID metalbID;
 
+	std::list<CkArrayElementMigrateMessage*> pendingImmigrate;
+
 	ck::ArrayIndexCompressor *compressor;
 	CkArrayIndex bounds;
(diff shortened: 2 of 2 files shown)


juanjgalvez:
A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).

Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.

juanjgalvez:
@nitbhat Can you test the patch on Cori and see if it fixes megatest?

epmikida:
> A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).
> 
> Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.

I'm not entirely sure how this should work. None of the other buffers are pupped, and I'm not sure if doing a checkpoint/restart is valid when there are chares in migration AND there is an array manager that has not been created on some PEs. If you checkpoint when only some of the CkArray objects have been created I think you will have an invalid checkpoint. So I don't think you can make a valid checkpoint without having flushed the immigration message buffer first anyways.

juanjgalvez:
> A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).
> Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.
> 
> I'm not entirely sure how this should work. None of the other buffers are pupped, and I'm not sure if doing a checkpoint/restart is valid when there are chares in migration AND there is an array manager that has not been created on some PEs. If you checkpoint when only some of the CkArray objects have been created I think you will have an invalid checkpoint. So I don't think you can make a valid checkpoint without having flushed the immigration message buffer first anyways.

Should we just put an abort inside pup if pendingImmigrate.size() > 0?

epmikida:
> > A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).
> > Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.
> > I'm not entirely sure how this should work. None of the other buffers are pupped, and I'm not sure if doing a checkpoint/restart is valid when there are chares in migration AND there is an array manager that has not been created on some PEs. If you checkpoint when only some of the CkArray objects have been created I think you will have an invalid checkpoint. So I don't think you can make a valid checkpoint without having flushed the immigration message buffer first anyways.
> 
> Should we just put an abort inside pup if pendingImmigrate.size() > 0?

Yeah an abort or CkAssert would probably be a good idea.

juanjgalvez:
> A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).
> Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.
> I'm not entirely sure how this should work. None of the other buffers are pupped, and I'm not sure if doing a checkpoint/restart is valid when there are chares in migration AND there is an array manager that has not been created on some PEs. If you checkpoint when only some of the CkArray objects have been created I think you will have an invalid checkpoint. So I don't think you can make a valid checkpoint without having flushed the immigration message buffer first anyways.
> 
> Should we just put an abort inside pup if pendingImmigrate.size() > 0?
> 
> Yeah an abort or CkAssert would probably be a good idea.

Ok. Since CkLocMgr::pup is not in any critical path I am inclined to use an abort so that it can be detected without error checking too.

epmikida:
> > A remaining issue is that presumably CkLocMgr::pup needs to serialize the pending immigrate messages if any (for things like checkpointing I guess).
> > Packing the messages with pup (p|*msg) builds, but I'm not sure how to do the unpacking since they are apparently variable size. The other matter would be testing the pup implementation.
> > I'm not entirely sure how this should work. None of the other buffers are pupped, and I'm not sure if doing a checkpoint/restart is valid when there are chares in migration AND there is an array manager that has not been created on some PEs. If you checkpoint when only some of the CkArray objects have been created I think you will have an invalid checkpoint. So I don't think you can make a valid checkpoint without having flushed the immigration message buffer first anyways.
> > Should we just put an abort inside pup if pendingImmigrate.size() > 0?
> > Yeah an abort or CkAssert would probably be a good idea.
> 
> Ok. Since CkLocMgr::pup is not in any critical path I am inclined to use an abort so that it can be detected without error checking too.

Sounds good to me. Probably mention in the abort message that it is likely caused by attempting to checkpoint when array creation has not fully completed. That way if people run into it they have a clue as to why something broke.

PR #2552, opened 2019-10-04
CkLocMgr::pup abort when packing if there are buffered immigrate msgs

juanjgalvez:


nitbhat:
Is checkpointing the only time when `CkLocMgr::pup` is invoked?

src/ck-core/cklocation.C
```diff
@@ -2122,6 +2122,9 @@ CkLocMgr::~CkLocMgr() {
 }
 
 void CkLocMgr::pup(PUP::er &p){
+  if (p.isPacking() && pendingImmigrate.size() > 0)
+    CkAbort("Attempting to pup location manager with buffered migration messages."
+            " Likely cause is checkpointing before array creation has fully completed\n");
 	IrrGroup::pup(p);
```
matthiasdiener: Can you fix the indentation of the lines below the 2 lines you added? Otherwise, indentation might look weird depending on the tab size that the editor uses.

epmikida:
> Is checkpointing the only time when `CkLocMgr::pup` is invoked?

I think its also invoked for other similar cases like double in-memory checkpoint, and maybe shrink expand. But the same principle holds in that if packing up before an array manager exists, it is invalid and should result in an abort.

src/ck-core/cklocation.C
juanjgalvez: I agree, but I really think a separate patch should clean up the indentation of that file, or at least the pup method.

matthiasdiener: A short-term option could be to just keep the tab-based indentation for the added lines.

juanjgalvez: I considered that, but figured that the CkAbort indentation would look inconsistent with tabs

epmikida: FWIW, I already have a commit that reformats much of CkLocMgr, including pup. It's part of a larger overhaul of CkLocMgr that I'm not pushing anywhere until after 6.10 is done and I've done a bit more cleanup on it.

PR #2554, opened 2019-10-04
cleanup: remove vestiges of CMK_AIX / CMK_USE_GM / CMK_USE_MX

matthiasdiener:
Followup from the removals of AIX support (39643aa) and Myrinet (102a60c).

### DIFF
--- a/examples/ampi/Cjacobi3D/jacobi.C
+++ b/examples/ampi/Cjacobi3D/jacobi.C
@@ -139,11 +139,7 @@ int main(int ac, char** av)
 
   MPI_Bcast(&niter, 1, MPI_INT, 0, MPI_COMM_WORLD);
 
-#if CMK_AIX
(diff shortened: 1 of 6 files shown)


matthiasdiener:
> Given that its been so long since a Myrinet machine was on the TOP500 that dropped myrinet from the pull down list, I think we can drop Myrinet.
> 
> However, AIX still exists. IBM released a new version last year. Regardless, the two things should not be compounded in one PR. Also, each of their removals should be preceded by a deprecation warning in CHANGES.

As far as I can tell, AIX support was removed in 2016 (39643aa4e780e1ffdfc4e60cd511496c6a41ce01), and Myrinet in 2017 (102a60c8f052af25974d073bfefda946395cccce). This patch is only meant to remove the now (apparently) unused CMK_* #ifdefs that were missed in the previous removals.

PR #2557, opened 2019-10-07
cmake: add conv-static dependency to romio

### LINKED ISSUE #2556: CMake build may fail when building with -j
When building with CMake and `-j4`, I get a failure in ROMIO:

```
configure: error: in `<charm location>/charm/netlrts-linux-x86_64/src/libs/ck-libs/ampi/romio-prefix/src/romio':
configure: error: C compiler cannot create executables
See `config.log' for more details
src/libs/ck-libs/ampi/CMakeFiles/romio.dir/build.make:108: recipe for target 'src/libs/ck-libs/ampi/romio-prefix/src/romio-stamp/romio-configure' failed
make[2]: *** [src/libs/ck-libs/ampi/romio-prefix/src/romio-stamp/romio-configure] Error 77
CMakeFiles/Makefile2:3071: recipe for target 'src/libs/ck-libs/ampi/CMakeFiles/romio.dir/all' failed
make[1]: *** [src/libs/ck-libs/ampi/CMakeFiles/romio.dir/all] Error 2
```

Looking at `config.log`indicates the error is caused by `g++: error: <charm location>/charm/netlrts-linux-x86_64/bin/../lib/conv-static.o: No such file or directory`. However, `conv-static.o` does exist and is in the correct location. This, coupled with the fact that the ROMIO output is interleaved with lines about building some files in `conv-core` makes me think this is due to a race from an unspecified dependency or something like that. (Note that I haven't been to reproduce this when building serially and if I go back and just run `make` in the build folder after encountering this failure, that works fine.)

matthiasdiener:
Fixes #2556.

### DIFF
--- a/src/libs/ck-libs/ampi/CMakeLists.txt
+++ b/src/libs/ck-libs/ampi/CMakeLists.txt
@@ -87,7 +87,7 @@ if(${CMK_AMPI_WITH_ROMIO})
 
     # Since we link a program with charmc as part of ROMIO's configure,
     # we need all libraries built at this point.
-    add_dependencies(romio moduleampi ampi-compat ck modulecompletion conv-partition conv-util hwloc conv-static)
+    add_dependencies(romio moduleampi ampi-compat ck modulecompletion conv-partition conv-util hwloc conv-static conv-core)
     if(${CMK_CAN_LINK_FORTRAN})
         add_dependencies(romio moduleampif)
     endif()
--- a/src/libs/ck-libs/ampi/CMakeLists.txt
+++ b/src/libs/ck-libs/ampi/CMakeLists.txt
@@ -87,7 +87,7 @@ if(${CMK_AMPI_WITH_ROMIO})
 
     # Since we link a program with charmc as part of ROMIO's configure,
     # we need all libraries built at this point.
-    add_dependencies(romio moduleampi ampi-compat ck modulecompletion conv-partition conv-util hwloc)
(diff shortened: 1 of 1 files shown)


rbuch:
This doesn't fix the `configure` error I am seeing, I still get the same error as before.

matthiasdiener:
> This doesn't fix the `configure` error I am seeing, I still get the same error as before.

Is the message in config.log still the same (ie, missing cons-static.o)?

rbuch:
> > This doesn't fix the `configure` error I am seeing, I still get the same error as before.
> 
> Is the message in ROMIO's config.log still the same (ie, missing conv-static.o)?

Yes, it's still the same error.

It works fine if I change the dependency to be on `conv-core` instead of `conv-static`, but I'm not quite sure why.

matthiasdiener:
> > > This doesn't fix the `configure` error I am seeing, I still get the same error as before.
> > 
> > 
> > Is the message in ROMIO's config.log still the same (ie, missing conv-static.o)?
> 
> Yes, it's still the same error.
> 
> It works fine if I change the dependency to be on `conv-core` instead of `conv-static`, but I'm not quite sure why.

Hmm, that's quite strange. I added conv-core as an additional dependency to fix the immediate issue, but we will need to create more comprehensive dependency handling for 6.11.

evan-charmworks:
The need for `conv-core` might be that it contains the code that depends on `conv-static`.

matthiasdiener:
Needs another review.

Issue #1895, opened 2018-05-03
Charm's hwloc can collide with OpenMPI's


stwhite91:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1895*

---
https://lists.cs.illinois.edu/lists/arc/charm/2018-05/msg00004.html

evan-charmworks:
*Original date: 2018-05-08 16:21:00*

---
~~https://charm.cs.illinois.edu/gerrit/4145~~ https://github.com/UIUC-PPL/charm/commit/4348423d70ba1078bf77f429410017e3da87ff52

I am also seeking to merge this fix upstream: https://github.com/open-mpi/hwloc/pull/311

PR #2558, opened 2019-10-10
Update hwloc to v2.1.0

evan-charmworks:


matthiasdiener:
Since this changes hwloc's API, there are a few things we should consider:
- Are Charm++/AMPI applications allowed to use hwloc directly if they want to? 
- Do we support mixing our hwloc with an external hwloc?

evan-charmworks:
> * Are Charm++/AMPI applications allowed to use hwloc directly if they want to?

It looks like we expose hwloc.h in our include folder. I would discourage this though. 

Note that I didn't need to change our code that makes use of hwloc for this update.

> * Do we support mixing our hwloc with an external hwloc?

Yes, see #1895.

matthiasdiener:
> > * Are Charm++/AMPI applications allowed to use hwloc directly if they want to?
> 
> It looks like we expose hwloc.h in our include folder. 

So in case the application requires another version of hwloc.h, there might be conflicts?

> Note that I didn't need to change our code that makes use of hwloc for this update.

Ok, but there have been substantial API changes in 2.0: https://www.open-mpi.org/projects/hwloc/doc/v2.1.0/a00343.php

stwhite91:
As long as the API changes don't affect our current usage, I think the upgrade should be merged on its own, and any other issues cleaned up in separate commits

PR #2560, opened 2019-10-16
remove ccache from travis ci Mac build

matthiasdiener:
Currently, `brew install ccache` errors, failing the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ before_install:
           exit
         }
       fi
-install:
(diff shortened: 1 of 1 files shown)


rbuch:
The Travis documentation still says to use the given lines for macOS, but if removing them makes it work empirically, then it's fine by me. Also, the utility of `ccache` for macOS here is dubious because it takes so long to install it in the first place, so going without it isn't much of an issue.

PR #2562, opened 2019-10-18
UCX: Several fixes

brminich:
- Do not post RX buffers for direct protocol recursively (can lead to huge stack consumption) 
- Fix probe (rendezvous) protocol to work correctly for ZCOPY messages (do not use `req->msgBuf` in case of immediate completion, because it is not initialized)


stwhite91:
Looks good to me, Nitin should review it though

src/arch/ucx/machine.C
```diff
@@ -372,99 +438,37 @@ static void UcxRxReqCompleted(void *request, ucs_status_t status,
...
-            CmiInvokeNcpyAck(ncpyOpInfo);
-
-        } else {
-            CmiAbort(" Cannot de-register on a different node than the source or destinaton");
-        }
-    }
-#endif
-
     if (req->msgBuf != NULL) {
         // Request is not completed immideately
```
nitbhat: Minor Spelling Fix (not related to this patch), but could go in this patch: immideately => immediately (has multiple occurrences in this file).

src/arch/ucx/machine.C
```diff
@@ -32,7 +32,7 @@
 #define CmiSetMsgSize(msg, sz)    ((((CmiMsgHeaderBasic *)msg)->size) = (sz))
 
 #define UCX_MSG_PROBE_THRESH            32768
-#define UCX_MSG_NUM_RX_REQS             64
+#define UCX_MSG_NUM_RX_REQS             16
```
nitbhat: Any specific reason for choosing 16 for the number of pre-posted buffers?

src/arch/ucx/machine.C
```diff
@@ -634,9 +641,11 @@ void LrtsExit(int exitcode)
 
     for (i = 0; i < ucxCtx.numRxReqs; ++i) {
         req = ucxCtx.rxReqs[i];
-        CmiFree(req->msgBuf);
-        ucp_request_cancel(ucxCtx.worker, req);
-        ucp_request_free(req);
+        if (!req->completed) {
```
nitbhat: Why aren't we freeing `req->msgBuf` if the request is completed?

src/arch/ucx/machine.C
```diff
@@ -372,99 +438,37 @@ static void UcxRxReqCompleted(void *request, ucs_status_t status,
...
-            ncpyOpInfo->opMode = CMK_EM_API_DEST_ACK_INVOKE;
-            CmiInvokeNcpyAck(ncpyOpInfo);
-
-        } else {
-            CmiAbort(" Cannot de-register on a different node than the source or destinaton");
-        }
-    }
-#endif
-
     if (req->msgBuf != NULL) {
```
nitbhat: How is this a condition for the request not completing immediately?

brminich: request is returned by `ucp_tag_recv_nb`. If receive is completed immediately, completion callback is invoked inside `ucp_tag_recv_nb` and `req->msgBuf` is not initialized yet. When it is not initialized it should be `NULL`, because we set it to `NULL` every time when return it to memory pool ( before invoke `ucp_request_free`).  In the callback we set `req->completed` to `1`. So, as soon as `ucp_tag_recv_nb` returns, we check `req->completed` and initialize `req->msgBuf` if `req->completed` is `0`. Thus, when completion callback is invoked later, `req->msgBuf`  is not `NULL`

src/arch/ucx/machine.C
brminich: I made some experiments with p2p benchmarks and looks like it gives slightly better performance

src/arch/ucx/machine.C
brminich: it should be done in `handleOneRecvedMsg`, right?

nitbhat: Ah, I see. Yes, it does happen at a higher level. So, this is just taking care of the incomplete ones. Got it.

src/arch/ucx/machine.C
nitbhat: I was wondering if having more buffers is useful for an "all-to-all" sort of situation, but since we've provided a switch (`+ucx_num_rx_reqs`) for controlling this, I think we should be good.

src/arch/ucx/machine.C
nitbhat: Okay, thanks for the clarification, that makes sense. 

So, since we need to set `req->msgBuf` to `NULL` before calling `UcxPostRxReq`, shouldn't we have to call `UCX_REQUEST_FREE` when we first pre-post the buffers in `LrtsInit`? (Inside the for loop on machine.C:417)

src/arch/ucx/machine.C
```diff
@@ -135,10 +136,125 @@ void UcxRequestInit(void *request)
...
+                                               ucp_dt_make_contig(1), msg,
+                                               UcxRxReqCompleted);
+    }
+
+    CmiEnforce(!UCS_PTR_IS_ERR(req));
+    UCX_LOG(3, "Posted RX buf %p size %zu, req %p, tag %zu, comp %d\n",
+            req->msgBuf, size, req, tag, req->completed);
+
+    // Request completed immediately
+    if (req->completed) {
```
nitbhat: I was thinking that it'll aid readability if we have the same condition for immediate completion. 

Isn't `req->completed` equivalent to `reg->msgBuf == NULL`?

src/arch/ucx/machine.C
```diff
@@ -282,9 +398,25 @@ void LrtsInit(int *argc, char ***argv, int *numNodes, int *myNodeID)
...
+        }
+    }
+
     UcxInitEps(*numNodes, *myNodeID);
 
-    UcxPrepostRxBuffers();
+    ucxCtx.rxReqs = (UcxRequest**)CmiAlloc(sizeof(UcxRequest*) * ucxCtx.numRxReqs);
+    CmiEnforce(ucxCtx.rxReqs);
+
+    // Post RX requests for direct protocol
```
nitbhat: Since have a Direct API, the usage of 'direct' here is a little confusing. 

Can you change the comment to 'Post RX requests for large messages (which are over the eager threshold and use RDMA)'

src/arch/ucx/machine.C
```diff
@@ -610,6 +614,9 @@ void LrtsAdvanceCommunication(int whileidle)
 #if CMK_SMP
        cnt += ProcessTxQueue();
 #endif
+       if (ucxCtx.numFreeRxReqs >= ucxCtx.rxReqsBatch) {
+           UcxPostRxBuffers();
+       }
```
nitbhat: What's the idea behind posting in batches (default being 16/4 = 4)? 

Was it only done to avoid the recursive scenario from before or is there any other advantage?

src/arch/ucx/machine.C
brminich: Yes, it is needed. It is done in `UcxRequestInit`, which is one time initializer for all UCX requests

src/arch/ucx/machine.C
brminich: Not exactly, because when `ucp_tag_recv_nb` returns a request we have to check whether it is already completed or not. We check `req->completed`, since it is filled in the completion callback. An alternative to using `req->completed` field, would be using `ucp_request_check_status` API.  
```
if (ucp_request_check_status(req) != UCS_INPROGRESS) {
}
```
In this case we will not need `req->completed` field.

src/arch/ucx/machine.C
brminich: ok

src/arch/ucx/machine.C
brminich: the main idea is to avoid recursion. Batches are used to reduce overhead, so we will repost a bunch of requests rather than posting them separately every time some request is completed

src/arch/ucx/machine.C
brminich: ok

src/arch/ucx/machine.C
nitbhat: Oh okay, understood. So, `UcxRequestInit` being passed as `cParams.request_init` makes `ucp_init` to call `UcxRequestInit` on each of the contexts? It's a little strange since the allocation of `UcxRequest`s happens in line 412. So, when exactly does `UcxRequestInit` get called on the contexts (for the first time)?

brminich: It just initializes every request before its first use. According to ucx docs (from ucp.h file)
```
 700     /**
 701      * Pointer to a routine that is used for the request initialization.
 702      * This function will be called only on the very first time a request memory
 703      * is initialized, and may not be called again if a request is reused.
 704      * If a request should be reset before the next reuse, it can be done before
 705      * calling @ref ucp_request_free.

nitbhat: Okay, makes sense. So, the UCX runtime takes care of it.

src/arch/ucx/machine.C
nitbhat: Okay, understood.

nitbhat:
All except two minor changes are pending: 
1. Spelling change 
2. Comment change

Do you want me to work on those? @brminich

brminich:
@nitbhat, I'll update this PR. Was waiting for conclusions

PR #2563, opened 2019-10-18
Implement isomalloc_sync using Converse messaging

### LINKED ISSUE #2252: Implement isomalloc_sync using Converse messaging rather than shared file system
AMPI's +isomalloc_sync option is currently implemented using a shared file system.
This can become a bottleneck during startup on large systems with ASLR, such as Stampede2 at TACC. A better implementation would use messages instead of the file system to exchange the starting addresses among all processes.

evan-charmworks:
Fixes #2252


evan-charmworks:
Confirmed working on GNI and Verbs

src/conv-core/isomalloc.C
```diff
@@ -664,233 +727,126 @@ static void CmiIsomallocInitExtent()
...
         CmiAbort("Isomalloc> call to read() failed during restart!");
-      }
+
       close(fd);
-      if (ss < s || ee > e)
-        CmiAbort(
-            "isomalloc_sync failed during restart, virtual memory regions do not "
-            "overlap.");
+      if (previous.s < IsoRegion.s || previous.e > IsoRegion.e)
+        CmiError("Isomalloc> Warning: virtual memory regions do not overlap: "
```
stwhite91: It's weird to have an abort that calls itself a "Warning"

matthiasdiener: It's not an abort, just printing an error message

evan-charmworks:
> It looks like this makes +isomalloc_sync essentially the default, which we want, but is there a way to turn it off, in case there is an issue with it?

Added `+no_isomalloc_sync`

src/conv-core/isomalloc.C
```diff
@@ -664,233 +728,132 @@ static void CmiIsomallocInitExtent()
...
-          CmiAbort(
-              "Isomalloc> call to system(\"sync\") failed while synchronizing memory "
-              "regions!");
-        }
-#endif
+        CsdSchedulePoll();
 
-        CmiBarrier();
+        if (IsoRegion.s >= IsoRegion.e)
+          CmiAbort("Isomalloc> failed to find consolidated region: %" PRIx64 " - %" PRIx64 ".\n", IsoRegion.s, IsoRegion.e);
```
stwhite91: Maybe add here that users can run with +no_isomalloc_sync if they don't need it

src/arch/mpi/machine.C
```diff
@@ -1787,17 +1786,16 @@ void LrtsPreCommonInit(int everReturn) {
     CpvAccess(crashedRankHdr) = NULL;
     CpvAccess(crashedRankPtr) = NULL;
 #endif
-}
-
-void LrtsPostCommonInit(int everReturn) {
```
stwhite91: Any insight on why these were here before, or what functionally the difference is between things that can go inside LrtsInit and thing that should go in LrtsPostCommonInit?

src/conv-core/isomalloc.C
evan-charmworks: Done

src/arch/mpi/machine.C
evan-charmworks: They were added in commit ff22b81b1c2058cd467e75071ac02fb7ebe6a19c, when these variables were turned into `Cpv`s. I don't see any reason to keep them there, since `LrtsPreCommonInit` has other `Cpv` initializations. If they are not moved, attempting to use Converse communication during `ConverseCommonInit` crashes due to null pointer deference on them.

evan-charmworks:
> This looks good, but currently is not scalable in that all nodes send messages directly to node 0 at the same time. This will flood node 0 with messages. That might be okay since the messages are small and it's a one-time operation, but it might fail at scale, particularly in non-SMP mode. A tree-based gather operation would be preferable. I'm not sure whether we want to hold this PR based on this point or not though.

I looked into using `CmiNodeReduce` but it is unimplemented.

```cpp
void CmiNodeReduce(void *data, int size, CmiReduceMergeFn mergeFn, int redID, int numChildren, int parent) {
  CmiAbort("Feel free to implement CmiNodeReduce...");
  /*
  CmiAssert(CmiRankOf(CmiMyPe()) == 0);
  CpvAccess(_reduce_data) = data;
  CpvAccess(_reduce_data_size) = size;
  CpvAccess(_reduce_parent) = CmiNodeFirst(CmiNodeSpanTreeParent(CmiMyNode()));
  _reduce_destination = (CmiHandler)CmiGetHandlerFunction(data);
  _reduce_pupFn = NULL;
  _reduce_mergeFn = mergeFn;
  CpvAccess(_reduce_num_children) = CmiNumNodeSpanTreeChildren(CmiMyNode());
  if (CpvAccess(_reduce_received) == CpvAccess(_reduce_num_children)) CmiSendReduce(size);
  */
}
#if 0
void CmiNodeReduce(void *data, int size, void * (*mergeFn)(void*,void**,int), int redID) {
  CmiNodeReduce(data, size, mergeFn, redID, CmiNumNodeSpanTreeChildren(CmiMyNode()),
      CmiNodeFirst(CmiNodeSpanTreeParent(CmiMyNode())));
}
void CmiNodeReduce(void *data, int size, void * (*mergeFn)(void*,void**,int), int numChildren, int parent) {
  CmiNodeReduce(data, size, mergeFn, CmiReduceNextID(), numChildren, parent);
}
void CmiNodeReduce(void *data, int size, void * (*mergeFn)(void*,void**,int)) {
  CmiNodeReduce(data, size, mergeFn, CmiReduceNextID(), CmiNumNodeSpanTreeChildren(CmiMyNode()),
      CmiNodeFirst(CmiNodeSpanTreeParent(CmiMyNode())));
}
#endif

void CmiNodeReduceStruct(void *data, CmiReducePupFn pupFn,
                         CmiReduceMergeFn mergeFn, CmiHandler dest,
                         CmiReduceDeleteFn deleteFn) {
  CmiAbort("Feel free to implement CmiNodeReduceStruct...");
/*
  CmiAssert(CmiRankOf(CmiMyPe()) == 0);
  CpvAccess(_reduce_data) = data;
  CpvAccess(_reduce_parent) = CmiNodeFirst(CmiNodeSpanTreeParent(CmiMyNode()));
  _reduce_destination = dest;
  _reduce_pupFn = pupFn;
  _reduce_mergeFn = mergeFn;
  _reduce_deleteFn = deleteFn;
  CpvAccess(_reduce_num_children) = CmiNumNodeSpanTreeChildren(CmiMyNode());
  if (CpvAccess(_reduce_received) == CpvAccess(_reduce_num_children)) CmiSendReduce(0);
  */
}
```

@epmikida Would `CmiNodeReduce` be difficult to implement?

stwhite91:
Let's discuss this in Core this week

evan-charmworks:
I rebased this PR over the latest changes in master and narrowed its scope to only replace the filesystem approach to synchronization with Converse messaging. It no longer enables `+isomalloc_sync` by default or removes the command line option. Reintroduction of these changes and replacement of the point-to-point messages to node 0 with `CmiNodeReduce` can happen in the future. This PR is still better than the filesystem approach despite not using a Converse reduction.

src/conv-core/memory-isomalloc.C
```diff
@@ -61,12 +61,10 @@ static void meta_init(char **argv)
    isomalloc_thread = 1;         /* isomalloc is allowed in this pthread */
 #endif
    if (CmiMyRank()==0) meta_inited = 1;
-#if CMK_SMP
-    if (CmiMyPe()==0 && _sync_iso == 0 && _sync_iso_warned == 0) {
+    if (CmiNumNodes() > 0 && CmiMyPe()==0 && _sync_iso == 0 && _sync_iso_warned == 0) {
         _sync_iso_warned = 1;
-        printf("Warning> Using Isomalloc in SMP mode, you may need to run with '+isomalloc_sync'.\n");
+        printf("Warning> Using Isomalloc with multiple logical nodes. You may need to run with '+isomalloc_sync'.\n");
```
stwhite91: Should this be prefaced with a check on CmiNumNodes > 1? Since we don't need to warn about this in a multicore build

stwhite91: Nevermind, I can't read

PR #2565, opened 2019-10-21
AMPI: Remove support for AMPI_Register_main and multiple worlds

evan-charmworks:


src/libs/ck-libs/ampi/ampi.C
```diff
@@ -1558,7 +1558,7 @@ TCharm *ampiParent::registerAmpi(ampi *ptr,ampiCommStruct s,bool forMigration) n
     } else if (isIntra(comm)) {
       intraChildRegister(s);
     }else
-      CkAbort("ampiParent received child with bad communicator: %d", comm);
```
matthiasdiener: If you prefer, after merging #2566, this fixup shouldn't be necessary anymore.

PR #2566, opened 2019-10-21
CkAbort: allow format strings in CkArray, CkMigratable

matthiasdiener:
also clarifies manual

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -1554,8 +1554,13 @@ CkMigratable::~CkMigratable() {
 }
 
 void CkMigratable::CkAbort(const char *format, ...) const {
-	CkError("CkMigratable '%s' aborting: ", _chareTable[thisChareType]->name);
-	::CkAbort(format);
+	char newmsg[256];
+	va_list args;
+	va_start(args, format);
+	vsnprintf(newmsg, sizeof(newmsg), format, args);
+	va_end(args);
+
+	::CkAbort("CkMigratable '%s' aborting: %s", _chareTable[thisChareType]->name, newmsg);
 }
 
 void CkMigratable::ResumeFromSync(void)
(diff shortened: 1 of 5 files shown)


src/ck-core/cklocation.C
```diff
@@ -1552,8 +1553,13 @@ CkMigratable::~CkMigratable() {
...
 
-void CkMigratable::CkAbort(const char *why) const {
-	::CkAbort("CkMigratable '%s' aborting: %s",_chareTable[thisChareType]->name, why);
+void CkMigratable::CkAbort(const char *format, ...) const {
+	char newmsg[256];
+	va_list args;
+	va_start(args, format);
+	vsnprintf(newmsg, sizeof(newmsg), format, args);
+	va_end(args);
+	::CkAbort("CkMigratable '%s' aborting: %s",_chareTable[thisChareType]->name, newmsg);
```
evan-charmworks: Would it be preferable to make the `CkMigratable` name message a separate call to `CkError` as in `ArrayElement`?

matthiasdiener: Yeah, that's a good idea. Changed in the most recent commit.

matthiasdiener:
> The class-level `CkAbort` functions aren't actually passing the arguments other than the format string to their callees. They would need to use a construct like this:

Couldn't we just interpolate the string directly in the CkAbort class functions?

evan-charmworks:
> > The class-level `CkAbort` functions aren't actually passing the arguments other than the format string to their callees. They would need to use a construct like this:
> 
> Couldn't we just interpolate the string directly in the CkAbort class functions (similar to what was done before the most recent commit for `CkMigratable::CkAbort`)?

Yes, that would also resolve the issue.

jcphill:
This is dangerous because you are changing the interpretation of the string passed to CkAbort, which now needs to be scrubbed of spurious % characters. Can you change this so that if there is only one argument it is not treated as a format string, e.g. make CkAbort(msg) equivalent to CkAbort("%s",msg).

jcphill:
If that isn't possible, the new behavior should be in a new function called CkAbortf.

PR #2579, opened 2019-10-23
Add attributes to class-level CkAbort function declarations

evan-charmworks:


evan-charmworks:
Fortunately I did not see any new warnings appear as a result of these attributes.

src/ck-core/ckarray.h
```diff
@@ -343,6 +343,10 @@ AllreduceMgr * allredMgr; // for allreduce
   CkArrayID thisArrayID;//My source array's ID
 
   //More verbose form of abort
+  CMK_NORETURN
+#if defined __GNUC__ || defined __clang__
+  __attribute__ ((format (printf, 2, 3)))
```
matthiasdiener: Isn't it `printf,1,2` (same for the others)?

evan-charmworks: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute:

> Since non-static C++ methods have an implicit this argument, the arguments of such methods should be counted from two, not one, when giving values for string-index and first-to-check.

I tried `printf, 1, 2` first without knowing this and got a compiler error, that `this` is not a C-string type.

PR #2581, opened 2019-10-26
HAPI: Remove dependency on Charm++ layer and fix WorkRequest instrumentation

### LINKED ISSUE #2467: Hapi Dependency on Charm++
Currently, Hapi depends on the Charm++ runtime (for `QdCreate`, `QdProcess` and `CkCallback::send`), which means that non-Charm++ applications (such as those that only use Converse) now require the Charm++ runtime to build and run. Converse itself requires Hapi for `hapiPollEvents` and `exitHybridAPI`.

See also #2459.

This causes the current auto build failure with cuda:
```
../../bin/charmc -optimize -production   -o check check.o -language converse++
../../bin/../lib/libcudahybridapi.a(hybridapi.o): In function `hapiPollEvents':
hapi_impl.cpp:(.text+0x1590): undefined reference to `CkCallback::send(void*) const'
hapi_impl.cpp:(.text+0x169a): undefined reference to `QdProcess(int)'
../../bin/../lib/libcudahybridapi.a(hybridapi.o): In function `hapiAddCallback':
hapi_impl.cpp:(.text+0x1afb): undefined reference to `QdCreate(int)'
../../bin/../lib/libcudahybridapi.a(hybridapi.o): In function `hapiEnqueue':
hapi_impl.cpp:(.text+0x1c41): undefined reference to `QdCreate(int)'
hapi_impl.cpp:(.text+0x1d17): undefined reference to `QdCreate(int)'
hapi_impl.cpp:(.text+0x1e56): undefined reference to `QdCreate(int)'
```

minitu:
Resolves #2467 by removing dependencies on the Charm++ layer: `CkVec`, `CkCallback`, `QdCreate`, and `QdProcess`.

### DIFF
--- a/src/arch/cuda/hybridAPI/hapi_impl.cpp
+++ b/src/arch/cuda/hybridAPI/hapi_impl.cpp
@@ -510,20 +510,12 @@ static void* hostToDeviceCallback(void* arg) {
   NVTXTracer nvtx_range("hostToDeviceCallback", NVTXColor::Asbestos);
 #endif
   hapiWorkRequest* wr = *((hapiWorkRequest**)((char*)arg + CmiMsgHeaderSizeBytes + sizeof(int)));
-  if (hapiInvokeCallback) {
-    hapiInvokeCallback(wr->host_to_device_cb);
-  }
-  else {
-    CmiAbort("hapiInvokeCallback is not set");
-  }
+  CmiAssert(hapiInvokeCallback);
+  hapiInvokeCallback(wr->host_to_device_cb);
 
   // inform QD that the host-to-device transfer is complete
-  if (hapiQdProcess) {
-    hapiQdProcess(1);
-  }
-  else {
-    CmiAbort("hapiQdProcess is not set");
-  }
+  CmiAssert(hapiQdProcess);
+  hapiQdProcess(1);
 
   return NULL;
 }
@@ -534,20 +526,12 @@ static void* kernelCallback(void* arg) {
   NVTXTracer nvtx_range("kernelCallback", NVTXColor::Asbestos);
 #endif
   hapiWorkRequest* wr = *((hapiWorkRequest**)((char*)arg + CmiMsgHeaderSizeBytes + sizeof(int)));
-  if (hapiInvokeCallback) {
-    hapiInvokeCallback(wr->kernel_cb);
-  }
-  else {
-    CmiAbort("hapiInvokeCallback is not set");
-  }
+  CmiAssert(hapiInvokeCallback);
+  hapiInvokeCallback(wr->kernel_cb);
 
   // inform QD that the kernel is complete
-  if (hapiQdProcess) {
-    hapiQdProcess(1);
-  }
-  else {
-    CmiAbort("hapiQdProcess is not set");
-  }
+  CmiAssert(hapiQdProcess);
+  hapiQdProcess(1);
 
   return NULL;
 }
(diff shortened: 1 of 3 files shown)


src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -1155,19 +1151,19 @@ hapiRequestTimeInfo* hapiQueryInstrument(int chare, char type, char phase) {
...
+  for (int chare = 0; chare < avg_times_.size(); chare++) {
+    for (int type = 0; type < avg_times_[chare].size(); type++) {
+      for (int phase = 0; phase < avg_times_[chare][type].size(); phase++) {
         avg_times_[chare][type][phase].transferTime = 0.0;
         avg_times_[chare][type][phase].kernelTime = 0.0;
         avg_times_[chare][type][phase].cleanupTime = 0.0;
         avg_times_[chare][type][phase].n = 0;
       }
-      avg_times_[chare][type].length() = 0;
+      avg_times_[chare][type].size() = 0;
```
matthiasdiener: Does this actually work? Why not use `resize()`?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -80,14 +80,12 @@ static inline int getMyCudaDevice(int my_node) {
...
-// the Charm++ callback.
-extern void CUDACallbackManager(void* fn);
-extern int CmiMyPe();
+// Used to invoke user's Charm++ callback function
+void (*hapiInvokeCallback)(void*, void*);
 
 // Functions used to support quiescence detection.
-extern void QdCreate(int n);
-extern void QdProcess(int n);
+void (*hapiQdCreate)(int);
```
matthiasdiener: Maybe set these explicitly to `NULL`.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -515,10 +512,10 @@ static void* hostToDeviceCallback(void* arg) {
   NVTXTracer nvtx_range("hostToDeviceCallback", NVTXColor::Asbestos);
 #endif
   hapiWorkRequest* wr = *((hapiWorkRequest**)((char*)arg + CmiMsgHeaderSizeBytes + sizeof(int)));
-  CUDACallbackManager(wr->host_to_device_cb);
+  hapiInvokeCallback(wr->host_to_device_cb);
 
   // inform QD that the host-to-device transfer is complete
-  QdProcess(1);
+  hapiQdProcess(1);
```
matthiasdiener: These function calls might crash if the function pointers haven't been initialized in `_initCharm`. Maybe add a check in `initHybridAPI` that these pointers are not `NULL` and `CkAbort` the application if they are.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -515,10 +510,20 @@ static void* hostToDeviceCallback(void* arg) {
   NVTXTracer nvtx_range("hostToDeviceCallback", NVTXColor::Asbestos);
 #endif
   hapiWorkRequest* wr = *((hapiWorkRequest**)((char*)arg + CmiMsgHeaderSizeBytes + sizeof(int)));
-  CUDACallbackManager(wr->host_to_device_cb);
+  if (hapiInvokeCallback) {
```
stwhite91: Instead of all these if-else, why not just assert that it is set?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -861,16 +911,19 @@ static inline void hapiWorkRequestStartTime(hapiWorkRequest* wr) {
 static inline void profileWorkRequestEvent(hapiWorkRequest* wr,
                                            WorkRequestStage event) {
 #ifdef HAPI_INSTRUMENT_WRS
-  if (initializedInstrument()) {
+#if CMK_SMP || CMK_MULTICORE
```
stwhite91: Doesn't multicore also set CMK_SMP?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -887,20 +940,26 @@ static inline void profileWorkRequestEvent(hapiWorkRequest* wr,
         printf("[HAPI] invalid event during profileWorkRequestEvent\n");
     }
   }
+  else {
+    printf("[HAPI] instrumentation not initialized!\n");
```
stwhite91: CmiPrintf

minitu: Will submit a separate patch to change all printf to CmiPrintf.

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: You're right, will change this in a separate patch.

PR #2584, opened 2019-10-29
cmake: fall back to ./build when cmake too old/nonexistent

matthiasdiener:


buildcmake
```diff
@@ -196,9 +228,13 @@ while [[ $# -gt 0 ]]; do
       ;;
 
   # Miscellaneous options
-    -j*)
+    -j[0-9]*)
       opt_parallel=$arg
       ;;
+    -j)
+      opt_parallel="$arg$1"
```
evan-charmworks: `-j` a valid parameter by itself (without a number following it). `$arg$1` will unconditionally eat the argument following `-j`.

matthiasdiener: Good point. Should be fixed now.

PR #2585, opened 2019-10-29
Charmrun: Fix ++ssh-display without ++batch

evan-charmworks:
Tested on the Adaptivity Demo cluster.


evan-charmworks:
This PR is useful for debugging on clusters where the head node that runs charmrun can only be accessed through SSH, and is **vital** in the case of syncft where launching a new process does not go through the batch mode path.

PR #2586, opened 2019-10-30
HAPI: Remove redundant CMK_MULTICORE, replace printf with CmiPrintf

minitu:
Remaining updates from #2581.


evan-charmworks:
Why is `CMK_MULTICORE` extraneous here?

minitu:
@evan-charmworks All multicore archs (at least the ones I've checked) define `CMK_SMP` as well as `CMK_MULTICORE`, so using `CMK_SMP` is sufficient.

PR #2596, opened 2019-11-01
ZC Bcast API: Fix de-registration bug on parent nodes after reverse op

nitbhat:
Previously, on intermediate nodes, when rgets on the child nodes were being
executed as reverse operations (PUT from source instead of a GET from destination),
the source buffer was de-registered twice on the intermediate nodes, causing a failure.
This de-registration (for UNREG modes) is only applicable to root nodes.
With this commit, the root node for a ZC Bcast is stored in NcpyOperationInfo
and de-registration is performed only when the reverse operation involves
the root node as the source. (and not when an intermediate node is the source)

### DIFF
--- a/src/ck-core/ckrdma.C
+++ b/src/ck-core/ckrdma.C
@@ -694,7 +694,7 @@ void performEmApiCmaTransfer(CkNcpyBuffer &source, CkNcpyBuffer &dest, CmiSpanni
 }
 #endif
 
-void performEmApiRget(CkNcpyBuffer &source, CkNcpyBuffer &dest, int opIndex, char *ref, int extraSize, unsigned int rootNode, ncpyEmApiMode emMode) {
+void performEmApiRget(CkNcpyBuffer &source, CkNcpyBuffer &dest, int opIndex, char *ref, int extraSize, int rootNode, ncpyEmApiMode emMode) {
 
   int layerInfoSize = CMK_COMMON_NOCOPY_DIRECT_BYTES + CMK_NOCOPY_DIRECT_BYTES;
   if(dest.regMode == CK_BUFFER_UNREG) {
@@ -750,7 +750,7 @@ void performEmApiRget(CkNcpyBuffer &source, CkNcpyBuffer &dest, int opIndex, cha
   // on the comm. thread as the message is being inside this for loop on the worker thread
 }
 
-void performEmApiNcpyTransfer(CkNcpyBuffer &source, CkNcpyBuffer &dest, int opIndex, CmiSpanningTreeInfo *t, char *ref, int extraSize, CkNcpyMode ncpyMode, unsigned int rootNode, ncpyEmApiMode emMode){
+void performEmApiNcpyTransfer(CkNcpyBuffer &source, CkNcpyBuffer &dest, int opIndex, CmiSpanningTreeInfo *t, char *ref, int extraSize, CkNcpyMode ncpyMode, int rootNode, ncpyEmApiMode emMode){
 
   switch(ncpyMode) {
     case CkNcpyMode::MEMCPY: performEmApiMemcpy(source, dest, emMode);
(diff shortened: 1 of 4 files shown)


nitbhat:
This bug went undetected because autobuild was reporting `Good` even though the bcast test was failing. (http://charm.cs.illinois.edu/autobuild/old.2019_10_23__01_01/gni-crayxc.txt). 

```
Running on 6 processors:  ./zerocopy_with_qd 142 +noCMAForZC
srun -n 6 -c 2 ./zerocopy_with_qd 142 +noCMAForZC
Charm++> Running on Gemini (GNI) with 6 processes
Charm++> static SMSG
Charm++> SMSG memory: 29.6KB
Charm++> memory pool init block size: 8MB, total memory pool limit 0MB (0 means no limit)
Charm++> memory pool registered memory limit: 200000MB, send limit: 100000MB
Charm++> Cray TLB page size: 8192K
Charm++> Running in non-SMP mode: 6 processes (PEs)
Converse/Charm++ Commit ID: cfec53c0a
Charm++> Running on 1 hosts (2 sockets x 16 cores x 2 PUs = 64-way SMP)
[0][0][0] Test 1: QD has been reached for RO Variable Bcast
[0][0][0] Test 2: QD has been reached for Direct API
[0][0][0] Test 3: QD has been reached for EM Send API
[0][0][0] Test 4: QD has been reached for EM Post API
[4] GNI_MemDeregister failed!; err=GNI_RC_INVALID_PARAM
[4] Stack Traceback:
  [4:0] zerocopy_with_qd 0x20160cf6 CmiAbortHelper(char const*, char const*, char const*, int, int)
  [4:1] zerocopy_with_qd 0x20160e15 
  [4:2] zerocopy_with_qd 0x201681fe 
  [4:3] zerocopy_with_qd 0x20121b97 deregisterMemFromMsg(envelope*, bool)
  [4:4] zerocopy_with_qd 0x20122ea4 CkRdmaEMBcastAckHandler(void*)
  [4:5] zerocopy_with_qd 0x2016b7fb CsdScheduleForever
  [4:6] zerocopy_with_qd 0x2016bab5 CsdScheduler
  [4:7] zerocopy_with_qd 0x2016ac8a ConverseInit
  [4:8] zerocopy_with_qd 0x200b2927 charm_main
  [4:9] libc.so.6 0x2aaaae6b2f8a __libc_start_main
```

I'm going to investigate how and why that happened.

src/util/cmirdmautils.h
```diff
@@ -53,6 +53,8 @@ typedef struct ncpystruct{
 
   int ncpyOpInfoSize;
 
+  unsigned int rootNode; // used only for Broadcast, -1 for p2p operations
```
stwhite91: Why unsigned if -1 is valid?

nitbhat: Ah, yes. I'll remove the unsigned. (-1 is valid and will be used for p2p operations, which is when the field is unused).

nitbhat: Done.

src/ck-core/ckrdma.C
```diff
@@ -1555,9 +1559,10 @@ void handleBcastReverseEntryMethodApiCompletion(NcpyOperationInfo *info) {
     invokeRemoteNcpyAckHandler(info->destPe, info->refPtr, ncpyHandlerIdx::EM_ACK);
   }
 #if CMK_REG_REQUIRED
-  // De-register source for reverse operations when regMode == UNREG and deregMode == DEREG
-  if(info->isSrcRegistered == 1 && info->srcRegMode == CK_BUFFER_UNREG && info->srcDeregMode == CK_BUFFER_DEREG)
-    deregisterSrcBuffer(info);
+  // De-register source for reverse operations when regMode == UNREG and deregMode == DEREG on the root node
+  if(info->rootNode == CkMyNode())
```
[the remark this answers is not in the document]
nitbhat: I think this code is only exercised if CMK_ONESIDED_IMPL is true and it is disabled for bigsim builds, so it wouldn't matter.

Issue #2599, opened 2019-11-04
automake: warning: am_libhwloc_embedded_la_rpath was already defined in condition TRUE, which includes condition !HWLOC_BUILD_STANDALONE


matthiasdiener:
Building Charm++ results in the following warning message on multiple platforms (see e.g. https://github.com/spack/spack/issues/11125, https://lists.cs.illinois.edu/lists/arc/charm/2019-06/msg00008.html, https://my.cdash.org/viewBuildError.php?buildid=1722945):
```
autoreconf
autoheader
automake: warning: am_libhwloc_embedded_la_rpath was already defined in condition TRUE, which includes condition !HWLOC_BUILD_STANDALONE ...
configure.ac:2973: ... 'am_libhwloc_embedded_la_rpath' previously defined here
```

evan-charmworks:
This warning is an artifact of the only way I've found to get hwloc to build a shared libary while in standalone/embedded mode. It would require an upstream change or a patch to our copy of the library to resolve.

evan-charmworks:
For reference, this is the hack in `configure.ac` that is the source of the warning message:

```bash
if test "$CMK_NO_BUILD_SHARED" = 'false'
then
  enable_shared='yes'
  export am_libhwloc_embedded_la_rpath="-rpath $CHARMLIBSO"
else
  enable_shared='no'
  export am_libhwloc_embedded_la_rpath=''
fi
AC_SUBST(am_libhwloc_embedded_la_rpath)
```

matthiasdiener:
Considering the discussion in https://github.com/open-mpi/hwloc/issues/371, we have a few options:

- Build without embedded mode
- Build only the static hwloc library, even when building Charm++ with shared libraries
- Embed hwloc into `libconv-core.so` and `libconv-core.a`
- Merge all Charm++ libraries into one

ericjbohm:
Embedded mode allows us to control the hwloc version in use, so I would advocate for retaining that approach.  I don't see much downside to statically embedding it in one of the core libs.

Regardless, presenting the users with one -lcharm is a desirable option entirely independent of the hwloc issues.  Having the other libs available separately is a good idea for various niche cases where you don't want the kitchen sink.

matthiasdiener:
The underlying warning is fixed via #2601. However, this is still a hack, and we should consider implementing a more stable solution.

PR #2601, opened 2019-11-05
Bug #2599: workaround for building shared hwloc library

matthiasdiener:
Avoids the automake warning discussed in #2599.


evan-charmworks:
Did you test this on Linux, or just Darwin? This seems to break --build-shared on Linux.

PR #2602, opened 2019-11-06
charmc/ampicc: pass through certain compiler options even without file to compile

matthiasdiener:
Helps with (e.g.) configure scripts that parse the output of `-v`, `--version`, `-print-search-dirs` to detect compiler configurations.

### DIFF
--- a/src/libs/ck-libs/ampi/ampiCC
+++ b/src/libs/ck-libs/ampi/ampiCC
@@ -14,7 +14,7 @@ FSGLOBALS=''
 PIPGLOBALS=''
 EXPLICIT_COMPILATION=''
 BUILD_SHARE=''
-SHOW_COMPILER_VERSION=''
+PASSTHROUGH_COMPILER_OPTIONS=''
 
 ARGS=''
 OBJECT=''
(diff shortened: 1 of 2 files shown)


src/libs/ck-libs/ampi/ampiCC
```diff
@@ -75,6 +80,14 @@ done
 
 eval processArgs "$@"
 
+# If there are no args except -v or --version specified, just run
+# charmc with -v or --version. Helps with configure scripts that
+# parse the output to detect compiler versions.
+if [[ -z $ARGS ]] && [[ -n $SHOW_COMPILER_VERSION ]]; then
```
evan-charmworks: Should `$ARGS` and `$SHOW_COMPILER_VERSION` be quoted here and once below or is the quoting practice for bash's `[[` different from `[` and `test`?

matthiasdiener: Yeah, with `[[` you don't need to quote variables anymore (see e.g. https://stackoverflow.com/questions/3869072/test-for-non-zero-length-string-in-bash-n-var-or-var). If you prefer, I can add quotes if it makes it more consistent with the rest of the script.

matthiasdiener:
Don’t merge this yet please. I’m still working on a small improvement

src/libs/ck-libs/ampi/ampiCC
matthiasdiener: Added quotes.

matthiasdiener:
Ready for review.

PR #2603, opened 2019-11-07
AMPI: add support for using liveViz from AMPI, with a wave2d example code

stwhite91:


matthiasdiener:
Verbs error is an unrelated error hat can be fixed by merging this branch with master.

src/libs/ck-libs/Makefile
```diff
@@ -20,6 +20,10 @@ $(SIMPLE_DIRS) ParFUM ParFUM-Tops $(FEM_DIRS):
 
 FEM: $(FEM_DIRS) 
 
+ifeq ($(AMPI_WITH_LIVE_VIZ),1)
```
evan-charmworks: How is this specified? You can pass `-DAMPI_WITH_LIVE_VIZ=1` for all other uses but this would require something different.

PR #2611, opened 2019-11-13
Separate checkpoint files into subdirectories

minitu:
ChaNGa is having an issue with corrupt checkpoint files on Frontera, which is caused by too many files being stored in a single directory. This patch groups the PEs by 256 and directs the checkpoint files to separate subdirectories. For example, with 512 PEs, files from the first 256 PEs will be placed in `sub0` and files from the second group in `sub1`.


minitu:
Tested with `tests/charm++/chkpt`.

src/ck-core/ckcheckpoint.C
```diff
@@ -214,14 +220,25 @@ void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _reque
...
+  if (CmiNumPartitions() > 1) {
+    addPartitionDirectory(dirPath);
+    CmiMkdir(dirPath.str().c_str());
+  }
+
+  // Create subdirectories by grouping PEs
+  // Nodegroups are fine since # of logical nodes <= # of PEs
+  int numPes = CkNumPes();
+  int numPeGroups = ceil((double)numPes / PE_GROUP_SIZE);
+  for (int i = 0; i < numPeGroups; i++) {
```
rbuch: Do we want all of the PEs to create all of the directories? The partitions code only creates the directory for the current partition.

src/ck-core/ckcheckpoint.C
```diff
@@ -214,14 +220,25 @@ void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _reque
...
+    dirPath << "/sub" << i;
+    CmiMkdir(dirPath.str().c_str());
+  }
+
+  bool success = true;
```
rbuch: I think this block should match the indentation of the surrounding code.

src/ck-core/ckcheckpoint.C
```diff
@@ -214,14 +220,25 @@ void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _reque
...
+    addPartitionDirectory(dirPath);
+    CmiMkdir(dirPath.str().c_str());
+  }
+
+  // Create subdirectories by grouping PEs
+  // Nodegroups are fine since # of logical nodes <= # of PEs
+  int numPes = CkNumPes();
+  int numPeGroups = ceil((double)numPes / PE_GROUP_SIZE);
+  for (int i = 0; i < numPeGroups; i++) {
+    dirPath << "/sub" << i;
```
rbuch: Won't this keep making dirPath longer and longer for every subgroup? (e.g. `/sub0sub1sub2...`)

src/ck-core/ckcheckpoint.C
sanjaykale: there may be a race condition if you assume another pe has created the folder... but can be fixed by looping on error condition by non-creator PEs.  Right?

minitu: It should be fine to have a race with `CmiMkdir`, calls after the directory is created shouldn't do anything. There is similar race already with `CmiMkdir(dirname)` in line 222. I'll change the code to have each PE try to create a folder only for its PE group (PEs 0-255 try to create folder for PE group 0, PEs 256-511 for PE group 1, etc.). I think it's necessary to have 256 PEs try to create a folder because PEs in a group might be on different physical nodes.

src/ck-core/ckcheckpoint.C
minitu: The above will fix this problem.

src/ck-core/ckcheckpoint.C
minitu: Agreed, will be fixed.

src/ck-core/ckcheckpoint.C
```diff
@@ -214,14 +220,22 @@ void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _reque
...
 
+	// Create partition directories (if applicable)
+	ostringstream dirPath;
+	dirPath << dirname;
+	if (CmiNumPartitions() > 1) {
+		addPartitionDirectory(dirPath);
+		CmiMkdir(dirPath.str().c_str());
+	}
+
+	// Create subdirectories by grouping PEs
```
rbuch: I'd add some more descriptive comments about why this change was made to the code, something like: 
```suggestion
	// Due to file system issues we have observed, divide checkpoints
	// into subdirectories to avoid having too many files in a single directory 
```

minitu:
@rbuch @sanjaykale There is a hang issue that is not caused by this patch: #2613

src/ck-core/ckcheckpoint.C
```diff
@@ -30,6 +30,8 @@ void noopit(const char*, ...)
 #define DEBUGC(x) x
 //#define DEBUGC(x) 
 
+#define PE_GROUP_SIZE 256
```
rbuch: I think this (and other instances where "group" is used in this way) should be changed to avoid confusion with Charm++ groups. Maybe "subdirectory" or something like that.

PR #2612, opened 2019-11-14
UCX: Fix eps initialization

brminich:
Need to initialize `maxval` by `0` to avoid it having a garbage when
`runtime_get_max_vallen()` returns (we cast it to `int` as runtime API requires)

Fixes one of the problems reported in #2597

### DIFF
--- a/src/arch/ucx/machine.C
+++ b/src/arch/ucx/machine.C
@@ -141,18 +141,17 @@ void UcxRequestInit(void *request)
 
 static void UcxInitEps(int numNodes, int myId)
 {
-    size_t maxval = 0, addrlen = 0;
-    size_t len, partLen;
+    size_t addrlen;
     ucp_address_t *address;
     ucs_status_t status;
     ucp_ep_params_t eParams;
     ucp_ep_h ep;
-    int i, j, ret, peer, maxkey, parts;
+    int i, j, ret, peer, maxkey, maxval, parts, len, partLen;
     char *keys, *addrp, *remoteAddr;
 
     ret = runtime_get_max_keylen(&maxkey);
     UCX_CHECK_PMI_RET(ret, "UcxInitEps: runtime_get_max_keylen error");
-    ret = runtime_get_max_vallen((int*)&maxval);
+    ret = runtime_get_max_vallen(&maxval);
     UCX_CHECK_PMI_RET(ret, "UcxInitEps: runtime_get_max_vallen error");
 
     // Reduce maxval value, because with PMI1 it has to fit cmd + key + value
@@ -167,6 +166,7 @@ static void UcxInitEps(int numNodes, int myId)
 
     status = ucp_worker_get_address(ucxCtx.worker, &address, &addrlen);
     UCX_CHECK_STATUS(status, "UcxInitEps: ucp_worker_get_address error");
+    CmiEnforce(addrlen < std::numeric_limits<int>::max()); //address should fit to int
 
     parts = (addrlen / maxval) + 1;
 
@@ -177,7 +177,7 @@ static void UcxInitEps(int numNodes, int myId)
     UCX_CHECK_PMI_RET(ret, "UcxInitEps: runtime_kvs_put error");
 
     addrp = (char*)address;
-    len   = addrlen;
+    len   = (int)addrlen;
     for (i = 0; i < parts; ++i) {
         partLen = std::min(maxval, len);
         ret = snprintf(keys, maxkey, "UCX-%d-%d", myId, i);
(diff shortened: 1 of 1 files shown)


brminich:
the CI failure looks like hang with MPI ML. Should I restart the testing somehow?

trquinn:
I tried this PR with the #2597 problem.  It still fails similarly (here with everything -optimize):
```
Running on 2 processors:  ./ChaNGa.smp +useckloop +bench -wall 60 -maxbal 10.0 +
ppn 27 +setcpuaffinity +commap 0,1 +pemap 2-54:2,3-55:2 
charmrun>  /bin/setarch x86_64 -R  mpirun -np 2  ./ChaNGa.smp +useckloop +bench 
-wall 60 -maxbal 10.0 +ppn 27 +setcpuaffinity +commap 0,1 +pemap 2-54:2,3-55:2 
Charm++> Running in SMP mode: 2 processes, 27 worker threads (PEs) + 1 comm thre
ads per process, 54 PEs total
Converse/Charm++ Commit ID: v6.10.0-rc2-48-gca65d76d1
Warning> Randomization of virtual memory (ASLR) is turned on in the kernel, thre
ad migration may not work! Run 'echo 0 > /proc/sys/kernel/randomize_va_space' as
 root to disable it, or try running with '+isomalloc_sync'.
Charm++> cpuaffinity PE-core map : 2-54:2,3-55:2
Charm++> Running on 1 hosts (2 sockets x 28 cores x 1 PUs = 56-way SMP)
------------- Processor 54 Exiting: Called CmiAbort ------------
Reason: Unhandled C++ exception in user code.

[54] Stack Traceback:
  [54:0] ChaNGa.smp 0x80202f CmiAbortHelper(char const*, char const*, char const
*, int, int)
  [54:1] ChaNGa.smp 0x802150 CmiAsyncBroadcastFn
  [54:2] ChaNGa.smp 0x717ef0 _initDone()
  [54:3] libstdc++.so.6 0x2b31bba6e106 
  [54:4] libstdc++.so.6 0x2b31bba6e151 
  [54:5] libstdc++.so.6 0x2b31bba6e385 
  [54:6] libstdc++.so.6 0x2b31bba621c6 
  [54:7] ChaNGa.smp 0x57c4fb 
  [54:8] 0x2b31b9e9f010
```

matthiasdiener:
> @brminich Please update this PR, changing the type of `maxval` to an int and removing the problematic cast in the call to `runtime_get_max_vallen`. If there is an issue with the later uses in `std::min`, change them to `std::min((size_t)maxval, len)`.

With this change, isn't there going to be an issue with the `addrp` calculation around line 333?

brminich:
I'll have to do some casting anyway, because runtime API and UCX API require different type of parameters: runtime API rely on `int`, while UCX API use `size_t`

> Since size_t is unsigned, CmiEnforce(maxval > 0) is equivalent to CmiEnforce(maxval != 0), i.e., would-be negative values are actually huge.

yes, it is check for non `0` values. Can change to `!=`

jcphill:
> I'll have to do some casting anyway, because runtime API and UCX API require different type of parameters: runtime API rely on `int`, while UCX API use `size_t`
> 
> > Since size_t is unsigned, CmiEnforce(maxval > 0) is equivalent to CmiEnforce(maxval != 0), i.e., would-be negative values are actually huge.
> 
> yes, it is check for non `0` values. Can change to `!=`

It looks like it should be checking for negative values, unless you have some guarantee that maxval >= 48:

    maxval -= 48;
    CmiEnforce(maxval > 0);

brminich:
> > I'll have to do some casting anyway, because runtime API and UCX API require different type of parameters: runtime API rely on `int`, while UCX API use `size_t`
> > > Since size_t is unsigned, CmiEnforce(maxval > 0) is equivalent to CmiEnforce(maxval != 0), i.e., would-be negative values are actually huge.
> > 
> > 
> > yes, it is check for non `0` values. Can change to `!=`
> 
> It looks like it should be checking for negative values, unless you have some guarantee that maxval >= 48:
> 
> ```
> maxval -= 48;
> CmiEnforce(maxval > 0);
> ```

right, better use signed type (even though maxval is not less than 256 with all PMI versions).
I'll revise types once #2562  is reverted.

brminich:
will open a new one

nitbhat:
New patch that contains fixes: https://github.com/UIUC-PPL/charm/pull/2624

@brminich: Is there a reason that you prefer creating a new PR rather than work on the existing one? It would be better to rework on the existing PR so that we have the review history.

nitbhat:
This commit has been cherry-picked on top of charm-6.10.0.

PR #2615, opened 2019-11-20
Zcpy API: Remove unnecessary for loop from getSrcPe() method

nitbhat:


src/ck-core/ckrdma.C
```diff
@@ -854,11 +854,10 @@ int getSrcPe(envelope *env) {
   PUP::fromMem up((void *)((CkMarshallMsg *)EnvToUsr(env))->msgBuf);
   up|numops;
   up|rootNode;
-  for(int i=0; i<numops; i++){
-    CkNcpyBuffer w;
-    up|w;
-    return w.pe;
-  }
+  CmiEnforce(numops > 0);
```
matthiasdiener: Could be CmiAssert maybe

src/ck-core/ckrdma.C
```diff
@@ -854,11 +854,10 @@ int getSrcPe(envelope *env) {
   PUP::fromMem up((void *)((CkMarshallMsg *)EnvToUsr(env))->msgBuf);
   up|numops;
```
stwhite91: PUPing numops is useless here and below in getRootNode(). That can be done here or in a separate patch

matthiasdiener: Isn't the goal of this just to unpack in the same way it was packed before?

nitbhat: Unpacking numops and rootNode is required since that was how it was packed.

src/ck-core/ckrdma.C
nitbhat: Doesn't CmiAssert do nothing except for when built with '--enable-error-checking'?

matthiasdiener: > Doesn't CmiAssert do nothing except for when built with '--enable-error-checking'?

Yes, I think so. I was just wondering if we should check the value of numops here in all build configs. If it's better with the check enabled, that's fine too.

nitbhat:
This commit has been cherry-picked on top of charm-6.10.0

PR #2622, opened 2019-11-25
UCX: Replace recursion with iteration in UcxPostRxReq

evan-charmworks:
@brminich, do you see anything wrong with this patch? I believe each request will end up with the same idx as it would have without the patch, but the order in which `handleOneRecvedMsg` is called and `ucxCtx.rxReqs[idx]` is assigned is now FIFO instead of LIFO. Hopefully that won't break anything.

### DIFF
--- a/src/arch/ucx/machine.C
+++ b/src/arch/ucx/machine.C
@@ -337,17 +337,16 @@ static inline UcxRequest* UcxPostRxReq(ucp_tag_t tag, size_t size,
                                        ucp_tag_message_h msg)
 {
     UcxRequest *req = UcxPostRxReqInternal(tag, size, msg);
+    int idx = req->idx;
 
     do {
         if (req->completed) {
-            int idx = req->idx;
-
             UCX_REQUEST_FREE(req);
 
             if (tag & UCX_MSG_TAG_EAGER) {
                 req = UcxPostRxReqInternal(UCX_MSG_TAG_EAGER, ucxCtx.eagerSize, NULL);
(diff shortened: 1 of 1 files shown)


src/arch/ucx/machine.C
```diff
@@ -323,28 +323,41 @@ static inline UcxRequest* UcxPostRxReq(ucp_tag_t tag, size_t size,
...
+    int idx = req->idx;
+
+    do {
+        if (req->completed) {
+            UCX_REQUEST_FREE(req);
+
+            if (tag & UCX_MSG_TAG_EAGER) {
+                req = UcxPostRxReqInternal(UCX_MSG_TAG_EAGER, ucxCtx.eagerSize, NULL);
+                ucxCtx.rxReqs[idx] = req;
+                std::swap(req->idx, idx);
```
brminich: swap seems to be excessive, because you are working with the same reqs element (so `idx` will remain the same)
can do just `req->idx = idx;`

evan-charmworks: I see. Is it guaranteed that every req will have the same idx? I wasn't sure if I could assume that here, so I used the swap to preserve idx assignment relationship from the recursive version of the function.

brminich: here you are posting new request to the specific index of the requests array. If  it is completed immediately, you try to repost it again, but the index remains the same.

evan-charmworks:
I've replaced the `std::swap` with an assignment.

PR #2619, opened 2019-11-21
Revert new ucx changes which are introducing crashes

nitbhat:
Reverting should fix:
https://github.com/UIUC-PPL/charm/issues/2597
https://github.com/UIUC-PPL/charm/issues/2616

Reverting this will cause one of the ChaNGa test cases to fail (reported by @trquinn), so we'd still have to find a fix for that crash generated by the recursion shown. However, the generic case of running charm programs will still work, which is more important.  
```
Core was generated by `./ChaNGa.ucx +ppn 55 +setcpuaffinity +commap 0 +pemap 1-55 h148.cosmo50PLK.6144'.
Program terminated with signal 11, Segmentation fault.
#0  0x00002af5abfe0ab9 in sysmalloc () from /lib64/libc.so.6
warning: File "/opt/apps/gcc/8.3.0/lib64/libstdc++.so.6.0.25-gdb.py" auto-loading has been declined by your `auto-load safe-path' set to "$debugdir:$datadir/auto-load:/usr/bin/mono-gdb.py".
To enable execution of this file add
       add-auto-load-safe-path /opt/apps/gcc/8.3.0/lib64/libstdc++.so.6.0.25-gdb.py
line to your configuration file "/home1/00333/tg456090/.gdbinit".
To completely disable this security protection add
       set auto-load safe-path /
line to your configuration file "/home1/00333/tg456090/.gdbinit".
For more information about this security protection see the
"Auto-loading safe path" section in the GDB manual.  E.g., run from the shell:
       info "(gdb)Auto-loading safe path"
Missing separate debuginfos, use: debuginfo-install glibc-2.17-260.el7_6.6.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 libibcm-41mlnx1-OFED.4.1.0.1.0.46101.x86_64 libibverbs-41mlnx1-OFED.4.6.0.4.1.46101.x86_64 libjpeg-turbo-1.2.90-6.el7.x86_64 libmlx5-41mlnx1-OFED.4.6.0.0.4.46101.x86_64 libnl3-3.2.28-4.el7.x86_64 librdmacm-41mlnx1-OFED.4.6.0.0.1.46101.x86_64 lustre-client-2.10.7_ddn8-1.el7.x86_64 ncurses-libs-5.9-14.20130511.el7_4.x86_64 numactl-libs-2.0.9-7.el7.x86_64 readline-6.2-10.el7.x86_64 zlib-1.2.7-18.el7.x86_64
(gdb) where
[...]

PR #2624, opened 2019-11-26
UCX: Fix eps initialization

brminich:
Use int for fetching maximum PMI value length by `runtime_get_max_vallen()`.
Otherwise it may contain a garbage in the most significant bytes when`runtime_get_max_vallen()` returns.


brminich:
just created on top of #2619.
Is it  better to rebase/merge and continue with #2612?

nitbhat:
> just created on top of #2619.
> Is it better to rebase/merge and continue with #2612?

Yes, that would be better because we have the review history in the PR.

brminich:
closing, fixin in #2612 
sorry for the confusion

PR #2625, opened 2019-11-26
build: switch to cmake build by default

matthiasdiener:
Switch to the cmake build system by default.
The old build system is still fully functional and can be used by running `./buildold` instead of `./build`. 

The old build system is used automatically:
- On Windows (where the cmake build doesn't work fully yet).
- On `gni-` and `-crayx?` builds (where the cmake build doesn't work fully yet).
- If cmake is not installed.
- If the cmake version is too old (we need at least version 3.4).

Note that this **will** break things, as not all possible combinations of options have been tested. It should be good enough for the common scenarios.

Additional failures are tracked in #2839.


.travis.yml
```diff
@@ -17,8 +17,8 @@ matrix:
     script:
     - sudo apt-get update && sudo apt-get install -y mpich libmpich-dev
     - ./build LIBS mpi-linux-x86_64 smp -j3 --with-production --enable-error-checking --enable-lbuserdata -g -Werror=vla || exit 1
-    - make -C mpi-linux-x86_64-smp/tests TESTOPTS="+setcpuaffinity"
-    - make -C mpi-linux-x86_64-smp/examples TESTOPTS="+setcpuaffinity"
+    - make -C mpi-linux-x86_64/tests TESTOPTS="+setcpuaffinity"
+    - make -C mpi-linux-x86_64/examples TESTOPTS="+setcpuaffinity"
```
evan-charmworks: Shouldn't buildcmake maintain buildold's directory naming conventions?

matthiasdiener: most recent commit should fix it.

matthiasdiener:
Current status is shown here: https://charm.cs.illinois.edu/autobuild/old.2020_04_01__19_52/

matthiasdiener:
> Should `$@` be `"$@"`?

Not sure, I tried both ways and it didn't seem to make a difference.

matthiasdiener:
This PR is ready for review. It passes autobuild in the same way as the old build system. As mentioned in the updated commit message, the new build system is disabled for Windows and gni/crayx, support for which is future work.

evan-charmworks:
Thanks for adding fsglobals and pipglobals support to this. It looks like the Darwin build failed because it tried to build the pipglobals object. The Makefile does this, which would need to be replicated in CMake.

```
ifeq (1,$(CMK_SUPPORTS_PIPGLOBALS))
  AMPI_TARGET += $(FUNCPTR_PIPGLOBALS_OBJ)
endif
ifeq (1,$(CMK_SUPPORTS_FSGLOBALS))
  AMPI_TARGET += $(FUNCPTR_FSGLOBALS_OBJ)
endif
```

matthiasdiener:
> CMK_SUPPORTS_PIPGLOBALS

Thanks, should be fixed now.

minitu:
What's the reason for renaming the script to `buildold`? CMake doesn't use this script, right?

matthiasdiener:
> What's the reason for renaming the script to `buildold`? CMake doesn't use this script, right?

Yes, CMake doesn't use `./buildold`. However, we still need to use `./buildold` for the unsupported configs mentioned in the first comment in this PR, as well as a fallback in case certain configs fail. My hope is to remove the old build system for 6.12.

minitu:
I guess my point was that we don't need to rename the script and leave it as `build` until it gets removed.

matthiasdiener:
> I guess my point was that we don't need to rename the script and leave it as `build` until it gets removed.

Oh, the reason is that by renaming it, people will actually use the new build system (while still having the possibility to run the old one in case things break). Testing the new system without doing this change is almost impossible (e.g., autobuild won't really exercise it; few (if any) people have actually tried running the CMake build since it was merged last year).

minitu:
One side point: can define flags be supported by something like https://cmake.org/cmake/help/v3.0/command/add_definitions.html?

matthiasdiener:
> https://cmake.org/cmake/help/v3.0/command/add_definitions.html

You mean, when `./build` is run with a `-D` option? Yes, I think that would be possible.

minitu:
I mean if `-D` can be supported with CMake, as you said it's not supported in the manual. I've used `-D` with the original `./build` and it gets passed in correctly.

PR #2628, opened 2019-11-27
HAPI: Fix and update CUDA Callback based implementation

minitu:
Various fixes to make the CUDA Callback based implementation of HAPI work properly.
Also updated `cudaStreamAddCallback` to `cudaLaunchHostFunc` as it is set for deprecation.
There is no performance difference between the two, and this implementation is still significantly slower (2-4x) than the CUDA Event based implementation.

Should be merged after #2626.


evan-charmworks:
Would it be better to use a struct for the message instead of doing pointer arithmetic? Something like

```c++
struct hapiCallbackMessage {
  char header[CmiMsgHeaderSizeBytes];
  int rank;
  void * cb;
  void * cb_msg;
  void * fixme;
};
```

minitu:
@evan-charmworks That is a good idea, I'll look into that.

minitu:
@matthiasdiener Could you review this?

Issue #2629, opened 2019-11-28
DistributedLB crashes execution on LeanMD mini-app


viniciusmctf:
A sample execution log:
> Generating topology 1 for 480 elements
> [0] Total work/step = 0.930041 sec
> Building a 22 x 22 mesh, with 4 missing elements
> calibrated iterations 35928846
> TIME PER STEP    10    3.133226    0.236043
> TIME PER STEP    20    3.452940    0.319715
> [DistributedLB] Load balancing step 0 starting at 4.135175
> DistLB> DistributedLB: step 0 starting at 4.135195 Memory: 5.031250 MB
> In DistributedLB strategy at 4.135250
> DistributedLB>>>Before LB: max = 1.379950, avg = 0.728562, ratio = 1.894073
> DistributedLB>>>Gossip finished at 4.138232 (0.002982 elapsed)
> DistributedLB>>>After phase 1: max = 1.037523, avg = 0.728562, ratio = 1.424070
> DistributedLB>>>After phase 2: max = 0.880921, avg = 0.728562, ratio = 1.209123
> DistributedLB>>>After phase 3: max = 0.868374, avg = 0.728562, ratio = 1.191901
> DistributedLB>>>After phase 4: max = 0.868374, avg = 0.728562, ratio = 1.191901
> DistributedLB>>>After phase 5: max = 0.868374, avg = 0.728562, ratio = 1.191901
> DistributedLB>>>Balancing completed at 4.139300 (0.004050 elapsed)
> DistributedLB>>>146 total migrations with 4 negative ack max
> DistributedLB> Strategy took 0.004119s memory usage: 5.031250 MB.

DistributedLB won't go after the DistBaseLB phase, which indicates the count of migrates_expected is not matching the total number of migrations accounted in the migration message.

This is most likely related to the LB agent accounting itself as a possible migration target, as DistBase won't accept migrations from within the LB agent (as this is not a case stated in the ProcessMigrationDecision function in DistBaseLB).

In order to avoid this issue, I have modified DistributedLB.C in order to stop it from sending tasks to itself, but a fool-proof solution would be to update the ProcessMigrationDecision() in DistBaseLB.C

PR #2630, opened 2019-11-28
Distributed will not try to migrate tasks to itself

viniciusmctf:
Related to issue #2629 , stop DistributedLB agents from migrating tasks to itself, in order to avoid execution stall in some applications.

### DIFF
--- a/src/ck-ldb/DistributedLB.C
+++ b/src/ck-ldb/DistributedLB.C
@@ -459,7 +459,7 @@ void DistributedLB::MapObjsToPe(minHeap *objs, CkVec<int> &obj_no,
       if (rand_pe == -1) break;
       p_id = pe_no[rand_pe];
       p_load = loads[rand_pe];
-      if (p_load + obj->load < transfer_threshold) {
+      if ((p_load + obj->load < transfer_threshold) && (CkMyPe() != p_id)) {
         success = true;
       }
       kMaxTrials--;


kavithachandrasekar:
It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.

viniciusmctf:
I believe that is exactly the issue. Previously, Grapevine was a single phase algorithm, which avoided this problem.
I detected the situation while running experiments at around 1000 cores, in cases which distributed was taking more than 5 phases to converge.
I may get the exact parameters and application with my colleague who ran the experiment, if you wish to reproduce the error.

evan-charmworks:
Could you change the commit message to follow the style described in https://chris.beams.io/posts/git-commit/, something like "Prevent DistributedLB from migrating objects to itself"?

epmikida:
> It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.

The gossip is only done initially, but the objects should only be accepted if it doesn't make the PE overloaded, so this would also be a bug. I don't believe an underloaded PE should ever become overloaded unless there is a bug.

kavithachandrasekar:
> > It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.
> 
> The gossip is only done initially, but the objects should only be accepted if it doesn't make the PE overloaded, so this would also be a bug. I don't believe an underloaded PE should ever become overloaded unless there is a bug.

Yes, that seems right, the objects would be accepted only if the PE is overloaded. But in the future phases, it seems like the target threshold is reduced and this could make a previously underloaded PE to be overloaded in the new phase?

epmikida:
> > > It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.
> > 
> > 
> > The gossip is only done initially, but the objects should only be accepted if it doesn't make the PE overloaded, so this would also be a bug. I don't believe an underloaded PE should ever become overloaded unless there is a bug.
> 
> Yes, that seems right, the objects would be accepted only if the PE is not overloaded. But in the future phases, it seems like the target threshold is reduced and this could make a previously underloaded PE to be overloaded in the new phase?

If I remember right, its just the transfer threshold that is lowered, not the acceptance threshold. If not, then that is the actual source of the bug and that's what should be fixed.

kavithachandrasekar:
> > > > It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.
> > > 
> > > 
> > > The gossip is only done initially, but the objects should only be accepted if it doesn't make the PE overloaded, so this would also be a bug. I don't believe an underloaded PE should ever become overloaded unless there is a bug.
> > 
> > 
> > Yes, that seems right, the objects would be accepted only if the PE is not overloaded. But in the future phases, it seems like the target threshold is reduced and this could make a previously underloaded PE to be overloaded in the new phase?
> 
> If I remember right, its just the transfer threshold that is lowered, not the acceptance threshold. If not, then that is the actual source of the bug and that's what should be fixed.

Sorry, I meant the transfer threshold. The `MapObjsToPe` method in the commit compares `my_load > transfer_threshold`. Wouldn't lowering the transfer threshold in a new phase make a previously underloaded PE to be overloaded?

epmikida:
> > > > > It might be good to check if an underloaded PE that becomes overloaded after a few phases can send itself an object by error. It seems like gossip of load information about underloaded PEs is only done initially and the same load information may be used over multiple phases of load transfer.
> > > > 
> > > > 
> > > > The gossip is only done initially, but the objects should only be accepted if it doesn't make the PE overloaded, so this would also be a bug. I don't believe an underloaded PE should ever become overloaded unless there is a bug.
> > > 
> > > 
> > > Yes, that seems right, the objects would be accepted only if the PE is not overloaded. But in the future phases, it seems like the target threshold is reduced and this could make a previously underloaded PE to be overloaded in the new phase?
> > 
> > 
> > If I remember right, its just the transfer threshold that is lowered, not the acceptance threshold. If not, then that is the actual source of the bug and that's what should be fixed.
> 
> Sorry, I meant the transfer threshold. The `MapObjsToPe` method in the commit compares `my_load > transfer_threshold`. Wouldn't lowering the transfer threshold in a new phase make a previously underloaded PE to be overloaded?

Ah yes, it looks like this is the bug. The transfer threshold is lowered each round, and the algorithm terminates when its lowered below the kTargetRatio * avg, which is also what defines a PE as underloaded. However, the transfer threshold is not constrained by it, so in the final phase of the algorithm, it could be lowered below the target ratio, at which point underloaded PEs would try to transfer their load as well. Looks like the offending code is in the line 339-351 range. Should probably set the new threshold to max of what we are setting it to now and kTargetRatio * avg, to ensure we never try to transfer from underloaded.

viniciusmctf:
I also believe this bug is related to a previous modification of DistBaseLB.C.
The ProcessMigrationDecision() function lacks an else case which could treat this exception. Since it has a decision tree (lines 181 and 183) that may neither abort, nor attempt a migration (to_pe == me, from_pe == me), this leads to migrates_expected != migrates_completed.
If this is left to the DistributedLB algorithm implementation, future algorithms will go through a much rougher debugging process.

Maybe a final else case in this function is also a possibility for avoiding this sort of problem?

epmikida:
I made a PR for this here: #2638, @viniciusmctf could you see if that fixes the issues you were seeing?

viniciusmctf:
Unfortunately, we lost access to the machine in scale to perform the experiment, but the issue was indeed self migration. I will look up the parameters and application and reply to this thread tomorrow with the information you'll need to perform it, if possible.

Regarding my concern on DistBase, should I open a new issue, or you believe this is still in the scope of this one?

epmikida:
> Unfortunately, we lost access to the machine in scale to perform the experiment, but the issue was indeed self migration. I will look up the parameters and application and reply to this thread tomorrow with the information you'll need to perform it, if possible.
> 
> Regarding my concern on DistBase, should I open a new issue, or you believe this is still in the scope of this one?

I think it's probably still in scope. I'll probably make a PR to fix it either tonight or tomorrow.

epmikida:
> > Unfortunately, we lost access to the machine in scale to perform the experiment, but the issue was indeed self migration. I will look up the parameters and application and reply to this thread tomorrow with the information you'll need to perform it, if possible.
> > Regarding my concern on DistBase, should I open a new issue, or you believe this is still in the scope of this one?
> 
> I think it's probably still in scope. I'll probably make a PR to fix it either tonight or tomorrow.

PR #2639 created for this. The CentralLB part will be lost when moving to the new LB framework, but the basic idea should be brought into the new framework. The new framework doesn't affect DistBaseLB so that part of the PR should be fine moving forward.

PR #2633, opened 2019-12-02
Cleanup: Make mpi-darwin-x86_64/conv-mach.sh use common/conv-mach-darwin.sh

evan-charmworks:
This fixes building mpi-darwin-x86_64, which fails due to not seeing -stdlib=libc++.


evan-charmworks:
Should this be cherry-picked onto 6.10, since it fixes a broken target?

PR #2621, opened 2019-11-25
charmc #2617: pass through rsp files

matthiasdiener:
Partially fixes #2617.

PR #2637, opened 2019-12-04
Fix GROMACS with AMPI

evan-charmworks:
This plus the `charmc-response-files` branch from #2621 should allow GROMACS to build and run with AMPI.

@marmistrz You'll need to add `#include "mpi.h"` to src/programs/gmx.cpp so we properly capture the entry point. If this step is forgotten at least AMPI will now warn about it instead of silently segfaulting.


marmistrz:
So the missing `mpi.h` is a bug in Gromacs, isn't it?

marmistrz:
I'm confirming that GROMACS builds with the following patch after rebasing on top of the latest master. Unfortunately, the performance under AMPI is horrible compared with pure OpenMPI (only single node tested so far). A workload, which has a runtime of about 5 minutes under OpenMPI is expected to run about 4 hours with AMPI. (about 50x slowdown).

AMPI was configured using: 
```
./build AMPI mpi-linux-x86_64 --with-production --build-shared -j8
```
the version of MPI used: (self build)
```
$ mpirun --version
mpirun (Open MPI) 3.1.4
```
GROMACS is executed as:
```
bin/gmx_mpi mdrun -v -deffnm tip4p_min2
```
I can share the workload with you. The performance deterioration doesn't occur if `multicore-linux-x86_64` is used.

stwhite91:
We usually advise against using the MPI networking layer in Charm++/AMPI and instead preferring the native one for your machine, such as verbs/ucx for Infiniband systems. The MPI layer means that AMPI is built on top of OpenMPI and will use it internally for all communication. Regardless, the MPI layer should not be causing a 50x slowdown, so it would be good for us to debug this case. If you could provide instructions and any data sets needed to reproduce this, we can dig in further

evan-charmworks:
> GROMACS is executed as:
> 
> ```
> bin/gmx_mpi mdrun -v -deffnm tip4p_min2
> ```
> 
> I can share the workload with you. The performance deterioration doesn't occur if `multicore-linux-x86_64` or `verbs-linux-x86_64` is used.

If you could share the workload, that would be helpful. We have seen prior cases where an significant slowdown occurs but is fixable with a straightforward change to AMPI.

marmistrz:
This is the workload I'm using, the command as in my previous post: 
[tip4p_min2.tpr.gz](https://github.com/UIUC-PPL/charm/files/3935377/tip4p_min2.tpr.gz)

Please let me know if I can help debugging in any way. For instance, I could do some profile but I'm not quite sure if using generic MPI profilers is the way to go here. (mpiP, for instance)

evan-charmworks:
> This is the workload I'm using, the command as in my previous post:
> [tip4p_min2.tpr.gz](https://github.com/UIUC-PPL/charm/files/3935377/tip4p_min2.tpr.gz)

Thanks for providing this. I can't reproduce the performance degradation with an mpi-linux-x86_64 build of AMPI. Is there anything in this program output that distinguishes my run from yours?

```
Reading file ../../tip4p_min2.tpr, VERSION 2019.3 (single precision)
Changing nstlist from 10 to 100, rlist from 1 to 1.122

Using 1 MPI process
Using 6 OpenMP threads 

1 GPU selected for this run.
Mapping of GPU IDs to the 2 GPU tasks in the 1 rank on this node:
  PP:0,PME:0
PP tasks will do (non-perturbed) short-ranged interactions on the GPU
PME tasks will do all aspects on the GPU
starting mdrun 'SOL'
200000 steps,    200.0 ps.
step  200: timed with pme grid 25 25 25, coulomb cutoff 1.000: 285.4 M-cycles
step  400: timed with pme grid 20 20 20, coulomb cutoff 1.250: 320.8 M-cycles
step  600: timed with pme grid 24 24 24, coulomb cutoff 1.042: 278.0 M-cycles
step  800: timed with pme grid 25 25 25, coulomb cutoff 1.000: 279.1 M-cycles
step 1000: timed with pme grid 24 24 24, coulomb cutoff 1.042: 276.4 M-cycles
step 1200: timed with pme grid 25 25 25, coulomb cutoff 1.000: 281.8 M-cycles
              optimal pme grid 24 24 24, coulomb cutoff 1.042
step 199900, remaining wall clock time:     0 s          
Writing final coordinates.
step 200000, remaining wall clock time:     0 s          
               Core t (s)   Wall t (s)        (%)
       Time:      937.702      156.284      600.0
                 (ns/day)    (hour/ns)
Performance:      110.569        0.217
```

marmistrz:
The first thing I've noticed: I'm not using the GPU.

evan-charmworks:
I can reproduce it by adding `-DGMX_GPU=OFF` to my cmake line.

evan-charmworks:
I tried again after changing `-DCMAKE_BUILD_TYPE=Debug` to `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and the expected time is in the range of 9 minutes.

marmistrz:
I suggest we merge this change as it is and take care of performance issues in subsequent PRs.

PR #2639, opened 2019-12-06
Prevent self migrations in LB strategies

epmikida:
Another fix related to PR #2630. If there are bugs in strategies which create migration requests those are not currently caught. This was the case in Distributed LB, an external collaborator caught the weird behavior.

Change-Id: I75cd3c67bbe3f9d7e705d1f4ca96f9c8451f4b93


evan-charmworks:
Is this ready to merge, since it adds an abort without adding a fix for the issue it detects?

epmikida:
> Is this ready to merge, since it adds an abort without adding a fix for the issue it detects?

PR #2638 fixes the issue it detects, which is just waiting on one more review but then they should both be ready to merge.

Issue #2627, opened 2019-11-27
Checkpoint file open failure with new subdirectory checkpointing


trquinn:
I occasionally get the following error with the new checkpointing scheme (pr  #2611):
```
------------- Processor 11448 Exiting: Called CmiAbort ------------
Reason: PE 11448 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_424.dat, mode: wb, status: No such file or directory
[11448] Stack Traceback:
------------- Processor 11475 Exiting: Called CmiAbort ------------
Reason: PE 11475 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_425.dat, mode: wb, status: No such file or directory
[11475] Stack Traceback:
------------- Processor 10746 Exiting: Called CmiAbort ------------
Reason: PE 10746 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_398.dat, mode: wb, status: No such file or directory
------------- Processor 11421 Exiting: Called CmiAbort ------------
Reason: PE 11421 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_423.dat, mode: wb, status: No such file or directory
[11421] Stack Traceback:
[10746] Stack Traceback:
------------- Processor 6183 Exiting: Called CmiAbort ------------
Reason: PE 6183 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub0/NodeGroups_229.dat, mode: wb, status: No such file or directory
[6183] Stack Traceback:
------------- Processor 10773 Exiting: Called CmiAbort ------------
Reason: PE 10773 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_399.dat, mode: wb, status: No such file or directory
[10773] Stack Traceback:
------------- Processor 11394 Exiting: Called CmiAbort ------------
Reason: PE 11394 failed to open checkpoint file: h148.cosmo50PLK.6144g3HbwK1BH.chk0/sub1/NodeGroups_422.dat, mode: wb, status: No such file or directory
------------- Processor 6156 Exiting: Called CmiAbort ------------
[...]

PR #2641, opened 2019-12-07
ckCheckpoint: create subdirectories for nodegroup files, if needed.

trquinn:
Here is my proposed fix to issue #2627.  The nodegroup file goes into a different directory than the group or array files, so make sure that directory is created if needed.

### DIFF
--- a/src/ck-core/ckcheckpoint.C
+++ b/src/ck-core/ckcheckpoint.C
@@ -230,25 +230,27 @@ void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _reque
 	}
 
 	// Due to file system issues we have observed, divide checkpoints
-	// into subdirectories to avoid having too many files in a single directory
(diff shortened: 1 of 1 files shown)


trquinn:
This wasn't quite ready to be merged.  Could someone check the use of the rdbuf() method on line 239?

minitu:
@trquinn Just did a small test:

```
#include <string>
#include <sstream>
#include <iostream>

int main() {
  std::ostringstream x, y;
  x << "hello";
  y << x.rdbuf();
  std::cout << y.str() << std::endl;
}
```

This doesn't work as intended, because nothing gets printed out.
I think `x.rdbuf()` should be changed to `x.str()`, so in our case, `dirPathNode << dirPath.str()`.

trquinn:
I concur.  I just tried with dirPath.str(), and things are working.  I'll do another pull request.

PR #2642, opened 2019-12-09
HAPI: Add +gpumap and +allgpus options for PE-GPU mapping

minitu:
Update PE-GPU mapping in Charm++/HAPI initialization to **1)** support block and round-robin mappings (block is default) and **2)** work properly on both `jsrun` and non-`jsrun` machines.

Machines that use `jsrun` can restrict the number of GPUs visible to each process, where the device indices start from 0 for each process. For example, on a machine with 4 GPUs per node (2 per socket) such as LC Lassen, `jsrun -n2 -a1 -c8 -g2 -K1 -r2` and `+ppn 8` will spawn 2 processes (1 per socket) on a single node, with each process (logical node) containing 8 PEs/worker threads. GPUs with global indices of 0-3 will be split into 2 sets each with 2 GPUs, where the first set is visible to process 0 (PEs 0-7) with indices of 0-1, and the second is visible to process 1 (PEs 8-15) also with indices of 0-1 (instead of 2-3).

Because the current multi-GPU machines that we are interested in use `jsrun` (e.g. OLCF Summit, LC Lassen), the GPU resource allocation provided by `jsrun` is regarded as the default. On other machines where all GPUs are visible to all processes (e.g. PSC Bridges) the user can pass `+allgpus` to override the default.

Currently two options are supported for PE-GPU mapping: **1) blocking** and **2) round-robin**. The user should pass `+gpumap block` and `+gpumap roundrobin`, respectively. Custom PE-GPU mappings similar to those possible with `+pemap` should be supported but is left as future work.


minitu:
This patch is required for direct GPU-GPU messaging, which builds on top of the Zero Copy Entry Method Post API.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -771,10 +762,60 @@ void initHybridAPI() {
...
 }
 
-// Set HAPI device for non-0 ranks
-void setHybridAPIDevice() {
-  // set which device to use
-  hapiCheck(cudaSetDevice(getMyCudaDevice(CmiMyNode())));
+// Set up PE to GPU mapping, invoked from all PEs
+// TODO: Support custom mappings
+void initDeviceMapping(char** argv) {
+  int map_type = 0; // 0: block, 1: round-robin (default is block mapping)
```
matthiasdiener: A nicer way might be to make this into a `class enum` (especially if there are more mapping in the future), but not sure if it is worth it here.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -771,10 +762,78 @@ void initHybridAPI() {
...
+      map_type = Mapping::None;
+    }
+    else if (strcmp(gpumap, "block") == 0) {
+      map_type = Mapping::Block;
+    }
+    else if (strcmp(gpumap, "roundrobin") == 0) {
+      map_type = Mapping::RoundRobin;
+    }
+    else {
+      CmiAbort("Unsupported mapping type!");
```
matthiasdiener: Here you could mention the mapping type the user chose, and which types are actually available.

PR #2644, opened 2019-12-09
Enable within-node direct GPU-GPU messaging

minitu:
This PR depends on #2642 (changes to HAPI and `CkInit` code) to ensure proper mapping of PEs to GPUs.

Building on the Zero Copy Entry Method Post API, this PR enables direct GPU-GPU messaging (bypassing the host memory) between chares on the same node.

Based on MVAPICH2's implementation, we use peer-to-peer memcpys to transfer data between GPUs accessible by PEs in the same process/logical node, and CUDA Inter-Process Communication (IPC) for GPUs accessed by different processes on the same node.

A separate patch that uses GPUDirect RDMA is needed to enable direct GPU-GPU communication between chares on different physical nodes.

References:
- MVAPICH2-GPU: optimized GPU to GPU communication for InfiniBand clusters: https://link.springer.com/article/10.1007/s00450-011-0171-3
- Optimizing MPI Communication on Multi-GPU Systems Using CUDA Inter-Process Communication: https://ieeexplore.ieee.org/abstract/document/6270863/
- Efficient Inter-node MPI Communication using GPUDirect RDMA for InfiniBand
Clusters with NVIDIA GPUs: https://ieeexplore.ieee.org/abstract/document/6687341/


examples/charm++/cuda/zerocopy/zerocopy.C
```diff
@@ -0,0 +1,131 @@
...
+
+    // Initialize data
+    invokeInitKernel(d_local_data, block_size, (double)thisIndex, stream);
+
+    // Start iterating once data is initialized
+    CkCallback* cb = new CkCallback(CkIndex_Block::iterate(), thisProxy[thisIndex]);
+    hapiAddCallback(stream, cb);
+  }
+
+  void receive(int ref, int &size, double *&arr, CkNcpyBufferPost *ncpyPost) {
```
nitbhat: This receive method is the Post Entry Method. You would need another regular entry method overloaded (of the same name `receive`). Control inside that method signifies the completion of the Rget transfer. I'm curious as to how the compiler doesn't complain about that entry method. Which machine did you compile this on and what was your build command?

minitu: The second `receive` method is defined in the `.ci` file.

nitbhat: Ah okay, didn't realize it was an sdag entry method.

minitu:
Also depends on #2659.

minitu:
Single-process messaging works now (albeit synchronously, using cudaMemcpy). Asynchronous support with cudaMemcpyAsync should be added, which requires adding a CUDA stream parameter.

One possible bug that I encountered with the existing zero-copy implementation is that there is no handling of the (reused) message enqueued in `CkRdmaIssueRgets` with the header `CMK_ZC_P2P_RECV_DONE_MSG`. This causes duplicate calls of the receiver-side zero-copy API entry methods.

nitbhat:
> Single-process messaging works now (albeit synchronously, using `cudaMemcpy`). Asynchronous support with `cudaMemcpyAsync` should be added, which requires adding a CUDA stream parameter.
> 
> One possible bug that I encountered with the existing zero-copy implementation is that there is no handling of the (reused) message enqueued in `CkRdmaIssueRgets` with the header `CMK_ZC_P2P_RECV_DONE_MSG`. This causes duplicate calls of the receiver-side zero-copy API entry methods.

Hmm, that’s weird. enqueueNcpyMessage is the only way for CMK_ZC_P2P_RECV_DONE_MSG to be enqueued. It should get handled again by the charm level 2 handler i.e. _processHandler.
Do you know the code path for it to be called twice? If not, you can you put a breakpoint in `_processHandler`, potentially with an `if(CMI_ZC_MSGTYPE(env) == CMK_ZC_P2P_RECV_DONE_MSG` and provide the stack trace for the two times it gets called?

minitu:
Now also depends on #2667 for proper execution on `CMK_ONESIDED_IMPL` enabled archs such as `mpi-linux-ppc64le`.

minitu:
Pingpong on OLCF Summit
Building Charm++:
./build charm++ pamilrts-linux-ppc64le smp cuda -j --with-production
Testing:
cd charm/examples/charm++/cuda/zerocopy-pingpong
make
Intra-process: jsrun -n1 -a1 -c2 -g2 ./zerocopy -s 1 -x 131072 -i 100 -z ++ppn 2 +pemap L0-28:4 +gpumap block +gpuipceventpool 2048 +gpucommbuffer 1073741824 (on a batch node)
Inter-process: jsrun -n2 -a1 -c1 -g1 ./zerocopy -s 1 -x 131072 -i 100 -z ++ppn 2 +pemap L0-28:4 +gpumap block +gpuipceventpool 2048 +gpucommbuffer 1073741824

+gpumap specifies block mapping of PEs to GPUs, +gpuipceventpool specifies the number of CUDA IPC events to be created per PE (used for inter-process communication), and +gpucommbuffer specifies the size of the device communication buffer per device (1GB in this case, also used for inter-process comm.)

minitu:
Implemented mechanism (_not_ a message-based ACK mechanism) based on CUDA IPC Event to free allocated block of device communication buffer and used IPC events on the sender. Now the pingpong benchmark should work for all sizes and however many number of iterations.

**TODOs**
- Improve performance, especially for small message sizes. First thing is to speed up the process on the sender, and one thing that can be easily done is to not search and free complete IPC events and corresponding block on device comm buffer for every zero-copy message. It can only be done when either the CUDA IPC event pool or device comm buffer is somewhat used up (more than half?).
- Check if it works for more than 2 processes
- Check if there can be race conditions regarding the use of CUDA IPC events on two processes.
- How to let the user know that the source buffer can be reused? Internally this is known when the first recorded CUDA IPC event is complete.

src/xlat-i/xi-Entry.C
```diff
@@ -1870,21 +1914,51 @@ void Entry::genRegularCall(XStr& str, const XStr& preCall, bool redn_wrapper, bo
       if(isRdmaPost) {
         str << "#if CMK_ONESIDED_IMPL\n";
         // Allocate an array of rdma pointers
-        str << "  void *buffPtrs["<< numRdmaRecvParams <<"];\n";
-        str << "  int buffSizes["<< numRdmaRecvParams <<"];\n";
+        if (param->hasDevice()) {
+          str << "  void *buffPtrs["<< numRdmaRecvParams + numRdmaDeviceParams <<"];\n";
+          str << "  int buffSizes["<< numRdmaRecvParams + numRdmaDeviceParams <<"];\n";
+          str << "#else\n";
+          str << "  void *buffPtrs["<< numRdmaDeviceParams <<"];\n";
```
nitbhat: What's the plan for supporting ZC Post Device API on non-RDMA layers? (like netlrts). Is that why this code was added? (the two lines after `#else`). I am guessing that you can always support Cuda IPC transfers as long as you have GPU, but support for GPU direct is limited to only a few networks?

src/xlat-i/xi-Parameter.C
```diff
@@ -690,12 +837,9 @@ void ParamList::beginUnmarshallSDAGCall(XStr& str, bool usesImplBuf) {
         << ";\n";
     if (hasRdma()) {
       if(hasRecvRdma()) {
-        str << "  CkNcpyBufferPost ncpyPost[" << entry->numRdmaRecvParams << "];\n";
-        for(int index=0; index < entry->numRdmaRecvParams; index++) {
```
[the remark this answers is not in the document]
minitu: I moved the code to initialize `regMode` and `deregMode` to the default constructor of `CkNcpyBufferPost`, in ckrdma.h.

src/xlat-i/xi-Parameter.C
```diff
@@ -508,20 +613,20 @@ void ParamList::beginUnmarshall(XStr& str) {
...
-      str << "  int impl_num_rdma_fields; implP|impl_num_rdma_fields; \n";
+      str << "  int impl_num_rdma_fields; implP|impl_num_rdma_fields;\n";
       str << "  int impl_num_root_node; implP|impl_num_root_node;\n";
       callEach(&Parameter::beginUnmarshallRdma, str, true);
       str << "#else\n";
       callEach(&Parameter::beginUnmarshallRdma, str, false);
       str << "#endif\n";
       if (hasRecvRdma()) {
-        str << "  CkNcpyBufferPost ncpyPost[" << entry->numRdmaRecvParams << "];\n";
-        for(int index=0; index < entry->numRdmaRecvParams; index++) {
```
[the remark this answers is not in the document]
minitu: I moved the code to initialize `regMode` and `deregMode` to the default constructor of `CkNcpyBufferPost`, in ckrdma.h.

src/xlat-i/xi-Entry.C
minitu: I think the device-side ZC API should be supported also for non-RDMA layers (i.e. `CMK_ONESIDED_IMPL` disabled), which is why there have been some changes to accomodate this in the code. Adding in GPUDirect RDMA for inter-node GPU messaging will make things more complicated, because the architectures that support it may not match the ones that support host-side RDMA. I'm leaving that for the future me to mull over, for now.

minitu:
This patch is now mostly ready to go into release 6.11, passes both intra-node and inter-node tests with `verify`, `jacobi2d`, and `pingpong` (inter-node uses basic host-staged mechanism). I've looked into supporting `CkCallback` in `hapiAddCallback`, but it seems to require signifcant changes (the problem is allowing `CkCallback` in `hapi.h` - this would make it a Charm++ level header file instead of a Converse header file as it is currently).

minitu:
Will be split into smaller PRs, with additional changes from the inter-node patch (#2791).

PR #2645, opened 2019-12-11
Checkpoint: use str() instead of rdbuf() to copy dirPath.

trquinn:
ostreamstring::rdbuf() does not do the obvious thing.  Use str() instead.


minitu:
Fixes #2641.

PR #2650, opened 2019-12-12
Fix PUP routine for vector<bool>

stwhite91:


evan-charmworks:
Would it also work to remove the `std::vector<bool>` specialization entirely? I don't see anything this does that isn't also achieved by the template method.

stwhite91:
> Would it also work to remove the `std::vector<bool>` specialization entirely? I don't see anything this does that isn't also achieved by the template method.

Without it the generic one tries to use 'emplace', which isn't defined for vector<bool> until C++14

evan-charmworks:
> > Would it also work to remove the `std::vector<bool>` specialization entirely? I don't see anything this does that isn't also achieved by the template method.
> 
> Without it the generic one tries to use 'emplace', which isn't defined for vector until C++14

I was thinking that the

```cpp
  template <class dtype, class... Args>
  inline void emplace(std::vector<dtype> &c, Args&&... args)
  {
    c.emplace_back(std::forward<Args>(args)...);
  }
```

should take care of that. I tried removing the declaration and definition of `PUP_stl_container_items<std::vector<bool>,bool>` to test and I get:

```
pup_stl.h: In instantiation of ‘void PUP::PUP_stl_container_items(PUP::er&, container&, size_t) [with container = std::vector<bool>; dtype = bool; size_t = long unsigned int]’:
pup_stl.h:239:46:   required from ‘void PUP::PUP_stl_container(PUP::er&, container&) [with container = std::vector<bool>; dtype = bool]’
pup_stl.h:363:52:   required from here
pup_stl.h:230:21: error: taking address of temporary [-fpermissive]
         p|*(dtype *)&(*it);
                     ^~~~~~
```

So yeah, it looks like a dedicated overload is necessary.

> If this else clause is correct, we should put a comment in the code regarding the why the asymmetry between packing and unpacking is necessary.

The above is probably the reason for the asymmetry. Unpacking to a temporary won't work.

evan-charmworks:
After looking at it again the asymmetry is because the packing path constructs a `std::deque<bool>` from the vector and serializes that. The original bugged implementation did that for unpacking too without putting the data back into the vector.

PR #2652, opened 2019-12-12
HAPI: Implement CUDA Event pool

### LINKED ISSUE #2069: Create pool of CUDA Events
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2069*

---
CUDA Event creation could add overhead if it is done dynamically.
It may be better to create a pool of events, similar to how streams and mempool gets created up front.

minitu:
Closes #2069.
Based on `jchoi/gpu-overdecomp branch`.


minitu:
Also depends on `jchoi/hapi-gpumap` branch.

minitu:
I see about a 10% performance improvement in the concurrent kernel execution part of vector addition (1M elements):
64-way split: 1.94 ms -> 1.78 ms
256-way split: 7.68 ms -> 6.99 ms

PR #2653, opened 2019-12-13
Fix missing CmiInCommThread

matthiasdiener:
Followup from #2540.

### DIFF
--- a/src/conv-core/converse.h
+++ b/src/conv-core/converse.h
@@ -562,6 +562,8 @@ extern CmiNodeLock CmiMemLock_lock;
 
 #if CMK_SMP
 #define CmiInCommThread()  (CmiMyRank() == CmiMyNodeSize())
+#else
+#define CmiInCommThread()  (0)
 #endif
 
 #endif /* CMK_SHARED_VARS_NT_THREADS */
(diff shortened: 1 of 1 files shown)


evan-charmworks:
Should there be one for `CMK_SHARED_VARS_UNIPROCESSOR` too?

evan-charmworks:
Additionally, this patch should `#define CmiInCommThread() (0)` if not SMP

matthiasdiener:
> Additionally, this patch should `#define CmiInCommThread() (0)` if not SMP

Done.

matthiasdiener:
> Should there be one for `CMK_SHARED_VARS_UNIPROCESSOR` too?

Not sure. What should CmiInCommThread() be defined to in this case?

matthiasdiener:
> From this, my guess is that we should just define `#define CmiInCommThread() (0)`?:
> 
> ```c
> #if CMK_SHARED_VARS_UNIPROCESSOR /*Used only by uth- and sim- versions*/
> 
> #define CmiMyPe()              _Cmi_mype
> #define CmiMyRank()            _Cmi_mype
> ```

done

PR #2658, opened 2019-12-17
Cleanup: Replace CmmTable with std::map in CPU affinity and topology

evan-charmworks:


src/conv-core/cputopology.C
```diff
@@ -227,26 +231,25 @@ static void cpuTopoHandler(void *m)
...
+      hostTable.emplace(ip, proc);
     }
     topomsg->nodes[pe] = rec->nodeID;
     rec->rank ++;
   }
 
-  printTopology(CmmEntries(hostTable));
+  printTopology(hostTable.size());
 
     // clean up CmmTable
```
matthiasdiener: comment should be adjusted

src/conv-core/cpuaffinity.C
```diff
@@ -415,8 +418,13 @@ typedef struct _affMsg {
 #endif
 } affMsg;
 
+static inline bool operator< (const skt_ip_t & a, const skt_ip_t & b)
```
matthiasdiener: Instead of adding the function twice, could you put it into (e.g.) sockRoutines.h?

src/conv-core/cpuaffinity.C
```diff
@@ -14,6 +14,9 @@ cpu affinity.
 #define _GNU_SOURCE
 #endif
 
+#include <map>
+#include <algorithm>
```
matthiasdiener: Is this include actually used?

src/conv-core/cpuaffinity.C
```diff
@@ -415,8 +418,13 @@ typedef struct _affMsg {
...
 } affMsg;
 
+static inline bool operator< (const skt_ip_t & a, const skt_ip_t & b)
+{
+  return memcmp(&a, &b, sizeof(a)) < 0;
+}
+
 static rankMsg *rankmsg = NULL;
-static CmmTable hostTable;
+static std::map<skt_ip_t, hostnameMsg *> hostTable;
```
matthiasdiener: Could this also be an unorderd_map?

src/conv-core/cputopology.C
evan-charmworks: Done

src/conv-core/cpuaffinity.C
evan-charmworks: I could, but I would need to add `#include <cstring>` (for memcmp) to that file. Would that be okay?

src/conv-core/cpuaffinity.C
evan-charmworks: <algorithm> for the std::swap I added

src/conv-core/cpuaffinity.C
evan-charmworks: Both could, but I would need to figure out how to implement the necessary hash function.

src/conv-core/cpuaffinity.C
matthiasdiener: Imho, that would be ok.

src/conv-core/cpuaffinity.C
matthiasdiener: Ok, might not be necessary then.

src/conv-core/cpuaffinity.C
evan-charmworks: Done

PR #2659, opened 2019-12-17
Support logical PU indices in +pemap and +commap

### LINKED ISSUE #2657: Support logical PU indices in +pemap
Related to #2265 .

Currently `+pemap` only recognizes PU indices assigned by the OS.
For machines where the user does not know how the OS maps the indices or some PUs are reserved for OS use (e.g. LLNL Lassen with `-core_isolation`), being able to provide logical PU indices would be useful.
Also, hwloc recommends the use of logical indices for portability (see page 65 of their [tutorial slides](https://www.open-mpi.org/projects/hwloc/tutorials/20150921-EuroMPI-hwloc-tutorial.pdf)).

minitu:
Closes #2657.

The user can pass options like `+pemap L0-7` to use logical PU indices instead of indices assigned by the OS.
Also updates program output and error messages to make it clear which version (OS indices vs. logical indices) is being used.
Tested on LLNL Lassen.

### DIFF
--- a/doc/charm++/manual.rst
+++ b/doc/charm++/manual.rst
@@ -11541,8 +11541,10 @@ are available:
    ``0-15:4.3``. ``++ppn 10 +pemap 0-11:6.5+12`` equals
    ``++ppn 10 +pemap 0,12,1,13,2,14,3,15,4,16,6,18,7,19,8,20,9,21,10,22``
 
-   By default, this option accepts PU indices assigned by the OS. To
-   provide logical PU indices, prepend the sequence with an alphabet L
+   By default, this option accepts PU indices assigned by the OS. The
+   user might want to instead provide logical PU indices used by hwloc
+   (see `link <https://www.open-mpi.org/projects/hwloc/doc/v2.1.0/a00342.php#faq_indexes>`
+   for details). To do this, prepend the sequence with an alphabet L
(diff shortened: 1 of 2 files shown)


doc/charm++/manual.rst
```diff
@@ -11541,6 +11541,11 @@ are available:
    ``0-15:4.3``. ``++ppn 10 +pemap 0-11:6.5+12`` equals
    ``++ppn 10 +pemap 0,12,1,13,2,14,3,15,4,16,6,18,7,19,8,20,9,21,10,22``
 
+   By default, this option accepts PU indices assigned by the OS. To
+   provide logical PU indices, prepend the sequence with an alphabet L
+   (case-insensitive). For instance, ``+pemap L0-3`` will instruct the
+   runtime to bind threads to PUs with logical indices 0-3.
```
evan-charmworks: Should mention that the logical indices come from hwloc, otherwise there is no context for what logical means.

Could link to https://www.open-mpi.org/projects/hwloc/doc/v2.1.0/a00342.php#faq_indexes

jcphill:
This needs to apply the same modification to +commap, and should should also issue a warning if both +pemap and +commap are specified but only one uses logical indices, as this is almost certainly a user error.

PR #2660, opened 2019-12-18
doc-readme: clarify sphinx version

matthiasdiener:


matthiasdiener:
This PR is incorrect, the dependency is actually on pygments 2.4.0 (not Sphinx), and I'm not sure if it is worth adding this to the manual.

PR #2661, opened 2019-12-18
Generate hwloc's configure script and keep it in the repository

evan-charmworks:
This removes the CMake build system's dependency on autotools.


evan-charmworks:
> This removes the CMake build system's dependency on autotools.

Maybe it doesn't...

```
[ 11%] Performing build step for 'hwloc'
make[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
 cd /home/runner/work/charm/charm/netlrts-linux-x86_64/hwloc-prefix/src/hwloc && /bin/bash /home/runner/work/charm/charm/netlrts-linux-x86_64/hwloc-prefix/src/hwloc/config/missing automake-1.16 --foreign
/home/runner/work/charm/charm/netlrts-linux-x86_64/hwloc-prefix/src/hwloc/config/missing: line 81: automake-1.16: command not found
WARNING: 'automake-1.16' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
Makefile:526: recipe for target '/home/runner/work/charm/charm/netlrts-linux-x86_64/hwloc-prefix/src/hwloc/Makefile.in' failed
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>
make[3]: *** [/home/runner/work/charm/charm/netlrts-linux-x86_64/hwloc-prefix/src/hwloc/Makefile.in] Error 1
make[2]: *** [hwloc-prefix/src/hwloc-stamp/hwloc-build] Error 2
make[1]: *** [CMakeFiles/hwloc.dir/all] Error 2
CMakeFiles/hwloc.dir/build.make:112: recipe for target 'hwloc-prefix/src/hwloc-stamp/hwloc-build' failed
CMakeFiles/Makefile2:184: recipe for target 'CMakeFiles/hwloc.dir/all' failed
```

Did I miss any details in CMakeLists.txt?

evan-charmworks:
Looks liked I fixed the above problem.

matthiasdiener:
See also #2142 - is the command mentioned in item 2. in update-hwloc.sh the only thing necessary to refresh this?

evan-charmworks:
This PR only generates hwloc's configure script, while #2142 generates Charm's. It should only change when updating hwloc.

evan-charmworks:
Confirmed that `./buildcmake` completes successfully with this patch by uninstalling autoconf and automake locally.

evan-charmworks:
This PR removes the need to have autotools installed to use the CMake build system. The only maintenance cost associated with this PR takes place during the process of updating our embedded hwloc version, and most of that is taken care of by the `update-hwloc.sh` script. It also sets a better precedent for any future external projects we add to `contrib/`.

matthiasdiener:
> This PR removes the need to have autotools installed to use the CMake build system. The only maintenance cost associated with this PR takes place during the process of updating our embedded hwloc version, and most of that is taken care of by the `update-hwloc.sh` script. It also sets a better precedent for any future external projects we add to `contrib/`.

+1 in general. Out of curiosity, what is the reason of not adding an unpacked release tarball of hwloc, instead of the development version?

evan-charmworks:
> Out of curiosity, what is the reason of not adding an unpacked release tarball of hwloc, instead of the development version?

The hwloc update script takes a release tarball as its input.

```
$ ./update-hwloc.sh
./update-hwloc.sh <hwloc-X.Y.Z.tar.(gz|bz2)>
```

matthiasdiener:
> > Out of curiosity, what is the reason of not adding an unpacked release tarball of hwloc, instead of the development version?
> 
> The hwloc update script takes a release tarball as its input.
> 
> ```
> $ ./update-hwloc.sh
> ./update-hwloc.sh <hwloc-X.Y.Z.tar.(gz|bz2)>
> ```

Sure, but why not add the release tarball directly?

evan-charmworks:
> > > Out of curiosity, what is the reason of not adding an unpacked release tarball of hwloc, instead of the development version?
> > 
> > 
> > The hwloc update script takes a release tarball as its input.
> > ```
> > $ ./update-hwloc.sh
> > ./update-hwloc.sh <hwloc-X.Y.Z.tar.(gz|bz2)>
> > ```
> 
> Sure, but why not add the release tarball directly? (in other words, wouldn't the update-hwloc script be unnecessary then?)

I assume you mean committing the extracted tarball contents directly and not the .tar.* file itself, which would be bad for the size and compressibility of the git repository. The script currently serves to get the directory structure exactly right, as well as thin out the contents of the tarball so the git history and users' drives are less impacted by things hwloc includes that are irrelevant to an embedded package. Additionally, running autotools to regenerate the configure scripts is the best way to handle our disabling of the C99 check (#2614).

PR #2649, opened 2019-12-12
Bug #2647: Revert use of logical CPU indices in cpuAffinityRecvHandler

evan-charmworks:
PR #2448 caused a performance regression for existing users of
+setcpuaffinity and redefined the meaning of values passed to the
+excludecore parameter.

On machines such as Bridges where +setcpuaffinity crashes after this
reversion (#2445), try also passing +excludecore 0.

PR #2664, opened 2019-12-28
Bug #2647: Improve +setcpuaffinity defaults

### LINKED ISSUE #2647: Inefficient +setcpuaffinity defaults
PR #2448 changed `cpuAffinityRecvHandler` to use logical indices when assigning CPU affinity instead of non-portable OS indices in response to Issue #2445, where the latter causes an abort with `+setcpuaffinity` on Bridges. This new numbering scheme may result in poor performance in some cases due to grouping threads as closely as possible on adjacent SMTs while leaving entire cores unused. The example reported is with `+p12` on a 2 sockets x 8 cores x 2 PUs = 32-way SMP machine.

evan-charmworks:
PR #2448 caused a performance regression for existing users of +setcpuaffinity and redefined the meaning of values passed to the +excludecore parameter. This PR resolves both issues by improving the affinity assignment algorithm.

Closes #2647, #2649


src/conv-core/cpuaffinity.C
```diff
@@ -442,62 +448,149 @@ static void cpuAffinityHandler(void *m)
...
   }
-  myrank = rec->rank%rec->ncores;
-  while (in_exclude(myrank)) {             /* skip excluded core */
-    myrank = (myrank+1)%rec->ncores;
-    rec->rank ++;
+
+  int & node_host = rankmsg->node_hosts[node];
+  if (node_host == -1)
+  {
+    rankmsg->node_ranks[node] = rec->node_rank++;
```
minitu: Shouldn't `rec->node_rank` also be incremented each time a different PE of the same logical node is processed? With the current code, it is only incremented when the first PE of each logical node is processed.

src/conv-core/cpuaffinity.C
```diff
@@ -442,62 +448,149 @@ static void cpuAffinityHandler(void *m)
...
+  int pe_rank = m->pe_ranks[mype];
+  int myhost = m->pe_hosts[mype];
+
+  int nodesOnHost = 0;
+  for (int i = 0; i < numnodes; ++i)
+    if (m->node_hosts[i] == myhost)
+      ++nodesOnHost;
+
+  hwloc_obj_t obj;
+  const int package_count = cmi_hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_PACKAGE);
```
minitu: Some comments on what this part of the code is doing would be nice.

src/conv-core/cpuaffinity.C
```diff
@@ -400,18 +402,22 @@ static int cpuPhyNodeAffinityRecvHandlerIdx;
...
 typedef struct _rankMsg {
   char core[CmiMsgHeaderSizeBytes];
-  int *ranks;                  /* PE => core rank mapping */
-  int *nodes;                  /* PE => node number mapping */
+  int *node_ranks;                /* node => core rank mapping */
+  int *node_hosts;                /* node => host number mapping */
+  int *pe_ranks;                  /* PE => core rank mapping */
+  int *pe_hosts;                  /* PE => host number mapping */
 } rankMsg;
+#define RANKMSG_SIZE(numnodes, numpes) (sizeof(rankMsg) + (numnodes)*sizeof(int)*2 + (numpes)*sizeof(int)*2)
```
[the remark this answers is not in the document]
evan-charmworks: Done, I've added a static method that does this.

src/conv-core/cpuaffinity.C
evan-charmworks: No, the variable refers to the rank of the node on the host. I've updated its name to `node_ranks_on_host` to make this clear.

src/conv-core/cpuaffinity.C
evan-charmworks: Added.

minitu:
**Test on Lassen**
- Single host
- 2 sockets
- 20 CPU cores per socket, with each core up to 4 SMT threads (total of 160 PUs available per host)
- Test with 2 processes per socket (4 processes total)

**1. SMT1: 10 threads per process**

`jsrun -n4 -a1 -c10 -K2 -r4 ./hello +ppn 10 +setcpuaffinity`

PEs 0-9 (process 0) -> logical PUs 0-36:4 (socket 0)
PEs 10-19 (process 1) -> logical PUs 80-116:4 (socket 1)
PEs 20-29 (process 2) -> logical PUs 40-76:4 (socket 0)
PEs 30-39 (process 3) -> logical PUs 120-156:4 (socket 1)

**2. SMT4: 40 threads per process**

`jsrun -n4 -a1 -c10 -K2 -r4 ./hello +ppn 40 +setcpuaffinity`

PEs 0-39 (process 0) -> logical PUs 0-39 (socket 0)
PEs 40-79 (process 1) -> logical PUs 80-119 (socket 1)
PEs 80-119 (process 2) -> logical PUs 40-79 (socket 0)
PEs 120-159 (process 3) -> logical PUs 120-159 (socket 1)

The mapping seems to work correctly.

minitu:
**Test on Bridges**
- Single host
- 2 sockets
- 14 CPU cores per socket, no SMT (total of 28 PUs available per host)
- 1 comm thread per process
- Test with 2 processes per socket (4 processes total)

`mpirun -np 4 ./hello +ppn 6 +setcpuaffinity`
(6 worker threads + 1 comm thread) x 4 processes = 28 threads mapped to 28 PUs


PEs 0-5 (process 0) -> logical PUs 0-5 (socket 0)
PEs 6-11 (process 1) -> logical PUs 14-19 (socket 1)
PEs 12-17 (process 2) -> logical PUs 6-11 (socket 0)
PEs 18-23 (process 3) -> logical PUs 20-25 (socket 1)

Also works correctly on Bridges.

src/conv-core/cpuaffinity.C
```diff
@@ -442,62 +453,226 @@ static void cpuAffinityHandler(void *m)
...
-  myrank = rec->rank%rec->ncores;
-  while (in_exclude(myrank)) {             /* skip excluded core */
-    myrank = (myrank+1)%rec->ncores;
-    rec->rank ++;
+
+  int & node_host = rankmsg->node_hosts[node];
+  if (node_host == -1)
+  {
+    rankmsg->node_ranks_on_host[node] = rec->node_rank++;
+    node_host = rec->seq;
```
rbuch: `node_host` appears to be unused. I think it'd be clearer if the assignment done to `node_hosts[node]` was done explicitly rather than via a reference.

src/conv-core/cpuaffinity.C
```diff
@@ -442,62 +453,226 @@ static void cpuAffinityHandler(void *m)
...
+    DEBUGP(("Cpuaffinity> %zu unique hosts detected!\n", hostTable.size()));
     for (const auto & pair : hostTable)
       CmiFree(pair.second);
     hostTable.clear();
-#if 1
+
     /* bubble sort ranks on each node according to the PE number */
-    {
     int i,j;
+    for (i=0; i<nnodes-1; i++)
```
rbuch: Another nitpick, but unless I'm missing something, I think this loop and the below loop can both be done in linear time by iterating through and directly assigning values in ascending order per node/PE on the same host rather than doing this.

src/conv-core/cpuaffinity.C
evan-charmworks: `node_host` is checked against -1 and then assigned. I made it a reference because of these two uses.

src/conv-core/cpuaffinity.C
evan-charmworks: Could you share some code or pseudocode to illustrate what you mean?

src/conv-core/cpuaffinity.C
rbuch: Yeah, I see that, I'm fine with it as it is, I just think I stylistically prefer not doing it like that because it's less clear to me that the assignment is really happening in `node_hosts`.

src/conv-core/cpuaffinity.C
rbuch: I've implemented it in #2672.

Issue #1315, opened 2016-11-28
examples/charm++/jacobi*d are non-exemplary HPC code, using nested arrays


PhilMiller:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1315*

---

``` c++
// allocate a three dimensional array
temperature = new double**[blockDimX+2];
new_temperature = new double**[blockDimX+2];
for(int i=0; i<blockDimX+2; i++)
{
  temperature[i] = new double*[blockDimY+2];
  new_temperature[i] = new double*[blockDimY+2];
  for(int j=0; j<blockDimY+2; j++)
  {
    temperature[i][j] = new double[blockDimZ+2];
    new_temperature[i][j] = new double[blockDimZ+2];
  }
}
```

This is exactly what we would tell people *not* to do if we saw it in application code that was getting poor performance. Why do we have examples which do exactly that?

ericjbohm:
*Original date: 2018-05-09 16:07:06*

---
Assigning to Juan for eventual reassignment to interns or newer junior students as this seems like a worthy learning curve task.

PR #2670, opened 2020-01-08
jacobi: do not use VLA

matthiasdiener:


evan-charmworks:
Surely there is a better option than three-deep std::vector. Would it be difficult to replace this with a 1D vector and replace accesses with a macro for the index that does multiplication?

matthiasdiener:
> Surely there is a better option than three-deep std::vector. Would it be difficult to replace this with a 1D vector and replace accesses with a macro for the index that does multiplication?

Hmm, what's the issue with a three-deep std::vector? This seems such an uncritical part of the code that replacing them with macro indexing or nested mallocs appears to be more trouble than it's worth.

evan-charmworks:
> > Surely there is a better option than three-deep std::vector. Would it be difficult to replace this with a 1D vector and replace accesses with a macro for the index that does multiplication?
> 
> Hmm, what's the issue with a three-deep std::vector? This seems such an uncritical part of the code that replacing the vectors with macro indexing or nested mallocs appears to be more trouble than it's worth.

It, like nested mallocs, are poor form. Fortunately macro indexing was not hard to put in place.

matthiasdiener:
This is the type of code I was hoping to avoid:

```c++
std::vector<int> jmap(num_chare_x * num_chare_y * num_chare_z);
#define jmap_3d(x, y, z) (jmap[(x) * num_chare_y * num_chare_z + (y) * num_chare_z + (z)])
```

> It, like nested mallocs, are poor form.

You didn't say why it is poor form. I'd say that nested vectors are clearer and more readable, while performance seems to be mostly a wash (see e.g. https://scicomp.stackexchange.com/questions/3159/is-it-a-good-idea-to-use-vectorvectordouble-to-form-a-matrix-class-for-high). Like I mentioned above, performance isn't really an issue here anyways, since the code is run at initialization only.

evan-charmworks:
> You didn't say why it is poor form.

It took me a while to find, but I was thinking of issue #1315.

matthiasdiener:
> > You didn't say why it is poor form.
> 
> It took me a while to find, but I was thinking of issue #1315.

For #1315, I do agree that flat arrays are the better choice due to the performance sensitivity and the fact that the arrays get used throughout the whole execution. I'd say this would be overkill for the arrays in question here.

As an aside, the code in #1315 seems to have been changed already, so that issue might be closed now?

matthiasdiener:
Ready for review.

matthiasdiener:
Needs another review.

PR #2671, opened 2020-01-08
 Memory: Clean up locking around allocator implementations

evan-charmworks:
- Use of the ptmalloc3 "gnu" allocator has been made more modular so that it is now easier to add new malloc backends besides it and the OS allocator, as well as to allow further work toward using it as a memory pool for Isomalloc, HAPI, or more, if desired.
- Underlying malloc implementations are now exposed in two ways: `mm_*` symbols lock if the implementation requires it (such as ptmalloc3), while `mm_impl_*` symbols do no locking (unless the implementation, such as the OS, does its own). This allows our memory layers to handle locking as efficiently as possible.
- `-memory leak`, `-memory paranoid`, and `-memory verbose` are now thread-safe in all circumstances, including the default `-memory os-*`, by handling locking as part of their implementations. `-memory charmdebug` is also closer to being thread-safe, but it would require more changes outside of this PR's scope.
- Previously, `-memory gnu-*` always locked around the calls into our Converse memory layers, which allowed thread-safe usage of most memory modes, but also added unnecessary locking to ones that do not need it. This has been eliminated.
- The hook approach to intercepting malloc has been modularized so it does not require manual placement of `BEFORE`/`AFTER_MALLOC_CALL` at call sites. With this in place, in theory it may not require much work to add `libmemory-hooks-` variants of modes besides `charmdebug` to `src/scripts/Makefile`, though this is likely of interest to no one, especially since the hook method is not thread-safe.

### DIFF
--- a/src/conv-core/memory-lock.C
+++ b/src/conv-core/memory-lock.C
@@ -1,10 +1,4 @@
 
-/* Wrap a CmiMemLock around this code */
-#define MEM_LOCK_AROUND(code) \
-  CmiMemLock(); \
-  code; \
-  CmiMemUnlock();
-
 static void meta_init(char **argv)
 {
 /*   if (CmiMyRank()==0) CmiMemoryIs_flag|=CMI_MEMORY_IS_OSLOCK;   */
@@ -13,40 +7,40 @@ static void meta_init(char **argv)
 void *meta_malloc(size_t size)
 {
   void *result;
-  MEM_LOCK_AROUND( result = mm_malloc(size); )
+  MEM_LOCK_AROUND( result = mm_impl_malloc(size); )
   if (result==NULL) CmiOutOfMemory(size);
   return result;
 }
 
 void meta_free(void *mem)
 {
-  MEM_LOCK_AROUND( mm_free(mem); )
+  MEM_LOCK_AROUND( mm_impl_free(mem); )
 }
 
 void *meta_calloc(size_t nelem, size_t size)
 {
   void *result;
-  MEM_LOCK_AROUND( result = mm_calloc(nelem, size); )
+  MEM_LOCK_AROUND( result = mm_impl_calloc(nelem, size); )
   if (result==NULL) CmiOutOfMemory(size);
   return result;
 }
 
 void meta_cfree(void *mem)
 {
-  MEM_LOCK_AROUND( mm_cfree(mem); )
+  MEM_LOCK_AROUND( mm_impl_cfree(mem); )
 }
 
 void *meta_realloc(void *mem, size_t size)
 {
   void *result;
-  MEM_LOCK_AROUND( result = mm_realloc(mem, size); )
+  MEM_LOCK_AROUND( result = mm_impl_realloc(mem, size); )
   return result;
 }
 
 void *meta_memalign(size_t align, size_t size)
 {
   void *result;
-  MEM_LOCK_AROUND( result = mm_memalign(align, size); )
+  MEM_LOCK_AROUND( result = mm_impl_memalign(align, size); )
   if (result==NULL) CmiOutOfMemory(align*size);
   return result;    
 }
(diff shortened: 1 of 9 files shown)


evan-charmworks:
This PR makes the following improvements:

- Use of the ptmalloc3 "gnu" allocator has been made more modular so that it is now easier to add new malloc backends besides it and the OS allocator, as well as to allow further work toward using it as a memory pool for Isomalloc, HAPI, or more, if desired.
- Underlying malloc implementations are now exposed in two ways: `mm_*` symbols lock if the implementation requires it (such as ptmalloc3), while `mm_impl_*` symbols do no locking (unless the implementation, such as the OS, does its own). This allows our memory layers to handle locking as efficiently as possible.
- `-memory leak`, `-memory paranoid`, and `-memory verbose` are now thread-safe in all circumstances, including the default `-memory os-*`, by handling locking as part of their implementations. `-memory charmdebug` is also closer to being thread-safe, but it would require more changes outside of this PR's scope.
- Previously, `-memory gnu-*` always locked around the calls into our Converse memory layers, which allowed thread-safe usage of most memory modes, but also added unnecessary locking to ones that do not need it. This has been eliminated.
- The hook approach to intercepting malloc has been modularized so it does not require manual placement of `BEFORE`/`AFTER_MALLOC_CALL` at call sites. With this in place, in theory it may not require much work to add `libmemory-hooks-` variants of modes besides `charmdebug` to `src/scripts/Makefile`, though this is likely of interest to no one, especially since the hook method is not thread-safe.

src/conv-core/memory-lock.C
```diff
@@ -13,72 +7,72 @@ static void meta_init(char **argv)
 void *meta_malloc(size_t size)
 {
   void *result;
-  MEM_LOCK_AROUND( result = mm_malloc(size); )
+  MEM_LOCK_AROUND( result = mm_impl_malloc(size); )
```
minitu: Does `MEM_LOCK_AROUND` still work with the macro definition removed?

evan-charmworks: Yes, it is in memory.C now.

stwhite91:
> This PR makes the following improvements:
> 
> * Use of the ptmalloc3 "gnu" allocator has been made more modular so that it is now easier to add new malloc backends besides it and the OS allocator, as well as to allow further work toward using it as a memory pool for Isomalloc, HAPI, or more, if desired.
> * Underlying malloc implementations are now exposed in two ways: `mm_*` symbols lock if the implementation requires it (such as ptmalloc3), while `mm_impl_*` symbols do no locking (unless the implementation, such as the OS, does its own). This allows our memory layers to handle locking as efficiently as possible.
> * `-memory leak`, `-memory paranoid`, and `-memory verbose` are now thread-safe in all circumstances, including the default `-memory os-*`, by handling locking as part of their implementations. `-memory charmdebug` is also closer to being thread-safe, but it would require more changes outside of this PR's scope.
> * Previously, `-memory gnu-*` always locked around the calls into our Converse memory layers, which allowed thread-safe usage of most memory modes, but also added unnecessary locking to ones that do not need it. This has been eliminated.
> * The hook approach to intercepting malloc has been modularized so it does not require manual placement of `BEFORE`/`AFTER_MALLOC_CALL` at call sites. With this in place, in theory it may not require much work to add `libmemory-hooks-` variants of modes besides `charmdebug` to `src/scripts/Makefile`, though this is likely of interest to no one, especially since the hook method is not thread-safe.

Could you put this in the commit message body?

evan-charmworks:
> Could you put this in the commit message body?

Done

minitu:
Approved. So how can I use this in HAPI with `cudaMalloc` as a backend?

evan-charmworks:
> Approved. So how can I use this in HAPI with `cudaMalloc` as a backend?

I'm still looking at that. This commit is cleanup that allows further work toward this goal.

evan-charmworks:
This diff may be easier to view by running `git fetch && git diff -w --color-moved origin/master...origin/evan/cleanup-memlock`.

evan-charmworks:
Could this get another review when someone has the chance? I would like to push other PRs that build on the work done here.

PR #2682, opened 2020-01-16
GitHub CI: limit jobs to 45 minutes

matthiasdiener:
Should prevent long-running CI hangs (default timeout is 360 mins) such as the one seen in #2458.


matthiasdiener:
Needs another review.

PR #2685, opened 2020-01-20
Add AMPI-only build target

evan-charmworks:
This target will be customized to disable parts of Charm++ unused by AMPI and tune remaining settings for it.


matthiasdiener:
> This target will be customized to disable parts of Charm++ unused by AMPI and tune remaining settings for it.

Can you add more info about what these parts/optimizations could be?

evan-charmworks:
> > This target will be customized to disable parts of Charm++ unused by AMPI and tune remaining settings for it.
> 
> Can you add more info about what these parts/optimizations could be?

From discussion at Charmworks by @epmikida and @stwhite91 , these things can be disabled:
- dynamic insertion
- demand creation
- anytime migration
- quiescence detection
- completion detection
- CcdCallbacks (except as used for LBAtSyncer)
- Charm-MPI interop
- message priorities
- scheduling/queueing policies
- seed load balancers
- Chare arrays other than 1D
- sections (until subcommunicator collectives are implemented)
- TRAM
- BigSim
- CkLoop
- libraries other than TCharm

buildcmake
```diff
@@ -261,6 +265,13 @@ done
 
 ############ End option parsing
 
+if [[ "$target" = "${target%AMPI-only}AMPI-only" ]]; then
```
matthiasdiener: why wouldnt this just be `if [[ "$target" = "AMPI-only" ]]` ?

evan-charmworks: For all-test-AMPI-only

matthiasdiener:
There appear to be two different ways to enable the AMPI-only build, `./build AMPI-only` and `./build ... --ampi-only`. Would it be sufficient to have only one of these? (I'd vote for using`./build AMPI-only` since this disables some core functionality).

Otherwise, the patch looks good. It'd be interesting to see some performance results before merging this though. For many of the things that can be disabled mentioned above, it's hard to see how that would improve an AMPI application.

epmikida:
> There appear to be two different ways to enable the AMPI-only build, `./build AMPI-only` and `./build ... --ampi-only`. Would it be sufficient to have only one of these? (I'd vote for using`./build AMPI-only` since this disables some core functionality).
> 
> Otherwise, the patch looks good. It'd be interesting to see some performance results before merging this though. For many of the things that can be disabled mentioned above, it's hard to see how that would improve an AMPI application.

Many of the things mentioned above add overhead to array messaging for every message. We don't know exactly what the overheads are yet, but especially stuff like dynamic insertion, anytime migration, and demand creation all add extra checks to array messaging. There is also possible overheads in terms of extra memory used for bookkeeping.

Part of the motivation for this is also as sort of a test for Charm++ 7.0 where one of the stated goals is to have more sensible defaults, and most of the above features fall into the category of things we think should not be on by default, and should be opt in instead. We can't just go ahead and change all these defaults in Charm++ without breaking a ton of stuff. But since AMPI doesn't use many of these features, nor do the users of AMPI, it gives a good testbed to see how this will work.

I'm not sure if we will see a huge impact from this change just yet, because simply disabling these features may not remove all the overhead simply because the code path for all of these features is still intertwined with the basis code paths. A next step from this I'd imagine would be some actual code refactoring of these features so that when they are disabled they are more effectively removed from the relevant code paths. Something like a basic array subclass that only has the most basic functionality would be one possibility.

epmikida:
Changes look fine to me, although along similar reasoning we are using for Charm 7.0 better defaults, I think we'd eventually want to switch AMPI-only to AMPI. Adding an extra option just adds to the already confusing set of Charm++ build options, and if this works as intended, I don't see any reason for people to need the AMPI target unless they are already advanced users and know what they are doing.

evan-charmworks:
> There appear to be two different ways to enable the AMPI-only build, `./build AMPI-only` and `./build ... --ampi-only`. Would it be sufficient to have only one of these? (I'd vote for using`./build AMPI-only` since this disables some core functionality).

`--ampi-only` is a failsafe option for something like `all-test-AMPI-tests` where we can't append the `-only` suffix and test for it easily.

evan-charmworks:
- Removed mention of `AMPI-only` from CMakeLists.txt, since it is handled by `buildcmake` and not the txt itself
- Fixed suffix removal in `build` for all-test-AMPI-tests

evan-charmworks:
Needs a second review

PR #2686, opened 2020-01-20
AmpiMsgPool: Use constant length vector instead of forward_list

evan-charmworks:
This avoids heap allocations when managing the pool.


matthiasdiener:
Could you use a std::array?

evan-charmworks:
This class is designed to have its parameters adjustable at runtime, while std::array requires knowing the bounds at compile time.

stwhite91:
> This class is designed to have its parameters adjustable at runtime, while std::array requires knowing the bounds at compile time.

This will allow us to expose the size of the pool as a runtime environment variable for easier tuning, but it's not currently exposed that way.

Issue #1925, opened 2018-05-29
Mac builds fail with "No gfortran found, exiting"


jcphill:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1925*

---
Fortran support was neither requested nor required.  Break is likely due to this change:

commit e82ae8242b3dde160c5a5cc4cfec6f4d86613859
Author: Evan Ramos <evan`hpccharm.com>
Date:   Fri May 18 16:48:04 2018 -0500

    Call conv-mach-gfortran.sh from conv-mach-darwin.sh
    
    Change-Id: Id6a48c630e30ea88f888d9628aa564bf59bbe767

evan-charmworks:
*Original date: 2018-05-29 21:45:05*

---
~~https://charm.cs.illinois.edu/gerrit/4237~~ https://github.com/UIUC-PPL/charm/commit/61e5995ef530fca7cebdfc1c224d047476fb8b87

PR #2687, opened 2020-01-22
uth: use gfortran

matthiasdiener:


src/arch/uth-linux-x86_64/conv-mach.sh
```diff
@@ -1,8 +1,7 @@
 . $CHARMINC/cc-gcc.sh
+. $CHARMINC/conv-mach-gfortran.sh
```
evan-charmworks: This should probably be:
```
if command -v gfortran >/dev/null 2>&1
then
  . $CHARMINC/conv-mach-gfortran.sh
fi
```
See #1925

src/arch/uth-linux/conv-mach.sh
```diff
@@ -1,8 +1,7 @@
 . $CHARMINC/cc-gcc.sh
+. $CHARMINC/conv-mach-gfortran.sh
```
evan-charmworks: Same

src/arch/uth-linux-x86_64/conv-mach.sh
matthiasdiener: done.

matthiasdiener: done

matthiasdiener:
Needs another review.

PR #2689, opened 2020-01-24
Windows autobuild fixes

evan-charmworks:

### DIFF
--- a/src/libs/ck-libs/armci/armci_impl.h
+++ b/src/libs/ck-libs/armci/armci_impl.h
@@ -240,10 +240,17 @@ class ArmciVirtualProcessor : public TCharmClient1D {
   ~ArmciVirtualProcessor();
  
   pointer BlockMalloc(int bytes) { 
-    return CmiIsomallocContextMalloc(CmiIsomallocGetThreadContext(thread->getThread()), bytes);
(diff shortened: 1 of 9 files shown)


src/libs/ck-libs/armci/armci_impl.h
```diff
@@ -240,10 +240,17 @@ class ArmciVirtualProcessor : public TCharmClient1D {
   ~ArmciVirtualProcessor();
  
   pointer BlockMalloc(int bytes) { 
-    return CmiIsomallocContextMalloc(CmiIsomallocGetThreadContext(thread->getThread()), bytes);
+    CmiIsomallocContext * ctx = CmiIsomallocGetThreadContext(thread->getThread());
+    if (ctx == nullptr)
```
stwhite91: Should this if-else instead go inside of CmiIsomallocContextFree()? Or is it sometimes not needed

evan-charmworks: It looks like everywhere else already has a check of its own near the call site. threads.C checks `base->isMigratable`, meta_active in memory-isomalloc.C checks `CpvAccess(isomalloc_state).context`. If we want to move these checks to the call site we can discuss that for future cleanup.

matthiasdiener:
The github CI build failures are spurious, this is ready for merge.

PR #2692, opened 2020-01-27
build: retire createlink, gather*.local, system_ln

matthiasdiener:


build
```diff
@@ -760,66 +760,35 @@ mkdir -p "$DESTINATION/tmp"
...
-  chmod +x $DESTINATION/tmp/system_ln
-  newdirlist='bin lib'
-  [ -n "$BUILD_SHARED" ] && newdirlist="$newdirlist lib_so"
-  newdirlist="$newdirlist include tmp"
-  for newdir in $newdirlist
-  do
-    Echo "Soft-linking over $newdir"
-    if [ -r $newdir ]
-    then
-      rm -fr $newdir || exit 1
```
rbuch: Should this return check still be there?

matthiasdiener: Not sure. I guess `rm -f` can not fail, so the `exit 1` is never executed.

matthiasdiener:
> For the sake of completeness, I assume you've tested this on dexterity?

Yes, it's tested on dexterity (and passes the Windows CI check on Travis). The reason this is still a draft PR is that there is a substantial build slowdown on Windows (~15 min, both in the symlink creation and the actual build) with this patch, and I have no idea why.

PR #2693, opened 2020-01-28
CPU Affinity: Remove unneeded Cray-specific code

evan-charmworks:


evan-charmworks:
When I tested this it didn't resolve all issues.

PR #2694, opened 2020-01-30
Use dlmalloc as Isomalloc's mempool

evan-charmworks:
(The current 32-bit netlrts-linux build failure relates to my prior Isomalloc refactoring PR, and in my testing it is fixed in this one.)

### DIFF
--- a/src/conv-core/isomalloc.C
+++ b/src/conv-core/isomalloc.C
@@ -909,6 +909,18 @@ struct CmiIsomallocContext
    * This class is a wholly self-contained allocator, depending only on mmap/munmap.
    * It batches allocations of all sizes into a slab region that can expand as needed.
    *
+   * Advantages:
+   * This mempool provides space-efficient memory management. It has been thoroughly
+   * verified for correctness. It contains robust assertions to pinpoint problems in the
+   * event of future modifications. The code is clean, self-documenting, modular, short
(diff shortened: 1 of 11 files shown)


src/conv-core/isomalloc.C
```diff
@@ -896,19 +896,210 @@ static void CmiIsomallocInitExtent()
...
+    backend.allocated_extent += length;
+    return mapped;
+  }
+  virtual void * call_direct_mmap(size_t length) override final
+  {
+    return this->isomalloc_dlmalloc::call_mmap(length); // call call_mmap without the vtable lookup
+  }
+  virtual int call_munmap(void * addr, size_t length) override final
+  {
+    // TODO
```
stwhite91: Is these something more to do here?

evan-charmworks: This implementation currently meets or exceeds what the Isomempool does. The `TODO` refers to  maintaining a record of mmapped regions so that migration can skip the gaps. Complexity arises when considering RDMA transfers, and we would need to discuss when it is better to invoke multiple transfers, or to leave small gaps mapped so they could be combined into one.

stwhite91: I see, can you just put this note into the TODO comment?

evan-charmworks: Done

PR #2695, opened 2020-01-30
Add Heap Record-Replay, -memory record

evan-charmworks:
Generates an MPI program that repeats the exact sequence of heap operations observed on each PE. Can be used for debugging and performance analysis of memory allocators as well as heap behavior of user codes.

### DIFF
--- a/src/conv-core/memory-record.C
+++ b/src/conv-core/memory-record.C
@@ -1,0 +1,329 @@
+// Heap Record-Replay
+// By Evan Ramos
+// Generates an MPI program that repeats the exact sequence of heap operations observed on each PE.
+
+#include <cstdint>
+#include <cstdio>
+#include <unordered_map>
+
+// raw string literals containing the output program
+
+static const char prologue[] =
+R"code(
+#include <stdlib.h>
+#include <stdio.h>
+#include "mpi.h"
+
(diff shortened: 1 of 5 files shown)


src/conv-core/memory-record.C
```diff
@@ -0,0 +1,329 @@
...
+  if (CmiMyRank() == 0)
+    initialized = 1;
+
+  CmiNodeAllBarrier();
+
+  if (mype == 0)
+  {
+    const int numpes = CmiNumPes();
+
+    FILE * mainsrc = fopen("heap-replay.cpp", "wb");
```
matthiasdiener: should probably be 'w' instead of 'wb' (same for the other fopen())

matthiasdiener:
This is an interesting idea. Can you add more information about possible use cases for this?

src/conv-core/memory-record.C
evan-charmworks: Done

evan-charmworks:
> This is an interesting idea. Can you add more information about possible use cases for this?

Added. I wrote this to help diagnose what Laghos was doing that caused the Isomempool to perform so poorly. By inspecting the output files I could see that it was allocating and then freeing the same two sequences of buffers tens of thousands of times.

PR #2696, opened 2020-01-30
Make array broadcasts node aware

epmikida:
Very little change was required to get this working at a basic level, because array bcasts are based on group broadcasts. Main changes were adding array broadcast message types (one for the initial broadcast, and one for forwarding the broadcast to each group on a node), stopping CkArray from modifying the message, and adding a nokeep broadcast to CkArray.


epmikida:
This is not quite ready for review yet since I need to add a test for it, and update charm debug information. It is also broken with BigSim in a much more significant way than the group broadcasts were so I need to figure out how to address that. It may also need small tweaks for ZC which me and Nitin had already briefly discussed.

epmikida:
And by more significant BigSim changes, I mean that even just trying to turn this off for BigSim as a work around requires a decent amount of change as far as I can tell so far, but I haven't yet got it to work.

epmikida:
The change is now fixed and ready for everything but ZC and BigSim. I will need to work with Nitin to figure out what is going wrong with ZC, and BigSim is now being deprecated.

nitbhat:
> The change is now fixed and ready for everything but ZC and BigSim. I will need to work with Nitin to figure out what is going wrong with ZC, and BigSim is now being deprecated.

I've pushed the fixes to the ZC implementation based on the changed semantics for chare array broadcast. These fixes include the following changes:
1) Not resetting the message type to `ForBocMsg` (which was required previously in order to reuse the same message)
2) For broadcast from non-zero root, by default `recvBroadcast` was called. This has been modified to called the expedited and noKeep versions of recvBroadcast based on entry method options.

evan-charmworks:
Updated to master that removes BigSim. @epmikida is this ready for review?

epmikida:
> Updated to master that removes BigSim. @epmikida is this ready for review?

Yes.

src/ck-core/charm.h
```diff
@@ -342,12 +342,14 @@ typedef enum {
   ForArrayEltMsg       =21,              // Array Element entry method message
   ForIDedObjMsg        =22,
   BocBcastMsg          =23,
+  ArrayBcastMsg        =24,
```
nitbhat: One-line comments describing these two message types would be helpful.

src/ck-core/ck.C
```diff
@@ -3043,11 +3065,11 @@ void CkMessageWatcherInit(char **argv,CkCoreState *ck) {
 
 int CkMessageToEpIdx(void *msg) {
```
nitbhat: It'll be good to fix the weird indentation within this function since you're already touching this code.

src/conv-core/debug-conv.h
```diff
@@ -10,7 +10,7 @@
 /*If you are incrementing these numbers, you also need to increment MAJOR and MINOR
 variables in ParDebug.java(in ccs_tools) to match, otherwise CharmDebug won't work*/
 #define CHARMDEBUG_MAJOR   10
-#define CHARMDEBUG_MINOR   10
+#define CHARMDEBUG_MINOR   11
```
nitbhat: Was this intentional? When is it required to increment these numbers? 

Also, note the comment about incrementing ccs_tools.

epmikida: It's mentioned in charm.h at line 316. Although the previous changes for group broadcasts were never merged from this PR: https://github.com/UIUC-PPL/ccs_tools/pull/2 so I'll have to address that as well.

nitbhat:
There is a crash in MPI-smp during automated testing in `examples/charm++/zerocopy/entry_method_post_api/prereg`, which is a p2p message test. It seems spurious, so I'll restart this build. 


```
make[5]: Entering directory '/home/travis/build/UIUC-PPL/charm/mpi-linux-x86_64-smp/examples/charm++/zerocopy/entry_method_post_api/prereg'
make -C simpleZeroCopy test OPTS='-optimize -production  -g -Werror=vla' TESTOPTS='+setcpuaffinity'
make[6]: Entering directory '/home/travis/build/UIUC-PPL/charm/mpi-linux-x86_64-smp/examples/charm++/zerocopy/entry_method_post_api/prereg/simpleZeroCopy'
../../../../../../bin/testrun  +p4 ./simpleZeroCopy 32 +balancer RotateLB +setcpuaffinity
Running on 4 processors:  ./simpleZeroCopy 32 +balancer RotateLB +setcpuaffinity 
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./simpleZeroCopy 32 +balancer RotateLB +setcpuaffinity 
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 4 processes, 1 worker threads (PEs) + 1 comm threads per process, 4 PEs total
Converse/Charm++ Commit ID: 0fa61e4
Charm++> Running on 1 hosts (1 sockets x 1 cores x 2 PUs = 2-way SMP)
Charm++> cpu topology info is gathered in 0.078 seconds.
Charm++> Warning: the number of SMP threads (8) is greater than the number of physical cores (2), so threads will sleep while idling. Use +CmiSpinOnIdle or +CmiSleepOnIdle to control this directly.
WARNING: Multiple PEs assigned to same core, recommend adjusting processor affinity or passing +CmiSleepOnIdle to reduce interference.
CharmLB> RotateLB created.
send: completed
Fatal error in MPI_Irecv: Invalid buffer pointer, error stack:
MPI_Irecv(153): MPI_Irecv(buf=(nil), count=840, MPI_BYTE, src=0, tag=1378, comm=0x84000002, request=0x7ffff003b910) failed
MPI_Irecv(118): Null buffer pointer
```

src/ck-core/ckarray.C
```diff
@@ -1539,6 +1548,12 @@ void CkArray::recvBroadcast(CkMessage* m) {
     bool doFree = true; // free it since all ops are done
     broadcaster->deliver(msg, (ArrayElement*)localElemVec[0], doFree);
   } else if (zc_msgtype == CMK_ZC_BCAST_RECV_MSG && len > 0 ) {
+
+    // Set Array manager for this message as it is required for ZC Post API
+    env->setMsgtype(ForArrayEltMsg);
```
epmikida: Won't this cause a race? I think once the message is received we should not be modifying the message? Same goes for a few things below. I think we can use the group id in the envelope rather than explicitly setting and getting array manager (they should be the same).

src/ck-core/ckrdma.C
```diff
@@ -1883,15 +1867,20 @@ void sendAckMsgToParent(envelope *env)  {
 }
 
 CkArray* getArrayMgrFromMsg(envelope *env) {
+
+  env->setMsgtype(ForArrayEltMsg); // Used to bypass assertion in getArrayMgr
```
epmikida: Same as above. I think the group id in the envelope can be used instead of array mgr.

src/ck-core/ckarray.C
nitbhat: The way the Post API is implemented for chare arrays is that a `CMK_ZC_BCAST_RECV_MSG` reaches only one array element of one of the PEs of the node. So, at this point, it is the only PE that is operating on this message. For that reason, it doesn't cause a race condition. 

However, I like the idea of not modifying the message type and getting the array manager using the group id. I'll incorporate that change.

src/ck-core/ckrdma.C
nitbhat: Similarly, getArrayMgrFromMsg is only called either from the pe performing the `Rget` or from the intermediate node in the bcast tree. Even there, it looks like only one PE actually calls it. However, even this can be replaced with your suggestion of using the group id. I'll make that change and push.

src/conv-core/debug-conv.h
nitbhat: Line 316 reads "/*** WARNING!!!! The following enum is linked to charmdebug finals in MsgInfo.java.
 *   Make sure the two remain synchronized if changing this one." 

So, everytime the message type enum is updated, is it required to increment the CHARMDEBUG version? Why is only CHARMDEBUG_MINOR updated?

nitbhat:
> I see some potential race conditions with the ZC changes which may need to be addressed.

I've modified that code to not write into the envelope and just use the GroupID instead.

epmikida:
> > I see some potential race conditions with the ZC changes which may need to be addressed.
> 
> I've modified that code to not write into the envelope and just use the GroupID instead.

Ah I didn't realize those calls were only on one PE. But I agree, I think the consistency of not modifying the message is a good change to make anyways.

src/ck-core/ck.C
```diff
@@ -3055,13 +3077,13 @@ void CkMessageWatcherInit(char **argv,CkCoreState *ck) {
...
-int CkMessageToEpIdx(void *msg) {
-        envelope *env=UsrToEnv(msg);
-	int ep=env->getEpIdx();
-	if (ep==CkIndex_CkArray::recvBroadcast(0))
-		return env->getsetArrayBcastEp();
-	else
-		return ep;
+int CkMessageToEpIdx(void* msg) {
+  envelope* env = UsrToEnv(msg);
+  int type = env->getMsgtype();
```
ericjbohm: Probably won't matter in this case, but shouldn't these convenience stack temporaries (i.e., type) be const or even const reference?

evan-charmworks: Couldn't hurt, but it could be a wider-ranging cleanup effort and it probably isn't worth holding back the PR.

PR #2697, opened 2020-01-30
AMPI: Include accessory implementation files directly from ampi.C

evan-charmworks:
This includes:
ampiMisc.C
mpich-alltoall.C
ampiOneSided.C
ampi_noimpl.C
ampi_mpix.C

This allows these files to better share the inline functions in ampiimpl.h without requiring link-time optimization. It also should eliminate time spent redundantly compiling ampiimpl.h.

### DIFF
--- a/src/libs/ck-libs/ampi/CMakeLists.txt
+++ b/src/libs/ck-libs/ampi/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(ampi-cxx-sources ampi.C ampiMisc.C ampiOneSided.C ampif.C ddt.C mpich-alltoall.C ampi_mpix.C)
+set(ampi-cxx-sources ampi.C ampif.C ddt.C)
 
 set(ampi-f90-sources ampifimpl.f90 ampimod.f90)
 
--- a/src/libs/ck-libs/ampi/Makefile
+++ b/src/libs/ck-libs/ampi/Makefile
@@ -11,8 +11,7 @@ HEADDEP=$(HEADERS) ampiimpl.h ddt.h \
 COMPAT=compat_ampi.o \
        compat_ampim.o compat_ampifm.o compat_ampicm.o \
 	   compat_ampicpp.o
-OBJS=ampi.o ampif.o ampiOneSided.o \
-     ampiMisc.o ddt.o mpich-alltoall.o ampi_mpix.o ampi_noimpl.o
+OBJS=ampi.o ampif.o ddt.o
 
 AMPI_LIB=libmoduleampi
 AMPI_LIBDIR=$(CDIR)/lib
@@ -160,27 +159,14 @@ compat_ampicm.o: compat_ampicm.C
 compat_ampi.o: compat_ampi.c
 	$(CHARMC) -c compat_ampi.c
 
-ampi_mpix.o: ampi_mpix.C $(HEADDEP)
-
-ampi_noimpl.o: ampi_noimpl.C $(HEADDEP)
-
 compat_ampicpp.o: compat_ampicpp.C
 	$(CHARMC) -c compat_ampicpp.C
 
-ampi.o: ampi.C $(HEADDEP)
-	$(CHARMC) -c ampi.C
+ampi.o: ampi.C ampiMisc.C mpich-alltoall.C ampiOneSided.C ampi_noimpl.C ampi_mpix.C $(HEADDEP)
+	$(CHARMC) -c $<
 
 ampif.o: ampif.C $(HEADDEP)
 	$(CHARMC) -c ampif.C
-
-ampiOneSided.o: ampiOneSided.C ampiimpl.h $(HEADDEP)
-	$(CHARMC) -c ampiOneSided.C
-
-ampiMisc.o: ampiMisc.C ampiimpl.h $(HEADDEP)
-	$(CHARMC) -c ampiMisc.C
-
-mpich-alltoall.o: mpich-alltoall.C $(HEADDEP)
-	$(CHARMC) -c mpich-alltoall.C
 
 ampi.decl.h ampi.def.h: ampi.ci
 	$(CHARMC) ampi.ci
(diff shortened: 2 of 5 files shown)


evan-charmworks:
> Have you measured the run-time performance improvements? Most functions in ampiimpl.h are inlined anyways (either explicitly or implicitly). How would these benefit from this patch?

I have not measured this commit's performance. I am working on another PR that makes more functions eligible for inlining from ampiimpl.h in response to observations I made while profiling Laghos on AMPI with callgrind.

In the case where the compiler decides not to inline a function, it will need to leave a copy of it in every object file that uses the non-inline version. I don't know if linkers will strip the duplicates without LTO. If they don't, that is inefficient with regard to instruction fetch at runtime.

Since optimizing compilers can perform reordering of functions, it should be useful for the compiler to see the entire set of ampiimpl.h users in a single translation unit, with or without LTO.

> The downside to this patch is that compiling ampi.C will now be even slower than before, potentially increasing overall build time.

In terms of ampi.C by itself, true. The big picture is more complicated when considering the entire build process. It may be a wash.

> It's also somewhat awkward to #include .C files.

True. Some of them may not need to be separate files at all. Others could be renamed. That could be left as further cleanup.

evan-charmworks:
Another benefit is that code in the accessory files can inline functions in ampi.C as long as they are tagged `inline`, without needing to move them entirely into ampiimpl.h.

matthiasdiener:
> I am working on another PR that makes more functions eligible for inlining from ampiimpl.h in response to observations I made while profiling Laghos on AMPI with callgrind.

It could be interesting to see these observations. Function call overhead should probably be negligible except for trivial or virtual functions.

> In the case where the compiler decides not to inline a function, it will need to leave a copy of it in every object file that uses the non-inline version. I don't know if linkers will strip the duplicates without LTO. If they don't, that is inefficient with regard to instruction fetch at runtime.

I believe the compiler is *required* to inline at least the member functions defined in classes (which make up the majority of functions in ampiimpl.h), otherwise we would get link-time errors regarding multiple definitions of the same symbol (weak symbols wouldn't work in this case since they are not supported on Darwin at least). Checking e.g. mpich-alltoall.o or ampiMisc.o (mpich-alltoall.C itself defines very few functions, which makes checking this easier) on Darwin and Linux shows that none of the functions defined in ampiimpl.h are included in these .o files.

> Since optimizing compilers can perform reordering of functions, it should be useful for the compiler to see the entire set of ampiimpl.h users in a single translation unit, with or without LTO.

This is certainly possible, but I'd guess the improvements are relatively minor if at all. Note that due to resource constraints (compile time and memory usage), having larger files might negatively impact performance of the generated code by preventing the compiler from applying more expensive optimizations. We have seen such cases in PlasCom2.

> > It's also somewhat awkward to #include .C files.

> True. Some of them may not need to be separate files at all.

I suggest we should consider going the other way around, i.e. splitting ampi.C into different files corresponding to different functionality (e.g., startup, P2P, collectives, ...), instead of having a 10,000 line file of doom ;-) Most other MPI runtimes have a separate file for each API function they implement.

> Another benefit is that code in the accessory files can inline functions in ampi.C as long as they are tagged inline, without needing to move them entirely into ampiimpl.h.

We could certainly benefit from moving some small functions to ampiimpl.h after careful analysis. Note that this is the case that LTO tries to solve, ie. inlining functions *after* they have been compiled into a .o file.

evan-charmworks:
The thing about the `inline` keyword is that it does not tell the compiler that a function _should_ be inlined, only that it is eligible for it. In this example you can see a case where I mark a very long function `inline`, and the compiler does not inline it into another function that calls it twice and nothing else. https://godbolt.org/z/3XFCV3

I could see a case for dividing ampi.C into separate files for the purpose of organization. Whether they are #included or linked could be its own discussion. A separate object/translation unit for each MPI symbol would be excessive. Have you heard of "unity builds"? https://stackoverflow.com/questions/543697/include-all-cpp-files-into-a-single-compilation-unit

stwhite91:
The reason that OpenMPI and MPICH use the "one file per MPI_ function" approach was/is, from what I know, historically for PMPI_ support on systems that didn't support weak symbols they would compile each file twice, once with the symbols prefixed by MPI_ and another time by PMPI_.

matthiasdiener:
> The thing about the `inline` keyword is that it does not tell the compiler that a function _should_ be inlined, only that it is eligible for it. In this example you can see a case where I mark a very long function `inline`, and the compiler does not inline it into another function that calls it twice and nothing else. https://godbolt.org/z/3XFCV3

I'm assuming that you mean that `HugeFunction` is not inlined into `DeclineInline`? In this case, what would be the benefit of inlining `HugeFunction` into `DeclineInline`? The function call overhead is negligible due to the size of `HugeFunction`, so inlining would just increase code size (as well as cache misses) without a substantial benefit. Since the compiler isn't required to inline in this instance, it is imho correct not to inline.

> A separate object/translation unit for each MPI symbol would be excessive. 

For AMPI, I agree. But again, this is common in other MPI implementations (see our ROMIO for example).

> Have you heard of "unity builds"? https://stackoverflow.com/questions/543697/include-all-cpp-files-into-a-single-compilation-unit

Yeah, I remember doing those a long time ago. KDE 2 (https://en.wikipedia.org/wiki/K_Desktop_Environment_2) was somewhat notorious for benefitting from it.
With modern C++ compilers, I would assume that most arguments are not really relevant anymore. Note that the question and most answers are from 2009 in the link. More recent arguments seem to focus on build time (see e.g. http://onqtam.com/programming/2018-07-07-unity-builds/), but this is probably only relevant for serial builds if at all.

evan-charmworks:
> > The thing about the `inline` keyword is that it does not tell the compiler that a function _should_ be inlined, only that it is eligible for it. In this example you can see a case where I mark a very long function `inline`, and the compiler does not inline it into another function that calls it twice and nothing else. https://godbolt.org/z/3XFCV3
> 
> I'm assuming that you mean that `HugeFunction` is not inlined into `DeclineInline`? In this case, what would be the benefit of inlining `HugeFunction` into `DeclineInline`? The function call overhead is negligible due to the size of `HugeFunction`, so inlining would just increase code size (as well as cache misses) without a substantial benefit. Since the compiler isn't required to inline in this instance, it is imho correct not to inline.

I agree, and my point is that it does not hurt to mark a function inline if it should not be inlined because the compiler will make that decision for us.

matthiasdiener:
> I agree, and my point is that it does not hurt to mark a function inline if it should not be inlined because the compiler will make that decision for us.

Ok, agreed.

matthiasdiener:
I still don't think this patch should be merged without having a clear idea of the benefits. Right now, it just appears to have disadvantages.

PR #2700, opened 2020-02-05
allows msa programs to build with std::move, as added by xlat-i

Wingpad:
Xlat-i adds `std::move` to function arguments while unmarshalling for zero-copy support. This did not preclude compilation of the MSA library, but anything trying to use its headers will fail to build. This patch allows MSA programs to build again.


evan-charmworks:
I would merge this but there is some kind of glitch with the CI. @matthiasdiener do you know what is happening? @Wingpad for your next PR try pushing a topic branch to your forked repository and requesting a pull from it instead of your master branch.

matthiasdiener:
> I would merge this but there is some kind of glitch with the CI. @matthiasdiener do you know what is happening? @Wingpad for your next PR try pushing a topic branch to your forked repository and requesting a pull from it instead of your master branch.

I'm not sure what the issue is, but having a pull request from the 'master' branch might be the explanation, like you said. This seems to be a bug in the CI pipeline. I set the check to optional so that you can merge this. @Wingpad If you prefer, you can create new branches directly on UIUC-PPL/charm, no need for a fork.

PR #2703, opened 2020-02-07
Remove BigSim

evan-charmworks:


src/ck-core/ckreduction.C
```diff
@@ -2050,8 +2023,6 @@ CkNodeReductionMgr::CkNodeReductionMgr()//Constructor
 {
 #ifdef BINOMIAL_TREE
   init_BinomialTree();
-#elif CMK_BIGSIM_CHARM
-  init_BinaryTree();
```
epmikida: Is there no case where we'd want to use a binary tree? And if not, the BinaryTree code can probably be removed as well.

evan-charmworks:
> Looks like there are some removals related with bluegene as well?

That is due to some pieces of BigSim being referred to as Blue Gene, perhaps confusingly. For example, "make bgtest" is for BigSim testing, not BlueGene. `src/langs/bluegene` and `trace-bluegene` are also specific to BigSim.

matthiasdiener:
Just to clarify, the cleanup operations will be one in a separate PR?

evan-charmworks:
This PR focuses on removing BigSim itself and anything that unconditionally uses its headers. Further cleanup to orthogonal code special-cased for BigSim would be best served by attention from someone familiar with it. For example, @epmikida asked "Is there no case where we'd want to use a binary tree?", and I don't know the answer.

I used the following commands to verify the removal:

```
find . -iname '*blue*'
find . -iname '*big*'
grep --exclude-dir=src/libs/ck-libs/ampi/romio --exclude-dir=contrib/hwloc -rIiE '(bigsim|bignet|bigemu|blue[^Gg]|bluegene[^LPQ/])'
```

These find BlueGene/Q support and comments that are still relevant to the surrounding code.

PR #2705, opened 2020-02-10
Update year to 2020 in scripts and readme

rbuch:


matthiasdiener:
Would it make sense to change the date in CITATION.cff as well?

rbuch:
> Would it make sense to change the date in CITATION.cff as well?

I thought about that, but considering that the Zenodo citation we use in the cff file is the overarching umbrella one that we made in 2019, I think it should remain this way.

The other question was section seven in the license, which reads:

>  7 Copyright Notices.  All copies of the Software or derivative works
  shall include the following copyright notice: "Copyright 2019 The
  Board of Trustees of the University of Illinois. All rights
  reserved."

I assume that year shouldn't be updated, because then anything that includes the notice would have to be updated every year to avoid being in violation, I'd think. However, I'd be interested to hear other opinions on this.

matthiasdiener:
> > Would it make sense to change the date in CITATION.cff as well?
> 
> I thought about that, but considering that the Zenodo citation we use in the cff file is the overarching umbrella one that we made in 2019, I think it should remain this way.

Ok, makes sense.

> The other question was section seven in the license, which reads:
> 
> > 7 Copyright Notices.  All copies of the Software or derivative works
> > shall include the following copyright notice: "Copyright 2019 The
> > Board of Trustees of the University of Illinois. All rights
> > reserved."
> 
> I assume that year shouldn't be updated, because then anything that includes the notice would have to be updated every year to avoid being in violation, I'd think. However, I'd be interested to hear other opinions on this.

In any case, we could change that in a separate patch as well.

stwhite91:
@ericjbohm can you review this point?

> The other question was section seven in the license, which reads:
> 
> 7 Copyright Notices. All copies of the Software or derivative works
> shall include the following copyright notice: "Copyright 2019 The
> Board of Trustees of the University of Illinois. All rights
> reserved."
> 
> I assume that year shouldn't be updated, because then anything that includes the notice would have to be updated every year to avoid being in violation, I'd think. However, I'd be interested to hear other opinions on this.

PR #2707, opened 2020-02-13
Isomalloc: Use pup_buffer for migration

evan-charmworks:


evan-charmworks:
The verification failures are due to a problem we didn't foresee. If pup_buffer transfers don't complete inline, any changes made to the buffers inside PUP methods will be overwritten by the deferred transfer after the PUP methods are finished. This means that any unpacking onto the stack in AMPI_Register_pup methods will not be seen by the user program and stale data will be there instead, such as pointers that are invalid after migration. This can be reproduced with examples/ampi/Cjacobi3D: `./charmrun +p2 ./jacobi 2 1 1 40 +vp2 +balancer RotateLB ++local ++debug`.

stwhite91:
> The verification failures are due to a problem we didn't foresee. If pup_buffer transfers don't complete inline, any changes made to the buffers inside PUP methods will be overwritten by the deferred transfer after the PUP methods are finished. This means that any unpacking onto the stack in AMPI_Register_pup methods will not be seen by the user program and stale data will be there instead, such as pointers that are invalid after migration. This can be reproduced with examples/ampi/Cjacobi3D: `./charmrun +p2 ./jacobi 2 1 1 40 +vp2 +balancer RotateLB ++local ++debug`.

You're saying the Isomalloc'ed ULT stacks are corrupt if a ULT migrates across node when the user is writing their own PUP routines using AMPI_Register_pup, right? So an AMPI program that only uses Isomalloc works, but not if a user writes a custom PUP routine?

evan-charmworks:
> > The verification failures are due to a problem we didn't foresee. If pup_buffer transfers don't complete inline, any changes made to the buffers inside PUP methods will be overwritten by the deferred transfer after the PUP methods are finished. This means that any unpacking onto the stack in AMPI_Register_pup methods will not be seen by the user program and stale data will be there instead, such as pointers that are invalid after migration. This can be reproduced with examples/ampi/Cjacobi3D: `./charmrun +p2 ./jacobi 2 1 1 40 +vp2 +balancer RotateLB ++local ++debug`.
> 
> You're saying the Isomalloc'ed ULT stacks are corrupt if a ULT migrates across node when the user is writing their own PUP routines using AMPI_Register_pup, right? So an AMPI program that only uses Isomalloc works, but not if a user writes a custom PUP routine?

Yes. The ULT stacks match what they were on the previous node and don't reflect any changes made in the custom PUP routine. The problem only happens with custom PUP routines.

stwhite91:
> Yes. The ULT stacks match what they were on the previous node and don't reflect any changes made in the custom PUP routine. The problem only happens with custom PUP routines.

We don't have any current users that depend on AMPI_Register_pup that I know of, so I think we should be willing to rethink the API. Obviously PUP can theoretically be used to minimize the amount of memory you migrate, and should work with AMPI, but our current users are all depend on Isomalloc, and generally the only use of PUP/aboutToMigrate/justMigrated in AMPI is to be able to close and reopen file descriptors before and after migration. We should bring this up in our next all-hands meeting and/or Core.

evan-charmworks:
I am able to reproduce the issue when passing the build-time arguments that CI does. For example:

`./build AMPI netlrts-darwin-x86_64 -j4 --suffix=production-no-optimize --with-production --enable-error-checking --enable-lbuserdata -g3 -no-optimize -no-production && cd netlrts-darwin-x86_64-production-no-optimize/examples/ampi/Cjacobi3D/ && make OPTS="-g3" -j4 && ../../../bin/testrun +p2 ./jacobi.iso 2 2 2 40 +vp8 +balancer RotateLB +LBDebug 1 ++local ++debug-no-pause`

This is the stack trace:

```lldb
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xcc)
  * frame #0: 0x0000000100051697 jacobi.iso`ampi::resumeThreadIfReady(this=0x0000000101809400) at ampiimpl.h:2664:30
    frame #1: 0x00000001000511be jacobi.iso`ampi::generic(this=0x0000000101809400, msg=0x0000000100f0d1b0) at ampi.C:2651:3
    frame #2: 0x0000000100051089 jacobi.iso`ampi::genericSync(this=0x0000000101809400, msg=0x0000000100f0d1b0) at ampi.C:2621:3
    frame #3: 0x00000001000a6ffd jacobi.iso`ampi::sendLocalMsg(this=0x0000000101801e00, tag=2, srcRank=4, buf=0x00003fff70950bf8, size=80000, type=0, count=10000, destRank=6, destComm=9000000, seq=11, destPtr=0x0000000101809400, sendType=I_SEND, reqIdx=8) at ampi.C:3103:14
    frame #4: 0x0000000100053b48 jacobi.iso`ampi::delesend(this=0x0000000101801e00, t=2, sRank=4, buf=0x00003fff70950bf8, count=10000, type=0, rank=6, destcomm=9000000, arrProxy=CProxy_ampi @ 0x00003fff7010fc98, sendType=I_SEND, reqIdx=-1) at ampi.C:3168:12
    frame #5: 0x00000001000538fa jacobi.iso`ampi::send(this=0x0000000101801e00, t=2, sRank=4, buf=0x00003fff70950bf8, count=10000, type=0, rank=6, destcomm=9000000, sendType=I_SEND, reqIdx=-1) at ampi.C:3015:21
    frame #6: 0x000000010006888a jacobi.iso`::MPI_Isend(buf=0x00003fff70950bf8, count=10000, type=0, dest=6, tag=2, comm=9000000, request=0x00003fff7010fef0) at ampi.C:7518:19
    frame #7: 0x0000000100001cd0 jacobi.iso`::AMPI_Main_cpp(ac=5, av=0x00003fff700003c0) at jacobi.C:176:5
    frame #8: 0x0000000100042827 jacobi.iso`::AMPI_threadstart(data=0x0000000100e01f70) at ampi.C:1096:10
    frame #9: 0x000000010000338e jacobi.iso`startTCharmThread(msg=0x0000000100e01f50) at tcharm.C:163:2
    frame #10: 0x000000010029544c jacobi.iso`::CthStartThread(arg=(fctx = 0x00007ffeefbff590, data = 0x0000000100e036d8)) at threads.C:1836:3
    frame #11: 0x000000010029432f jacobi.iso`trampoline + 3

(lldb) p this
(ampi *) $0 = 0x0000000101809400
(lldb) p this->parent
(ampiParent *) $1 = 0x0000000000000000
(lldb) p this->thread
(TCharm *) $2 = 0x0000000000000000

(lldb) f 3

(lldb) p this
(ampi *) $3 = 0x0000000101801e00
(lldb) p this->parent
(ampiParent *) $4 = 0x0000000101802800
(lldb) p this->thread
(TCharm *) $5 = 0x0000000100f0ed70
(lldb) p destPtr
(ampi *) $6 = 0x0000000101809400
(lldb) p destPtr->parent
(ampiParent *) $7 = 0x0000000000000000
(lldb) p destPtr->thread
(TCharm *) $8 = 0x0000000000000000
```

@stwhite91 `sendLocalMsg` is being called with a `ampi * destPtr` that still has null `parent` and `thread` pointers after migration.

stwhite91:
Interesting, I'll take a look at it

stwhite91:
I pushed a workaround for that issue, though the deeper problem of ZC PUPs not completing inline remains

stwhite91:
Needs another review

src/conv-core/isomalloc.C
```diff
@@ -1001,8 +1025,14 @@ struct isommap
...
   }
 
+  void EnableRDMA(int enable)
+  {
+    use_rdma = enable;
+  }
+
   // canonical data
   uint8_t * start, * end, * allocated_extent;
+  int use_rdma;
```
matthiasdiener, line 1035: defined as an int here, but used like a bool above (use_rdma{true}).

evan-charmworks:
Needs one more review

PR #2709, opened 2020-02-14
Implement CmiNodeReduce

evan-charmworks:


src/conv-core/convcore.C
```diff
@@ -2745,12 +2891,31 @@ void CmiHandleReductionMessage(void *msg) {
...
 
+void CmiHandleNodeReductionMessage(void *msg) {
+  CmiReduction *red = CmiGetNodeReduction(CmiGetRedID(msg));
+  if (red->numRemoteReceived == red->numChildren) red = CmiGetNodeReductionCreate(CmiGetRedID(msg), red->numChildren+4);
+  red->remoteData[red->numRemoteReceived++] = (char *)msg;
+  /*CmiPrintf("[%d] CmiNodeReduce::remote %hd\n",CmiMyNode(),red->seqID);*/
+  CmiSendNodeReduce(red);
+/*
+  CsvAccess(_nodereduce_msg_list)[CsvAccess(_nodereduce_received)++] = msg;
+  if (CsvAccess(_nodereduce_received) == CsvAccess(_nodereduce_num_children)) CmiSendNodeReduce();
```
epmikida: Is this left here intentionally?

evan-charmworks: It corresponds to commented code in the per-PE version. Would it be better to remove it, potentially from both locations?

epmikida: As someone who isn't intimately familiar with the code, the comment is just confusing because it's not explained as to why it's there, and why we do things the uncommented way vs. the commented way. If it's a useful alternative method to do the send the reduction, then I suppose we could leave it in, but commented saying what it's purpose is. Otherwise, I'd say remove it.

evan-charmworks: Removed

PR #2711, opened 2020-02-17
cmiArgDebugFlag and CMK_CCS_AVAILABLE fixes

nitbhat:


src/arch/util/machine-common-core.C
```diff
@@ -1505,15 +1505,9 @@ static void ConverseRunPE(int everReturn) {
...
- * a memory segmentation fault (bad memory access). Note, even
- * testing CpvInitialized(cmiArgDebugFlag) doesn't help to solve
- * this problem because the variable indicating whether cmiArgDebugFlag is
- * initialized or not is not initialized, thus possibly causing another
- * bad memory access. --Chao Mei
+ * to those functions (such as CmiPrintf).
  */
-  CpvInitialize(int, cmiArgDebugFlag);
-  CpvAccess(cmiArgDebugFlag) = 0;
+  cmiArgDebugFlag = 0;
```
jcphill: This is a race condition as each pe sets it to 0 and then possibly back to 1. In theory it could have an inconsistent value in the meantime.

evan-charmworks: Since `cmiArgDebugFlag` is now a global variable, there is no need to assign it to zero here.

src/conv-core/convcore.C
```diff
@@ -3818,7 +3818,7 @@ void ConverseCommonInit(char **argv)
   if (CmiGetArgFlagDesc(argv, "+cpd", "Used *only* in conjunction with parallel debugger"))
     {
 #if CMK_CCS_AVAILABLE
-      CpvAccess(cmiArgDebugFlag) = 1;
+      cmiArgDebugFlag = 1;
```
nitbhat: This is also called on each PE. (Another potential race condition). I'll push a fix for this soon.

nitbhat: Is this alright? It's basically multiple threads writing into the same variable. I see other global variables like `ccsRunning` also following the same pattern.

evan-charmworks: I think this is a data race by technicality, but in practice it will not cause any problems.

nitbhat: I was thinking that I can add an `if (CmiMyRank() == 0)` condition and then set the variable.

evan-charmworks: Could that cause a worse race condition where another rank could get past a use of `cmiArgDebugFlag` without it having been set to 1 first?

jcphill: The right solution is to set this once before multiple threads are launched.

nitbhat: Yes, that's a good idea. I've move that code that reads the variable from `ConverseCommonInit` (which runs on every PE) to `ConverseInit` (which runs on every node).

PR #2715, opened 2020-02-19
Revert CmiAbort segfaults in production builds

jcphill:
Partial revert of 8c85a3debc2254d68f8aa50c6ec0b1e302c86131, which
had set Cmi_truecrash=1 whenever CmiNumNodes()<=32.  This would
cause users to see a segfault when CmiAbort() was called due to
input errors, safety checks, etc., which they would interpret as
a bug in the application.  Now builds with CMK_OPTIMIZE=1 will
only segfault in CmiAbort() if ++debug or +truecrash are specified
by the user on the command line at run time.

### DIFF
--- a/src/arch/mpi/machine.C
+++ b/src/arch/mpi/machine.C
@@ -1464,8 +1464,8 @@ void LrtsInit(int *argc, char ***argv, int *numNodes, int *myNodeID) {
         Cmi_truecrash = 1;
 #endif
         int debug = CmiGetArgFlag(largv,"++debug");
-        if (CmiGetArgFlagDesc(*argv,"+truecrash","Do not install signal handlers") || debug ||
-            CmiNumNodes()<=32) Cmi_truecrash = 1;
+        if (CmiGetArgFlagDesc(*argv,"+truecrash","Do not install signal handlers") ||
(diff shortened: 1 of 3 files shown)


matthiasdiener:
I agree with this change. Additionally, applications might also want to switch to using `CkExit(1);` or so instead of `CkAbort();` to indicate input errors etc.

evan-charmworks:
Is there a significant drawback to generating core dumps unconditionally? You can always delete them if they are not necessary, but it can be hard to reproduce an issue that happens sporadically.

I agree that `CkExit(1)` would be a preferable semantic for cases where a core dump is unnecessary.

jcphill:
The drawback, in addition to leaving unnecessary core files around if they are actually written, is that it looks like the program crashed rather than aborted. The end user will have no use for the core files anyway since they aren't going to debug anything.

ericjbohm:
The downside is on very wide runs where getting a core dump for every
process ends up being a prohibitive amount of I/O.

evan-charmworks:
> The downside is on very wide runs where getting a core dump for every process ends up being a prohibitive amount of I/O.

That's what `CmiNumNodes() <= 32` avoids.

PR #2719, opened 2020-02-19
Update reserved words in the manual

epmikida:


stwhite91:
`python` should be listed as a reserved word but not an entry method attribute I think

matthiasdiener:
Out of curiosity, does it make sense to keep this list around?

We could also think of sorting this list alphabetically, and removing the empty line after each reserved  word.

epmikida:
> Out of curiosity, does it make sense to keep this list around?
> 
> We could also think of sorting this list alphabetically, and removing the empty line after each reserved word.

I was thinking of alphabetizing as well. Grouping by category (ie accelerator keywords) may also be another way to organize.

But your question about whether this is needed is also worth thinking about. It doesn't really provide that much utility, but at the same time when talking about a new language it seems pretty common to be able to look up what the reserved words are.

I also wonder if it could just be auto generated from the charmci code so it doesn't drift out of date?

PR #2720, opened 2020-02-19
Use CmiNodeReduceStruct in isomalloc_sync

evan-charmworks:

PR #2723, opened 2020-02-19
Enable isomalloc_sync by default

### LINKED ISSUE #1964: Enable +isomalloc_sync by default for AMPI
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1964*

---
If running an AMPI application, it's often necessary to pass the runtime option '+isomalloc_sync' to ensure correctness in the face of Address Space Layout Randomization (ASLR).
The current implementation of +isomalloc_sync requires support for a shared file system across nodes, which is common but not universal. It would be nicer if we enabled +isomalloc_sync runtime support by default (on certain builds/systems only?) and added an option to disable it.

### LINKED ISSUE #2043: Add +isomalloc_sync build-time options
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2043*

---
Add two options: one to run with '+isomalloc_sync' by default, another to disable the warning about it completely.
The warning is confusing to users who aren't using AMPI or ULTs at all, and for most systems we want it on by default anyway.

evan-charmworks:
Closes #1964
Closes #2043

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           command: ./build LIBS netlrts-linux-x86_64 -j3 --with-production --enable-error-checking --enable-lbuserdata -g -Werror=vla
       - run:
           name: Test
-          command: make -C netlrts-linux-x86_64/tmp -j3 all-test-tests all-test-examples OPTS="-g" &&  make -C netlrts-linux-x86_64/tmp test-tests test-examples TESTOPTS="++local +isomalloc_sync"
+          command: make -C netlrts-linux-x86_64/tmp -j3 all-test-tests all-test-examples OPTS="-g" &&  make -C netlrts-linux-x86_64/tmp test-tests test-examples TESTOPTS="++local"
  
   NetLRTS-smp:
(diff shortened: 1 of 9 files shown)


evan-charmworks:
Draft until #2720 is merged

matthiasdiener:
The command line is inverted and renamed, right?

evan-charmworks:
> The command line is inverted and renamed, right?

Yes, now there is `+no_isomalloc_sync`.

matthiasdiener:
> > The command line is inverted and renamed, right?
> 
> Yes, now there is `+no_isomalloc_sync`.

Ok, this could be clarified in the title. Not sure if it necessary to document it in the manual.

evan-charmworks:
> > > The command line is inverted and renamed, right?
> > 
> > 
> > Yes, now there is `+no_isomalloc_sync`.
> 
> Ok, this could be clarified in the title. Not sure if it necessary to document it in the manual.

I'll update the title when merging. If the sync fails we suggest the option to the user.

evan-charmworks:
Added documentation.

PR #2727, opened 2020-02-20
Charmrun: Expand ability to disable ASLR and do so by default

evan-charmworks:


matthiasdiener:
I'm not sure why this is using posix_spawn instead of fork/exec. Isn't the personality inherited on an exec call?

evan-charmworks:
> I'm not sure why this is using posix_spawn instead of fork/exec. Isn't the personality inherited on an exec call?

The `_POSIX_SPAWN_DISABLE_ASLR` flag is the only way to disable ASLR at runtime on macOS. I see no reason not to also enable posix_spawn on other platforms that support it, because no functionality is lost by using it instead of fork/exec.

matthiasdiener:
> > I'm not sure why this is using posix_spawn instead of fork/exec. Isn't the personality inherited on an exec call?
> 
> The `_POSIX_SPAWN_DISABLE_ASLR` flag is the only way to disable ASLR at runtime on macOS. I see no reason not to also enable posix_spawn on other platforms that support it, because no functionality is lost by using it instead of fork/exec.

Ok, but please add this as a comment. Is `_POSIX_SPAWN_DISABLE_ASLR` only used in macOS?

evan-charmworks:
> Is `_POSIX_SPAWN_DISABLE_ASLR` only used in macOS?

Yes.

matthiasdiener:
This looks good, but there seem to be some unrelated changes now.

evan-charmworks:
> This looks good, but there seem to be some unrelated changes now.

The extent of the changes to threads.C are adding `// If this crashes, disable ASLR.` comments while keeping line lengths under 72 characters through use of local variables.

PR #2728, opened 2020-02-21
Set up uFcontext on Windows

evan-charmworks:


src/arch/win/unix2nt_cc
```diff
@@ -178,52 +185,68 @@ do
...
 	-U*)
 		def=`echo $arg | awk -F@ '{print substr($1,3)}'`
 		CL="$CL -U$def"
+		ML="$ML -U$def"
 		;;
 	-I)
-		CL="$CL -I`cygpath -w $1`"
+		path="`cygpath -w $1`"
+		CL="$CL -I$path"
+		ML="$CL -I$path"
```
matthiasdiener: `CL`->`ML`. In bash, you can use `ML+=" -I..."` to avoid such an issue. (same below)

PR #2729, opened 2020-02-21
Set up Isomalloc mmap backend on Windows

evan-charmworks:


src/conv-core/isomalloc.C
```diff
@@ -168,6 +182,8 @@ static int init_map() { return 0; /*Isomalloc never works without mmap*/ }
 CpvStaticDeclare(int, zerofd); /*File descriptor for /dev/zero, for mmap*/
 #endif
 
+static constexpr void * const mmap_fail = (void *)~(uintptr_t)0;
```
matthiasdiener: the value of mmap_fail is different here from the value for _Win32. Is this intentional?

evan-charmworks: Yes, it varies by implementation. mmap returns -1, VirtualAlloc returns null.

matthiasdiener: Ok, can you add a brief comment there to explain this? Also, could this be `MAP_FAILED` instead of -1?

evan-charmworks: Yes to both, done

PR #2731, opened 2020-02-25
liveViz: Replace ASLR-unsafe callbacks

evan-charmworks:


rbuch:
Can you add more information to the commit message as to why these changes are useful?

PR #2734, opened 2020-02-26
Clean up uFcontext build process

evan-charmworks:
Fixes `-win64` builds building i386 assembly files


matthiasdiener:
Is this a followup from #2728?

evan-charmworks:
> Is this a followup from #2728?

Yes

src/util/boost-context/import.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
matthiasdiener: Can you add a comment here on when this file should be run? (I'm guessing when updating boost)

evan-charmworks: Done

evan-charmworks:
> There was a UCX build failure, but that was spurious?

It was the same issue that 32-bit `netlrts-linux` autobuilds are having, which is not related to this PR.

evan-charmworks:
Needs one more review

PR #2740, opened 2020-03-02
Update versioning guidance in configure.ac

rbuch:


rbuch:
> Reviewing other examples of `AC_INIT` (https://github.com/search?p=1&q=AC_INIT&type=Code), shouldnt this line be
> 
> ```
> AC_INIT([Charm++], [6.10.1], [charm@cs.illinois.edu], [charm], [https://charm.cs.illinois.edu/])
> ```
> 
> , i.e., using the real version number (with dots, without extra zero), using argument quotes (`[]`), and the https URL?
> 
> We only seem to use the version number from `AC_INIT` to create the `VERSION` file (`echo AC_PACKAGE_VERSION > ../include/VERSION`). The cmake build system puts `6.10.1` there already, and that seems fine.

Yeah, that seems like a good idea, I was merely following the precedent we had before in this change, but if we can have better version numbers in the code, then I think we should.

matthiasdiener:
> > Reviewing other examples of `AC_INIT` (https://github.com/search?p=1&q=AC_INIT&type=Code), shouldnt this line be
> > ```
> > AC_INIT([Charm++], [6.10.1], [charm@cs.illinois.edu], [charm], [https://charm.cs.illinois.edu/])
> > ```
> > 
> > 
> > , i.e., using the real version number (with dots, without extra zero), using argument quotes (`[]`), and the https URL?
> > We only seem to use the version number from `AC_INIT` to create the `VERSION` file (`echo AC_PACKAGE_VERSION > ../include/VERSION`). The cmake build system puts `6.10.1` there already, and that seems fine.
> 
> Yeah, that seems like a good idea, I was merely following the precedent we had before in this change, but if we can have better version numbers in the code, then I think we should.

I think the version from AC_INIT is also used for the `CHARM_VERSION` preprocessor macro, but I'm not sure if that is used anywhere else.

matthiasdiener:
> > Yeah, that seems like a good idea, I was merely following the precedent we had before in this change, but if we can have better version numbers in the code, then I think we should.
> 
> I think the version from AC_INIT is also used for the `CHARM_VERSION` preprocessor macro, but I'm not sure if that is used anywhere else.

It seems that `CHARM_VERSION` has been used outside Charm++ before (see e.g. here: https://charm.cs.illinois.edu/gerrit/#/c/namd/+/5281/5/src/common.C) so I suggest we do not change its format, ie, keep it as `AC_INIT([Charm++], [61001], [charm@cs.illinois.edu], [charm], [https://charm.cs.illinois.edu/])`. I have already adjusted the cmake build system to produce the same format as the current one.

src/scripts/configure.ac
```diff
@@ -1,8 +1,12 @@
-# Decimal representation of Charm++ release line and API revision
-# Increment the last 2 digits when making an API change
-# Increase the first three after branching for a release
+# Numerical representation of Charm++ release line and API revision
+# Essentially just the version number stripped of decimal points (e.g. 6.10.1 -> 61001)
```
matthiasdiener: Maybe add: “and extending the last two fields to 2 digits each.”

PR #2737, opened 2020-02-27
Update release notes and version for 6.10.1

rbuch:

PR #2741, opened 2020-03-02
Update release notes and version for 6.10.1 (#2737)

rbuch:
This merely duplicates the change just merged into master but now for the `charm-6.10.1` branch.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,16 @@ This file describes the most significant changes. For more detail, use
 'git log' on a clone of the charm repository.
 
 ================================================================================
+What's new in Charm++ 6.10.1
+================================================================================
+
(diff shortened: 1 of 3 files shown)


rbuch:
> Fine, imho, but having the same PR twice seems to be slightly dangerous (e.g., if the present PR diverges from the PR in master). This is due to the branch protection update a few days ago, I suppose? We might want to look at a different option for future releases.

Yeah, this is specifically due to the branch protection on release branches. Perhaps we should relax the protections a bit (maybe just prevent force pushes or restrict the set of users allowed to push to a small set of release masters or something). I'm open to suggestions, but I think we should at the very least have some kind of protection for public facing release branches.

PR #2744, opened 2020-03-03
Clean up CmiMemoryAtomic and CMK_PCQUEUE_LOCK

evan-charmworks:
Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.

Adds C++11 and C11 fallbacks for CmiMemoryRead/WriteFence, but keeps GNU intrinsics as the first choice when available.

Fixes `Charm++ warning> fences and atomic operations not available in native assembly` on Windows and POWER.


matthiasdiener:
> Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.

Whats the reason for keeping all three choices available?

evan-charmworks:
> > Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.
> 
> Whats the reason for keeping all three choices available (ie, why not just use C++11 atomics always)?

`converse.h` can be included by C programs such as our Converse tests. If we used C++11 atomics always, there would be no reason to have these macros at all instead of using `std::atomic` directly.

evan-charmworks:
> > > Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.
> > 
> > 
> > Whats the reason for keeping all three choices available (ie, why not just use C++11 atomics always)?
> 
> `converse.h` can be included by C programs such as our Converse tests. If we used C++11 atomics always, there would be no reason to have these macros at all instead of using `std::atomic` directly.

I forgot about one thing the macros provide, which is deactivating atomics entirely in non-SMP mode.

evan-charmworks:
I have tested this on {netlrts,mpi,multicore}-win-x86_64-smp and verified that it works.

matthiasdiener:
> > > Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.
> > 
> > 
> > Whats the reason for keeping all three choices available (ie, why not just use C++11 atomics always)?
> 
> `converse.h` can be included by C programs such as our Converse tests. If we used C++11 atomics always, there would be no reason to have these macros at all instead of using `std::atomic` directly.

Ok, my guess that this interface only exists since atomics in C++ are relatively new. Are these actually used in our Converse tests? A quick grep didn't show anything pertinent.

> I forgot about one thing the macros provide, which is deactivating atomics entirely in non-SMP mode.

Sure, but this could also be done when just supporting C++11 atomics?

evan-charmworks:
> > > > Improves portability and consistency, now using C++11 atomics as a first choice, before C11 atomics or GNU intrinsics.
> > > 
> > > 
> > > Whats the reason for keeping all three choices available (ie, why not just use C++11 atomics always)?
> > 
> > 
> > `converse.h` can be included by C programs such as our Converse tests. If we used C++11 atomics always, there would be no reason to have these macros at all instead of using `std::atomic` directly.
> 
> Ok, my guess was that this interface only exists since atomics in C++ are relatively new. Are these actually used in our Converse tests? A quick grep didn't show anything pertinent.

Yes, this interface predates the conversion of the Charm runtime to C++11 and by extension our ability to use `std::atomic`. I am not aware of any users of these macros in C code or outside the runtime itself.

> > I forgot about one thing the macros provide, which is deactivating atomics entirely in non-SMP mode.
> 
> Sure, but this could also be done when just supporting C++11 atomics?

Yes, it could. My intent with this PR is to clean up the existing infrastructure that the original author created, moving it toward its "best form". I am open to removing the other backends but that may be best suited for a separate PR, potentially after further discussion.

PR #2736, opened 2020-02-27
AMPI: Implement proper context activation for migration callbacks

evan-charmworks:
Fixes #1214

PR #2545, opened 2019-09-30
AMPI: Link with -memory isomalloc and -module CommonLBs by default

evan-charmworks:

PR #2751, opened 2020-03-05
Fix tlsglobals migration on macOS

evan-charmworks:
This commit implements the bookkeeping necessary for macOS to use
thread-local storage buffers as allocated by Isomalloc instead of
allocating and maintaining its own non-migratable buffers.


evan-charmworks:
#2545 and #2736 are both blocked on a verification failure that this PR fixes.

Issue #2747, opened 2020-03-04
Exhaustively document most runtime flags 


nitbhat:
I see that the [appendix](https://charm.readthedocs.io/en/latest/charm++/manual.html#running-charm-programs) doesn't have a lot of runtime flags. Notable ones that I was searching for are `+CmiSleepOnIlde` and `+CmiSpinOnIdle`. 

Exhaustively grepping for `CmiGetArg` through the source code and adding them to the documentation will make it easier for users to know about the runtime options that we provide.

PR #2753, opened 2020-03-06
Do not print warning message for +CmiSleepOnIdle when arg is passed

nitbhat:
Additionally, this commit also modifies the warning message printed
since the previous message was confusing.

### DIFF
--- a/src/ck-core/init.C
+++ b/src/ck-core/init.C
@@ -1641,7 +1641,8 @@ void _initCharm(int unused_argc, char **argv)
             if (!_Cmi_sleepOnIdle && !_Cmi_forceSpinOnIdle && num > CmiNumCores())
             {
               if (CmiMyPe() == 0)
-                CmiPrintf("\nCharm++> Warning: the number of SMP threads (%d) is greater than the number of physical cores (%d), so internally setting +CmiSleepOnIdle to make threads sleep while idling. Use +CmiSpinOnIdle to change this behavior and make the runtime system spin on message reception when idle, rather than sleeping.\n\n", num, CmiNumCores());
+                CmiPrintf("Charm++> Warning: Running with more SMP threads (%d) than physical cores (%d).\n"
+                          "         Use +CmiSleepOnIdle (default behavior) or +CmiSpinOnIdle to silence this message.\n", num, CmiNumCores());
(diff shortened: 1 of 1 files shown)


src/ck-core/init.C
```diff
@@ -1638,10 +1638,10 @@ void _initCharm(int unused_argc, char **argv)
             // XXX: Assuming uniformity of node size here
             num += num/CmiMyNodeSize();
 #endif
-            if (!_Cmi_forceSpinOnIdle && num > CmiNumCores())
+            if (!_Cmi_sleepOnIdle && !_Cmi_forceSpinOnIdle && num > CmiNumCores())
             {
               if (CmiMyPe() == 0)
-                CmiPrintf("\nCharm++> Warning: the number of SMP threads (%d) is greater than the number of physical cores (%d), so threads will sleep while idling. Use +CmiSpinOnIdle or +CmiSleepOnIdle to control this directly.\n\n", num, CmiNumCores());
+                CmiPrintf("\nCharm++> Warning: the number of SMP threads (%d) is greater than the number of physical cores (%d), so internally setting +CmiSleepOnIdle to make threads sleep while idling. Use +CmiSpinOnIdle to change this behavior and make the runtime system spin on message reception when idle, rather than sleeping\n\n", num, CmiNumCores());
```
matthiasdiener: Could add a '.' at the end of the string.

evan-charmworks:
I definitely support adding `!_Cmi_sleepOnIdle` to the if statement. I feel like the new message text is a bit wordy though, and could be made more concise. I don't think we need to fully explain these options in the warning text.

My suggestion:

```
"Charm++> Warning: Running with more SMP threads (%d) than physical cores (%d).\n"
"         Pass +CmiSleepOnIdle (default behavior) or +CmiSpinOnIdle to silence this message.\n"
```

This would match the message when running a Charm program without any provisioning arguments:
```
Charm++> No provisioning arguments specified. Running with a single PE.
         Use +auto-provision to fully subscribe resources or +p1 to silence this message.
```

nitbhat:
> Neither of these options is described in the manual, but this could be added in another PR.

I've added this issue (https://github.com/UIUC-PPL/charm/issues/2747) to have all/most runtime flags documented.

PR #2756, opened 2020-03-10
AMPI: first pass of AMPI-only build optimizations

stwhite91:
AMPI doesn't use Charm-MPI interop or message priorities, so disable them.

### DIFF
--- a/build
+++ b/build
@@ -863,6 +863,8 @@ then
       # strip "-only" suffix from AMPI target
       PROGRAM="${PROGRAM%-only}"
     fi
+    OPTS="$OPTS -DCMK_NO_INTEROP=1 -DCMK_NO_MSG_PRIOS=1 "
+    CONFIG_OPTS="$CONFIG_OPTS --with-prio-type=char "
 else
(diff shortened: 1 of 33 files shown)


stwhite91:
Build failure is in megacon on netlrts-smp

build
```diff
@@ -863,6 +863,8 @@ then
       # strip "-only" suffix from AMPI target
       PROGRAM="${PROGRAM%-only}"
     fi
+    OPTS="$OPTS -DCMK_NO_INTEROP=1 -DCMK_NO_MSG_PRIOS=1 "
+    CONFIG_OPTS="$CONFIG_OPTS --with-prio-type=char "
```
evan-charmworks: Maybe the `$OPTS` and `$CONFIG_OPTS` should be at the end so manually passed parameters can overwrite them.

stwhite91: Done

matthiasdiener:
Should  be replicated in the cmake build

stwhite91:
> Should be replicated in the cmake build

Done

Issue #2739, opened 2020-02-28
Recurring failures in AMPI migration due to invalid msgIdx


evan-charmworks:
CI and Autobuild occasionally fail in AMPI migration with an invalid `msgIdx` (which is -1 as far as observed so far), though it may only show in incomplete stack traces as an abort inside `CkDeliverMessageFree`. This likely originates from #2696.

evan-charmworks:
MPI Linux SMP on Travis CI:

```
../../../bin/testrun  ./pgm +p2 +vp2  +setcpuaffinity
Running on 2 processors:  ./pgm +vp2 +setcpuaffinity 
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 2  ./pgm +vp2 +setcpuaffinity 
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 2 processes, 1 worker threads (PEs) + 1 comm threads per process, 2 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: 4a0c8c2
Isomalloc> Synchronized global address space.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled. 
Charm++> Running on 1 hosts (1 sockets x 1 cores x 2 PUs = 2-way SMP)
Charm++> cpu topology info is gathered in 0.059 seconds.
Charm++> Warning: the number of SMP threads (4) is greater than the number of physical cores (2), so threads will sleep while idling. Use +CmiSpinOnIdle or +CmiSleepOnIdle to control this directly.
CharmLB> RandCentLB created.
[0] Testing: Testing...
[1] migrated from 1 to 0
[0] Testing: Testing...
[0] migrated from 0 to 1
[1] migrated from 0 to 1
[0] Testing: Testing...
register.h> CkRegisteredInfo<40,System> called with invalid index -1 (should be less than 22)
[0] Stack Traceback:
  [0:0] pgm 0x62670a CmiAbortHelper(char const*, char const*, char const*, int, int)
  [0:1] pgm 0x62682b 
  [0:2] pgm 0x55eef1 
  [0:3] pgm 0x559095 CkDeliverMessageFree
  [0:4] pgm 0x55b94a _processHandler(void*, CkCoreState*)
  [0:5] pgm 0x62b2c8 CsdScheduleForever
[...]

PR #2757, opened 2020-03-10
Fix nokeep & expedited bcasts

stwhite91:


src/ck-core/ck.C
```diff
@@ -548,13 +548,14 @@ void CkDeliverMessageFree(int epIdx,void *msg,void *obj)
 #if CMK_CHARMDEBUG
   CpdBeforeEp(epIdx, obj, msg);
 #endif    
+  const auto msgtype = UsrToEnv(msg)->getMsgtype();
```
epmikida: Why was this needed? The only reason I ask is because if it's needed due to AMPI modifying the message type, then wouldn't that cause a race?

src/ck-core/ck.C
```diff
@@ -548,12 +548,12 @@ void CkDeliverMessageFree(int epIdx,void *msg,void *obj)
 #if CMK_CHARMDEBUG
   CpdBeforeEp(epIdx, obj, msg);
 #endif    
-  const auto msgtype = UsrToEnv(msg)->getMsgtype();
+  const auto msgtype = (msg == NULL) ? LAST_CK_ENVELOPE_TYPE : UsrToEnv(msg)->getMsgtype();
```
epmikida: Why is it now possible for NULL messages to be coming through this code path?

stwhite91: I am seeing them come through during migration, for a constructor of an array element which doesn't take a message. This was identified in charm's megatest after the first commit of this patch.

src/ck-core/ck.C
stwhite91: It's needed because the message can be freed in the call(). I'm not sure how, but it appears this is a common pattern when handling nokeep messages elsewhere in ck-core. AMPI doesn't modify charm message types at all.

evan-charmworks:
I don't know if this fixes the recurring failures in #2739 because I had to re-run MPI Linux SMP to get it to pass.

Separately from that, Travis' Windows CI is having some issue that prevents the build from starting, independent of this PR.

evan-charmworks:
Even though there have been spurious failures since this was merged, none of them have been the AMPI issue, so looks like #2739 is indeed fixed.

PR #2764, opened 2020-03-12
add C++11 option for Cray compiler

matthiasdiener:
https://pubs.cray.com/content/S-2179/8.5/cray-c-and-c++-reference-manual-85/the-cray-compiling-environment


matthiasdiener:
This is untested.

PR #2766, opened 2020-03-12
CDash updates

matthiasdiener:
- Increase timeout for each test to 40 minutes.
- Split up building tests into tests/ examples/ benchmarks/ for better support with the cmake build system.
- Don't exit immediately on error in run_autobuild.sh (to make sure
  the build directory is cleaned up).
- More robust detection of whether the script was started by cron or not.

### DIFF
--- a/cdash/CTestTestfile.cmake
+++ b/cdash/CTestTestfile.cmake
@@ -1,4 +1,4 @@
-# Disable since that won't work in the cmake build:
+# Disabled since that won't work in the cmake build:
 # add_test(build-tests "make" "-C" "../tmp" "-j4" "all-test-tests" "all-test-examples" "all-test-benchmarks" "OPTS=\"-g\"")
 
 add_test(tests/ "make" "-C" "../tests" "test" "TESTOPTS=$ENV{AUTOBUILD_TEST_OPTS}")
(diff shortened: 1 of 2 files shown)


matthiasdiener:
Ready for review.

cdash/CTestTestfile.cmake
```diff
@@ -1,7 +1,9 @@
-add_test(build-tests "make" "-C" "../tmp" "-j4" "all-test-tests" "all-test-examples" "all-test-benchmarks" "OPTS=\"-g\"")
+# Disabled since that won't work in the cmake build:
+# add_test(build-tests "make" "-C" "../tmp" "-j4" "all-test-tests" "all-test-examples" "all-test-benchmarks" "OPTS=\"-g\"")
```
evan-charmworks: Would it work to add build-tests, build-examples, and build-benchmarks rules that can pass -j4 to keep build time down?

matthiasdiener: Yes, we can think about something like that. For the autobuild, this isn't very critical, since it is not time constrained.

evan-charmworks: Wouldn't it be important for machines with finite allocations of core-hours?

matthiasdiener: Maybe, but the total build time for a sequential build of tests/ examples/ and benchmarks/ is less than 10 mins from what I've seen. In any case, this won't matter for now, since we run `./build all-test` for these builds anyways.

matthiasdiener: I restored these by splitting up the building of tests into separate tests.

PR #2767, opened 2020-03-12
Restructure HAPI, fix crash when # of PEs < # of GPU devices

### LINKED ISSUE #2754: Running with fewer PEs than GPUs fails with floating point exception
This circumstance will likely only happen when debugging but:
Running ChaNGa on a node with multiple GPUs and requestion fewer PEs available GPUs will cause a floating point exception in initDeviceMapping() because "pes_per_gpu" rounds to 0 at line 823.

This bug was not in charm v6.9.0.

minitu:
Restructures HAPI with changes derived from #2644, for better code structure and ease of merging the referred PR.
Also fixes #2754, where a crash occurs when the number of PEs are less than the number of available GPU devices.


minitu:
`GPUManager` has been extracted out as a separate header file in `gpumanager.h`.
The only noteworthy change (aside from the structural/naming changes) is lines 166-191 of `hapi_impl.cpp`, to handle the case where the number of PEs is less than the number of available GPU devices.

minitu:
Passes compilation and (quick) testing.

PR #2771, opened 2020-03-17
Add support for using OpenMP workers as PEs instead of pthreads

evan-charmworks:


stwhite91:
The netlrts-smp build failed in AMPI-ROMIO's test to see if it can link an MPI code, and netlrts non-smp built successfully but aborted the first test with "unexpected call to exit by user program. Must use CkExit, not exit!"

stwhite91:
This patch is missing support for compiling and linking the runtime with `-fopenmp` (or whatever the compiler's OpenMP flag is).

PR #2772, opened 2020-03-17
Various cleanup relating to fsglobals work

evan-charmworks:
- Charmrun: Fix buffer overflow in local process launch on Windows
- unix2nt_cc: Recognize additional variants of -g
- Windows: Fix /usr/bin/link workaround with VS 2017 and newer
- Windows: Don't try to build ROMIO by default
- Windows: Avoid -fPIC and -fPIE unknown argument warnings
- unix2nt_cc: Add support for creating DLLs
- AMPI: Add $ORIGIN safeguard to pathstub.sh
- configure: Fix -rdynamic test when using clang
- src/libs/conv-libs/openmp_llvm: Fix "error: template with C linkage"

(Please rebase as a merge strategy.)

### DIFF
--- a/src/arch/common/cc-clang.sh
+++ b/src/arch/common/cc-clang.sh
@@ -19,6 +19,7 @@ if [ "$CMK_COMPILER" = "msvc" ]; then
   CMK_NATIVE_LDXX="$CMK_LDXX"
 fi
 
+CMK_PIC='' # empty string: will be reset to default by conv-config.sh
 CMK_PIE='-fPIE'
 
 CMK_COMPILER='clang'
--- a/src/arch/win/unix2nt_cc
+++ b/src/arch/win/unix2nt_cc
@@ -160,6 +160,7 @@ DOCOMPILE=''
 DOASSEMBLE=''
 DOLINK=''
 DODEPS=''
+SHARED=''
 OUTPUT=''
 CL_OPTS="$CL_COMMON"
 ML_OPTS="$ML_COMMON"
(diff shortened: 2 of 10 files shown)


src/arch/win/unix2nt_cc
```diff
@@ -206,11 +211,12 @@ do
 		CL_OPTS+=' -DNDEBUG -Ox'
 		ML_OPTS+=' -DNDEBUG'
 		;;
-	-g)
+	-g|-g1|-g2|-g3)
```
matthiasdiener: maybe write as `-g|-g?|-ggdb`. XLC supports up to -g9.

src/arch/common/cc-clang.sh
```diff
@@ -19,6 +19,7 @@ if [ "$CMK_COMPILER" = "msvc" ]; then
   CMK_NATIVE_LDXX="$CMK_LDXX"
 fi
 
+CMK_PIC='' # reset to default
```
matthiasdiener: could add here (and below) how this is reset. I believe it is in charmc?

src/arch/win/unix2nt_cc
evan-charmworks: Done

src/arch/common/cc-clang.sh
evan-charmworks: Done

evan-charmworks:
Needs one more review

PR #2775, opened 2020-03-19
Variable-sized message aggregation(TRAM)

VenkatSNarayanan:
Some of the formatting in the code probably needs some fixing, but I wanted to get some opinions on the actual code.


src/xlat-i/xi-grammar.tab.h
```diff
@@ -111,14 +111,15 @@ extern int yydebug;
...
     UNSIGNED = 323,
-    ACCEL = 324,
-    READWRITE = 325,
-    WRITEONLY = 326,
-    ACCELBLOCK = 327,
-    MEMCRITICAL = 328,
-    REDUCTIONTARGET = 329,
-    CASE = 330,
-    TYPENAME = 331
+    SIZET = 324,
```
stwhite91: We should also add 'bool' here

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -144,21 +175,25 @@ class MeshStreamer : public CBase_MeshStreamer<dtype, RouterType> {
...
-                    const void *dataItem, bool copyIndirectly = false);
+                    const DataItemHandle<dtype> *dataItem, bool copyIndirectly = false);
 
   void ctorHelper(int maxNumDataItemsBuffered, int numDimensions,
                   int *dimensionSizes, int bufferSize,
                   bool yieldFlag, double progressPeriodInMs);
 
 public:
-
+std::map<int,int> histogram;
```
stwhite91: Add a comment describing what the key/val pairs are

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -1217,71 +1110,73 @@ class ArrayMeshStreamer :
 #endif
 
     if (deliverInline && destinationPe == this->myIndex_) {
-      ArrayDataItem<dtype, itype>
-        packedDataItem(arrayIndex, this->myIndex_, dataItem);
-      localDeliver(packedDataItem);
+      size_t sz = PUP::size(const_cast<dtype&>(dataItem));
+      char* data = new char[sz];
+      PUP::toMemBuf(const_cast<dtype&>(dataItem),data, sz);
```
stwhite91: This looks to be a copy that can be avoided if we know the method data is an array of POD type, right?

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
```diff
@@ -1,17 +1,18 @@
...
   extern module completion;
-
   include "DataItemTypes.h";
 
-  template<class dtype>
-  message MeshStreamerMessage {
+  message MeshStreamerMessageV {
     int destinationPes[];
-    dtype dataItems[];
+    int sourcePes[];
```
stwhite91: Does this mean that TRAM messages are larger now for even fixed sized messages? Or do those use a different message?

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: That line is actually something I forgot about when cleaning it up, I'll get rid of it.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
VenkatSNarayanan: Yes, but note that the dtype in the old code actually would be something like a tuple<dtype, CkArrayIndex, int> in the new code, so the only extra item that is actually being added is the offset array.

stwhite91: Yeah but I still think that should be avoided since it needlessly bloats the amount of bytes sent over the network. I see two optimization possibilities here:
1. Add a flag to the message structure whose value is either the fixed contribution size (which tells you to not bother with 'offsets') or some special value saying they're variable-sized.
2. Shrink the type of 'offsets' from 'int' to 'uint16_t' (or whatever integral type is the smallest capable one), since you know the buffer is never larger than the compile-time constant 'tramBufferSize', IIRC.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -1016,9 +1066,14 @@ class ArrayMeshStreamer :
...
-      dtype dataItem;
-      PUP::fromMemBuf(dataItem,(void*)data,size);
-      misdeliveredItems[arrayId].push_back(dataItem);
+      if (!is_fixed_size<dtype>::value) {
+        dtype dataItem;
+        PUP::fromMemBuf(dataItem,(void*)data,size);
+        misdeliveredItems[arrayId].push_back(dataItem);
+      }
+      else {
+        misdeliveredItems[arrayId].push_back(const_cast<dtype&>(*reinterpret_cast<const dtype*>(data)));
```
stwhite91: This assumes that anything that is fixed_size is also PUPbytes or trivially serialized. I think fixed_size is either a misnomer or this needs to be changed. In general, we want to optimize TRAM for PUPbytes datatypes so that we can avoid the overheads of PUPing whenever possible, but I'm not sure if we've decided on changing the semantics for that or not.

src/xlat-i/xi-Entry.C
```diff
@@ -1094,7 +1094,7 @@ void Entry::genTramInstantiation(XStr& str) {
           << "    }\n"
           << "    CProxy_" << container->tramInstances[i].type.c_str() << " tramProxy =\n"
           << "    CProxy_" << container->tramInstances[i].type.c_str()
-          << "::ckNew(2, dims, gId, itemsPerBuffer, false, 10.0);\n"
+          << "::ckNew(1, dims, gId, itemsPerBuffer, false, 0.01);\n"
```
stwhite91: How was the period here selected? This is a significant change

VenkatSNarayanan: The 10.0 value was far too high, and was a remnant from some experiments. Ideally, this value should be set by the user(probably via Justin's attribute parameter change), not somewhere in the depths of charmxi. The same goes for the buffer size and associated fractions.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: Yes, fixed_size here would actually refer to PUPbytes types. I'll rename it.

src/xlat-i/xi-grammar.tab.h
VenkatSNarayanan: Resolved

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: The new PUPbytes implementation resolves this

examples/charm++/TRAM/Makefile
```diff
@@ -1,10 +1,6 @@
 -include ../../../include/conv-mach-opt.mak
 
-DIRS = \
-  randomAccessGroup \
-  randomAccessArray \
-  aggregateRandomAccessGroup \
-  aggregateRandomAccessArray \
+DIRS = aggregateRandomAccessArray
```
stwhite91, line 3: Why delete the non-aggregate example, and not refactor it to match the new API?

benchmarks/charm++/Makefile
```diff
@@ -5,7 +5,6 @@ DIRS = \
   migrate \
   taskSpawn \
   taskSpawnRecursive \
-  streamingAllToAll \
```
stwhite91, line 8: Why remove this from the list of benchmarks?

tests/charm++/varTRAM/vartest.C
```diff
@@ -0,0 +1,98 @@
...
+      CkPrintf("Usage: %s N Msgcount lambda [given: %d]\n",args->argv[0], args->argc);
+      CkExit();
+    }
+    delete args;
+  }
+  void startsum(int val) {
+    CkPrintf("Sum of generated data: %d\n",val);
+  }
+  void done(int val) {
+    CkPrintf("Sum of final data: %d\n",val);
```
stwhite91, line 51: If this is a test and not an example, it should be asserting correctness. We probably also want some timing code for performance checks.

tests/charm++/varTRAM/vartest.C
```diff
@@ -0,0 +1,98 @@
...
+    if (iter == (N-1)*msgcount) {
+      contribute(sizeof(int), &result2, CkReduction::sum_int, CkCallback(CkReductionTarget(main,done), mainProxy));
+    }
+  }
+  msg rand() {
+    //int length=lambda/4;
+    std::uniform_int_distribution<> gen1(lambda/8,3*lambda/8);
+    int length = gen1(mt);
+    std::uniform_int_distribution<> gen2(0,373);
+    std::vector<int> gener;
```
stwhite91, line 88: This is better as something like this:

std::vector<int> gener(length);
for (auto &g : gener) {
    g = gen2(mt);
}

tests/charm++/varTRAM/vartest.C
```diff
@@ -0,0 +1,98 @@
...
+    gener.reserve(gen1(mt));
+    for (int j=0;j!=length;++j) {
+      gener.push_back(gen2(mt));
+    }
+    return msg(gener);
+    //return msg(0);
```
stwhite91, line 94: remove

tests/charm++/varTRAM/Makefile
```diff
@@ -0,0 +1,24 @@
...
+	$(CHARMC) $(CHARMCFLAGS) -language charm++ -o vartest vartest.o -module NDMeshStreamer
+
+vartest.def.h: vartest.ci
+	$(CHARMC) $(CHARMCFLAGS) vartest.ci
+
+vartest.o: vartest.C vartest.def.h
+	$(CHARMC) $(CHARMCFLAGS) -c vartest.C
+
+test: vartest
+	$(call run, ./vartest +p4 4 100 100)
```
stwhite91, line 19: I think it would be better to test with overdecomposition

tests/charm++/varTRAM/Makefile
```diff
@@ -0,0 +1,24 @@
+-include ../../common.mk
+-include ../../../include/conv-mach-opt.mak
+CHARMC = ../../../bin/charmc $(OPTS)
+
+OBJS = vartest.o
+
+all: vartest
+
+vartest: vartest.o
+	$(CHARMC) $(CHARMCFLAGS) -language charm++ -o vartest vartest.o -module NDMeshStreamer
```
stwhite91, line 10: Is it necessary to link with '-module NDMeshStreamer' explicitly? I thought that was built in to charmc/charmxi when using [aggregate]

tests/charm++/varTRAM/vartest.ci
```diff
@@ -0,0 +1,15 @@
...
+  mainchare main {
+    entry main(CkArgMsg*);
+    entry [reductiontarget] void startsum(int);
+    entry [reductiontarget] void done(int);
+  };
+  array [1D] engine {
+    entry engine();
+    entry engine(int,int,int);
+    entry void simulate();
+    entry [aggregate(bufferSize : 2048, numDimensions : 2, thresholdNum : 1, thresholdDen : 2, cutoffNum : 1, cutoffDen : 2)] void ping(msg);
```
stwhite91, line 12: I think 'Numer' and 'Denom' would be more intuitive than 'Num' and 'Den'. I assumed Num meant Number and I didn't know what Den was until I read through the charmxi code changes. Maybe something to discuss in Core meeting.

stwhite91: Could 'msg' just be std::vector or does it have to be a message type?

stwhite91: Can you still combine aggregate(...) with other entry method attributes? Such as nokeep or inline or something.

benchmarks/charm++/Makefile
VenkatSNarayanan: It's one of the benchmarks that is currently incompatible with the new API. I need to refactor it, but it breaks test builds if I leave it in(which could mask other issues with the implementation).

examples/charm++/TRAM/Makefile
VenkatSNarayanan: The non-aggregate examples were there originally to test the non-aggregate API, which isn't supported any more. I could refactor them, but they would end up being near-identical to the the aggregate test.

tests/charm++/varTRAM/vartest.ci
VenkatSNarayanan: msg can be a std::vector, since it has a PUPer. I've tested, it is compatible with other entry method attributes.

stwhite91: Can you make it a std:vector then? It would be cleaner

tests/charm++/varTRAM/Makefile
VenkatSNarayanan: This was a mistake that has since been fixed.

stwhite91:
Needs more review- @minitu and @evan-charmworks?

minitu:
I'll work on reviewing in the next couple of days.

minitu:
Gone through the code once and pushed some cleanup changes.
Will post my comments tomorrow or so.

benchmarks/charm++/Makefile
minitu: This should be documented so it can be brought back later.

examples/charm++/TRAM/Makefile
minitu: The discontinued support for the non-aggregate API should be documented, as the non-aggregate API is the one currently discussed in the documentation (there is no mention of the `[aggregate]` tag there by the way).

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
```diff
@@ -27,57 +29,21 @@ module NDMeshStreamer {
     entry void init(CkCallback endCb, int prio);
     entry [reductiontarget] void syncInit();
 
-    entry void receiveAtDestination(MeshStreamerMessage<dtype> *msg);
+    entry void receiveAtDestination(MeshStreamerMessageV *msg);
   };
 
   template<class dtype, class ClientType, class RouterType, int (*EntryMethod)(char *, void *) = defaultMeshStreamerDeliver<dtype,ClientType> >
-    group [migratable] GroupMeshStreamer : MeshStreamer<dtype, RouterType> {
```
minitu, line 34: This `GroupMeshStreamer` has the support for chare groups and should be revived later, right?

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -24,6 +28,12 @@
 
 #define TRAM_BROADCAST (-100)
 
+//#define TRAM_BUFFER_SIZE 20480
+
+//#define TRAM_SEND_CUTOFF_NUMERATOR 5
+//#define TRAM_SEND_CUTOFF_DENOMINATOR 10
+//#define TRAM_SEND_FRACTION_NUMERATOR 5
+//#define TRAM_SEND_FRACTION_DENOMINATOR 10
```
minitu, line 36: Are these needed anymore? If not, removing them would be better.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -144,21 +223,26 @@ class MeshStreamer : public CBase_MeshStreamer<dtype, RouterType> {
   bool useCompletionDetection_;
   CompletionDetector *detectorLocalObj_;
   virtual int copyDataItemIntoMessage(
-              MeshStreamerMessage<dtype> *destinationBuffer,
-              const void *dataItemHandle, bool copyIndirectly = false);
-  void insertData(const void *dataItemHandle, int destinationPe);
-  void broadcast(const void *dataItemHandle, int dimension,
```
minitu, line 150: Looks like you removed the broadcasting function, does this affect the usage of TRAM in any way? If so it should also be documented.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -144,21 +223,26 @@ class MeshStreamer : public CBase_MeshStreamer<dtype, RouterType> {
...
-                 bool copyIndirectly);
+              MeshStreamerMessageV *destinationBuffer,
+              const DataItemHandle<dtype> *dataItemHandle, bool copyIndirectly = false);
+  virtual int copyDataIntoMessage(
+              MeshStreamerMessageV *destinationBuffer,
+              char *dataHandle, size_t size, CkArrayIndex index);
+  void insertData(const DataItemHandle<dtype> *dataItemHandle, int destinationPe);
+  void storeMessageData(int destinationPe,
```
minitu, line 232: If I remember correctly, you said `storeMessageData` was used only at intermediate TRAM nodes unlike `storeMessage`. If this is true, can you rename it so that it reflects this difference?

src/xlat-i/xi-Entry.C
```diff
@@ -2419,34 +2487,34 @@ void Entry::setAccelParam(ParamList* apl) { accelParam = apl; }
...
+int Entry::isCreate(void) { return (hasAttribute(SCREATEHERE)) || (hasAttribute(SCREATEHOME)); }
+int Entry::isCreateHome(void) { return (hasAttribute(SCREATEHOME)); }
+int Entry::isCreateHere(void) { return (hasAttribute(SCREATEHERE)); }
+int Entry::isPython(void) { return (hasAttribute(SPYTHON)); }
+int Entry::isNoTrace(void) { return (hasAttribute(SNOTRACE)); }
+int Entry::isAppWork(void) { return (hasAttribute(SAPPWORK)); }
+int Entry::isNoKeep(void) { return (hasAttribute(SNOKEEP)); }
 int Entry::isSdag(void) { return (sdagCon != 0); }
-bool Entry::isTramTarget(void) { return (attribs & SAGGREGATE) != 0; }
-int Entry::isWhenIdle(void) { return attribs & (SWHENIDLE ^ SLOCAL); }
```
minitu, line 2443: Is it ok to remove the `SLOCAL` flag here?

tests/charm++/varTRAM/vartest.C
```diff
@@ -0,0 +1,101 @@
...
+        auto temp = rand();
+        thisProxy[(thisIndex+j+1)%N].ping(temp);
+        result1 = std::accumulate(temp.begin(),temp.end(),result1);
+      }
+    }
+    contribute(sizeof(int), &result1, CkReduction::sum_int, CkCallback(CkReductionTarget(main,startsum), mainProxy));
+  }
+
+  // After sending each message to the other PE,
+  // send a self-message to add scheduling overhead (enforce msgcount)
```
minitu, line 78: I didn't understand how this comment is implemented in the code. After `ping()` is invoked, each chare seems to just contribute back to main - there doesn't seem to be a self-message anywhere.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
VenkatSNarayanan: Yes.

src/xlat-i/xi-Entry.C
VenkatSNarayanan: Yes, I confirmed with Justin that the way the flags are set allows not checking SLOCAL there(there is a change made elsewhere to make this behavior correct).

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: I couldn't find uses of the broadcast function in the code actually. I think it's something that was exclusive to the non-[aggregate] API, so the new API effectively does not support broadcasts.

tests/charm++/varTRAM/vartest.C
VenkatSNarayanan: That comment was a mistake, and has been removed.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: Done.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
VenkatSNarayanan: Removed.

PR #2778, opened 2020-03-25
charmc: make sure to link against pthread when using CkLoop

matthiasdiener:
Should help with issues such as the one seen here: https://github.com/spack/spack/pull/15666 @s-sajid-ali

### DIFF
--- a/src/scripts/charmc
+++ b/src/scripts/charmc
@@ -1081,6 +1081,12 @@ findModuleDep() {
 	PROCESSED_MODULES="$PROCESSED_MODULES $M"
 	#PRE_LIBRARIES=`AddList $PRE_LIBRARIES -lmodule$M`
 	PRE_LIBRARIES=`AddList -lmodule$M $PRE_LIBRARIES`
+
+	if [[ $M == "CkLoop" ]]; then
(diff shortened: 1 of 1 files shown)


jcphill:
Does this work on Windows?

matthiasdiener:
> Does this work on Windows?

That's a good point. Is CkLoop even supported on Windows? We could also restrict this to non-SMP mode (when pthreads is actually used).

matthiasdiener:
> > Does this work on Windows?
> 
> That's a good point. Is CkLoop even supported on Windows? We could also restrict this to non-SMP mode (when pthreads is actually used).

Fixed in the updated patch.

evan-charmworks:
If you grep for pthread in src/arch, there is a lot of `-lpthread`. Should `-pthread` go there instead of charmc?

matthiasdiener:
> If you grep for pthread in src/arch, there is a lot of `-lpthread`. Should `-pthread` go there instead of charmc?

These seem to be mostly for SMP support. The issue here is that we need pthread for non-SMP mode, but only when using CkLoop. I think putting this in multiple src/arch files would cause lots of duplication there.

Issue #2732, opened 2020-02-25
UCX Linux CI frequently times out


evan-charmworks:
The GitHub CI build for UCX Linux regularly fails due to taking more time than the 60 minute timeout we have specified, which is already longer than the 45 minute timeout for other builds. When the timeout occurs, `make test-tests test-examples` has been running for 35 minutes, but when the build completes successfully, the test step only takes 11 minutes.

matthiasdiener:
Seems to hang in the `examples/charm++/load_balancing/stencil3d` test (see the timestamps on the left):

```
../../../../bin/testrun  +p4 ./stencil3d 64 32 +balancer DistributedLB +LBDebug 1  
2020-02-27T17:43:28.3268399Z 
2020-02-27T17:43:28.3269138Z Running on 4 processors:  ./stencil3d 64 32 +balancer DistributedLB +LBDebug 1 
2020-02-27T17:43:28.3326320Z charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./stencil3d 64 32 +balancer DistributedLB +LBDebug 1 
2020-02-27T17:43:28.3919884Z Charm++> Running in non-SMP mode: 4 processes (PEs)
2020-02-27T17:43:28.3920500Z Converse/Charm++ Commit ID: v6.10.0-rc2-158-g5cdb2cb97
2020-02-27T17:43:28.4054224Z Isomalloc> Synchronized global address space.
2020-02-27T17:43:28.4060522Z CharmLB> Verbose level 1, load balancing period: 0.5 seconds
2020-02-27T17:43:28.4060827Z CharmLB> Load balancer assumes all CPUs are same.
2020-02-27T17:43:28.4174003Z Charm++> Running on 1 hosts (1 sockets x 2 cores x 1 PUs = 2-way SMP)
2020-02-27T17:43:28.4613310Z Charm++> cpu topology info is gathered in 0.054 seconds.
2020-02-27T17:43:28.4692479Z CharmLB> DistributedLB created: threshold 0.000000, max phases 0
2020-02-27T17:43:28.4692540Z 
2020-02-27T17:43:28.4692789Z STENCIL COMPUTATION WITH BARRIERS
2020-02-27T17:43:28.4692908Z Running Stencil on 4 processors with (2, 2, 2) chares
2020-02-27T17:43:28.4692996Z Array Dimensions: 64 64 64
2020-02-27T17:43:28.4693148Z Block Dimensions: 32 32 32
[...]

PR #2784, opened 2020-03-26
Ensure a PE that is initially underloaded never tries to transfer load.

epmikida:
With the new multiphase approach, as load thresholds change, it was possible
before that a PE which had accepted load became "overloaded" at a certain
threshold and would try to transfer load. This behavior could cause crashes
and/or hangs under the right circumstances and was never intended.

### DIFF
--- a/src/ck-ldb/DistributedLB.C
+++ b/src/ck-ldb/DistributedLB.C
@@ -156,11 +156,14 @@ void DistributedLB::LoadReduction(CkReductionMsg* redn_msg) {
   // can receive more work. So assuming there exists an overloaded PE that can
   // donate work, I will start gossipping my load information.
   if (my_load < transfer_threshold) {
+    underloaded = true;
 		double r_loads[1];
 		int r_pe_no[1];
     r_loads[0] = my_load;
     r_pe_no[0] = CkMyPe();
     GossipLoadInfo(CkMyPe(), 1, r_pe_no, r_loads);
+  } else {
+    underloaded = false;
   }
 
   // Start quiescence detection at PE 0.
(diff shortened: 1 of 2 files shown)


epmikida:
This is related to #2732, and fixes the crashes that occurred when using Kavitha's method for reproducing the issue. I also ran it on a long loop and couldn't get the hang to show up, so it may fix that as well too. The hang could have been due to underloaded PEs repeatedly transferring back and forth when they should never be transferring in the first place.

rbuch:
This looks mostly good, but does it impact the quality of the load balancing decisions? Perhaps an underloaded processor will try to move a small object off of itself when it gets a larger object from a different processor, which might make the final solution more equal.

If that indeed is the case, then perhaps we can instead check if there are any self migrates at the end and remove those from the migration.

This might not be a significant point at all, I'm not sure.

epmikida:
> This looks mostly good, but does it impact the quality of the load balancing decisions? Perhaps an underloaded processor will try to move a small object off of itself when it gets a larger object from a different processor, which might make the final solution more equal.
> 
> If that indeed is the case, then perhaps we can instead check if there are any self migrates at the end and remove those from the migration.
> 
> This might not be a significant point at all, I'm not sure.

It should not affect quality at least from the perspective of the original algorithm. That algorithm did not allow for migrations off of an underloaded PE. I think the only reason we were running into this bug is because the algorithm allows for some very small tolerance for a PE to be slightly overloaded. But then with multiple phases, we must have been trying to migrate off of slightly overloaded, which probably just made other PEs "slightly overloaded" and then rinse and repeat. There was never any way for a PE to become more significantly overloaded.

PR #2785, opened 2020-03-27
Update .clang-format: Change AccessModifierOffset to -2

rbuch:


epmikida:
I agree with the access modifier indentation change. For namespaces, where are those used in Charm++? I'm always used to seeing them unindented in Charm++ adjacent projects, but I'm not familiar with them too much in Charm++ itself. Also, does clang format touch .ci files at all?

You also mentioned in core that brace spacing was possibly inconsistent, but didn't change it here.

rbuch:
> I agree with the access modifier indentation change. For namespaces, where are those used in Charm++? I'm always used to seeing them unindented in Charm++ adjacent projects, but I'm not familiar with them too much in Charm++ itself. Also, does clang format touch .ci files at all?
> 
> You also mentioned in core that brace spacing was possibly inconsistent, but didn't change it here.

They're used sporadically in Charm++; changing them to indented is mostly personal preference, but I think that's contrary to the prevailing style in Charm++ and most other places, so I'll remove that from the patch for now.

I didn't change the brace style because we've already explicitly changed it from Attach to Allman here: https://charm.cs.illinois.edu/gerrit/c/charm/+/4309. I'm in favor of Allman, but most of the code in the codebase I've seen is closer to Attach.

rbuch:
The access modifier change essentially changes from this:
```
class Example
{
 public:
  void example() {}
}
```
to this:
```
class Example
{
public:
  void example() {}
}
```
(note that `public` is preceded by a space in the first example and is not in the second)

epmikida:
> > I agree with the access modifier indentation change. For namespaces, where are those used in Charm++? I'm always used to seeing them unindented in Charm++ adjacent projects, but I'm not familiar with them too much in Charm++ itself. Also, does clang format touch .ci files at all?
> > You also mentioned in core that brace spacing was possibly inconsistent, but didn't change it here.
> 
> They're used sporadically in Charm++; changing them to indented is mostly personal preference, but I think that's contrary to the prevailing style in Charm++ and most other places, so I'll remove that from the patch for now.
> 
> I didn't change the brace style because we've already explicitly changed it from Attach to Allman here: https://charm.cs.illinois.edu/gerrit/c/charm/+/4309. I'm in favor of Allman, but most of the code in the codebase I've seen is closer to Attach.

Ah yeah, I'm much more partial to Attach, and I thought that's what we used tbh. Most of the actually formatted code I run into uses Attach, so I've been matching that, and even updating some unformatted code to that here and there in PRs that touched that code.

PR #2786, opened 2020-03-30
Update hwloc to v2.2.0

evan-charmworks:


evan-charmworks:
Needs one more review

Issue #2042, opened 2019-01-03
Add [whenidle] to simplify speculative execution


ericjbohm:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2042*

---
There are various cases where some computation can be done in the absence of other work, but would be of a low priority, or speculative.  In such cases it might be desirable to have an entry method triggered by a CcdIdle event which could perform the speculative work if there is nothing better to do.  this can currently be manually implemented by setting a Ccd and having the callback queue itself again via CCD, but this involves some moderately ugly code that could be trivially generated automatically.  So the only new aspect to this is that it would reduce the burden on the developer to implement this and the activity would be more naturally traced and visible in the Charm++ parallelization instead of hidden in chain of callbacks that don't even have to be entry methods.

Several refinements could be applied here, possibly as a parameterized elements such as whether the ide callback should only be called once, or every time sustained idle is detected.  Furthermore, additional parameterization could set a threshold for how much idleness is necessary to trigger the event,

stwhite91:
*Original date: 2019-01-03 18:21:17*

---
If you have an example of the ugly Ccd code to do this, could you post it here? I anticipate needing to implement this myself soon

ericjbohm:
*Original date: 2019-01-03 18:38:46*

---
Basic pattern (without templates or dependencies on more than C based trickery) is:

```
static void idleProgress(void *myObj, double time){
  MyClass *m= static_cast<MyClass *>(myObj);
  if(m->speculativeWork()) m->registerIdleProgressFunction();
}

void MyClass::registerIdleProgressFunction()
{
  CcdCallOnCondition(CcdPROCESSOR_STILL_IDLE, &idleProgress, (void *) this);
}
```

In this scheme, the speculativeWork function returns a boolean to indicate that is worth checking for speculative work on future idle invocations.   That allows us to avoid future overhead of checking if we know for certain within the scope of the speculativeWork function that there will be nothing more to do.  The application can re-register if that changes.

Wingpad:
Catching up to speed here, can you verify that my understanding of this is correct? The `[whenidle]` property will register a `PROCESSOR_STILL_IDLE` callback for the Chare that invokes any `[whenidle]` entry method(s) when it is triggered and then automatically requeue itself. Do we expect there to be conditions under which it wouldn't automatically requeue itself? Like, will the entry method be required to return a boolean to indicate it has speculative work remaining?

[...]

PR #2787, opened 2020-03-30
#2042 added when idle mechanism

Wingpad:
#2042 This commit adds support for the `[whenidle]` attribute, and includes an example for its use.


Wingpad:
> Can you document this new feature in the manual? Otherwise it looks good

Added a blurb to the section about entry method attributes, can add more if it is insufficient.

PR #2789, opened 2020-04-01
Support IBM Power architecture with UCX

minitu:
Closes #2788.

`cc-gcc.*` and `cc-xlc.*` were copied from `pamilrts-linux-ppc64le`, and `conv-mach*` files from `ucx-linux-x86_64`.

Builds successfully with `./build charm++ ucx-linux-ppc64le smp -j --with-production`, but tests fail with the following error:
```
[0] Stack Traceback:
------- Partition 0 Processor 0 Exiting: Called CmiAbort ------
Reason: UCX: UcxInitEps: runtime_kvs_put error failed: 5
  [0:0] verify 0x101e8454 CmiAbort
  [0:1] verify 0x101f4c00 LrtsInit(int*, char***, int*, int*)
  [0:2] verify 0x101f80c8 ConverseInit
  [0:3] verify 0x100af980 charm_main
  [0:4] verify 0x10098e98 main
  [0:5] libc.so.6 0x200000885200
  [0:6] libc.so.6 0x2000008853f4 __libc_start_main
------- Partition 0 Processor 0 Exiting: Called CmiAbort ------
Reason: UCX: UcxInitEps: runtime_kvs_put error failed: 5
[0] Stack Traceback:
  [0:0] verify 0x101e8454 CmiAbort
  [0:1] verify 0x101f4c00 LrtsInit(int*, char***, int*, int*)
  [0:2] verify 0x101f80c8 ConverseInit
  [0:3] verify 0x100af980 charm_main
  [0:4] verify 0x10098e98 main
  [0:5] libc.so.6 0x200000885200
  [0:6] libc.so.6 0x2000008853f4 __libc_start_main
```

PR #2790, opened 2020-04-03
Enable inter-node direct GPU messaging

minitu:
Follow-up to #2644.
Depends on #2789 for buildilng UCX layer on OLCF Summit and LLNL Lassen.

This patch aims to integrate GPUDirect RDMA to accelerate messaging between chares located on different physical nodes. 
First target machine layers are UCX and PAMI.


minitu:
Closed and reopened as #2791, a draft PR.

PR #2791, opened 2020-04-03
Enable inter-node direct GPU messaging

minitu:
Follow-up to #2644.
Depends on #2789 for buildilng UCX layer on OLCF Summit and LLNL Lassen.

This patch aims to integrate GPUDirect RDMA to accelerate messaging between chares located on different physical nodes.
First target machine layers are UCX and PAMI.


minitu:
Build command: `./build charm++ ucx-linux-ppc64le smp ompipmix -j --with-production --basedir=$HOME/openmpi-4.0.1/install`

minitu:
Commits up to here implements `Ck/CmiDeviceBuffer`, which replaces `Ck/CmiNcpyBuffer` & `CkSendBuffer` to separate device-related code from the host code. The memory registration/deregistration errors on the UCX layer actually occurred because `CmiPeOnSamePhysicalNode` didn't work properly and made the code take the (uncompleted) inter-node path instead of inter-process (intra-node). This is because CPU topology initialization was skipped to bypass the `CmiBarrier` hang, so looks like determining the physical node topology is one of its key functions.

The `CmiBarrier` hang needs to eventually be fixed, but not using it seems to work, so doing that for now.

minitu:
Confirmed support for LB on LLNL Lassen.
Still need to comment out CmiBarrier in `ck-core/init.C` to avoid hang.

Build: `./build charm++ ucx-linux-ppc64le smp cuda ompipmix -j --with-production --basedir=$HOME/openmpi-4.0.1/install -DCMK_GLOBAL_LOCATION_UPDATE`

Intra-node: `jsrun -n2 -a1 -c2 -g1 -K1 -r2 --smpiargs="-disable_gpu_hooks" ./verify -s 1024 -i 100 -l +ppn 1 +pemap L0,80 +commap L4,84 +gpumap block +balancer RotateLB +LBDebug 1`
Inter-node: `jsrun -n2 -a1 -c2 -g1 -K1 -r1 --smpiargs="-disable_gpu_hooks" ./verify -s 1024 -i 100 -l +ppn 1 +pemap L0 +commap L4 +gpumap block +balancer RotateLB +LBDebug 1`

minitu:
Trying to fix hang with `CmiBarrier` after CPU topology init in `ck-core/init.C`, since removing this barrier sometimes causes the program to crash since IPC handles can be opened before they are created.

CMakeLists.txt
```diff
@@ -846,7 +846,11 @@ if(${TARGET} STREQUAL "charm4py")
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
   add_library(charm SHARED empty.cpp)
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib_so)
-  target_link_libraries(charm ck converse memory-default threads-default ldb-rand "-Llib/ -standalone -whole-archive -c++stl -shared")
+  target_link_libraries(charm ck converse memory-default threads-default ldb-rand "-L/ccs/home/jchoi/work/ucx-1.9.0/install/lib -L/ccs/home/jchoi/work/pmix-3.1.5/install/lib -L/sw/summit/gdrcopy/2.0/lib64 -Llib/ -lpmix -lucp -lucm -lucs -luct -lgdrapi -standalone -whole-archive -c++stl -shared")
```
evan-charmworks, line 849: I don't think the changes to this line are something we should merge. The `-L` additions at the very least shouldn't be here. They should be inherited from whatever specifies these when linking with the UCX layer normally. I'm not sure what the best solution for the list of PMIx and UCX libraries would be. Maybe they should only be added to this sequence if building with UCX and CUDA.

src/conv-core/conv-rdmadevice.h
```diff
@@ -20,24 +21,38 @@ class CmiDeviceBuffer {
...
+  //CmiDeviceBuffer() : ptr(NULL), cnt(0), src_pe(-1), dest_pe(-1) { init(); }
+  CmiDeviceBuffer() : ptr(NULL), cnt(0) {}
 
+  /*
   explicit CmiDeviceBuffer(const void* ptr_, size_t cnt_) : ptr(ptr_), cnt(cnt_),
     src_pe(CmiMyPe()), dest_pe(-1) { init(); }
+    */
+  explicit CmiDeviceBuffer(const void* ptr_, size_t cnt_) : ptr(ptr_), cnt(cnt_) {}
 
+  /*
```
evan-charmworks, line 55: There are a few places in this PR, including here, where existing code is commented out. I have two suggestions for this:

1. If the content of a comment is entirely code that would otherwise compile correctly, I prefer to surround the disabled section with `#if 0` and `#endif` because this reserves the semantic of `/* block comments */` for actual comments, and also because instances of `#if 0` can be nested while block comments cannot (the closing `*/` of the inner comment will be interpreted as closing the outer one).

2. If different sections being commented out are all related by some common meaning, such as how this file has different uses of `src_pe` and `dest_pe`, instead of using a plain `#if 0` I prefer to mark them as conditional on a preprocessor definition that isn't actually defined anywhere. I'm not sure what exactly would be accurate for this case, but it could be something like `#ifdef CMIDEVICEBUFFER_OLDMETHOD` or `#ifdef CMIDEVICEBUFFER_USEIPC`.

evan-charmworks (commented):
I found the command `git diff origin/master...origin/jchoi/gpu-zerocopy-inter-node -w --color-moved -- CMakeLists.txt buildold cmake/ src/` helpful for viewing the changes in this PR other than the added tests and examples.

src/conv-core/conv-rdmadevice.h
evan-charmworks: If the commenting out is temporary since this PR is a draft, no need to go through the trouble of the second suggestion unless you want to. I would still recommend the first.

minitu:
@evan-charmworks Thanks Evan, this PR is still in the works and needs a lot of cleanup before it is in mergeable state. We're working on a paper submission so will clean this up after that.

PR #2793, opened 2020-04-06
Converse randomTTL and kNeighbors benchmarks 

nitbhat:


nitbhat:
@minitu You had mentioned previously that you saw hangs for kNeighbor in the SMP mode (on Summit). Could you share your run details?

minitu:
@nitbhat I only observe the hang with the PAMILRTS layer, with MPI layer it runs fine. The commands are:

minitu:
Something else that's strange is that the startup time (before first output of the program appears) with PAMILRTS is much longer than the MPI layer. The hang also occurs maybe 1 out of 10 times.

minitu:
Just tested the UCX layer and it works fine, similar to the MPI layer.

nitbhat:
> Something else that's strange is that the startup time (before first output of the program appears) with PAMILRTS is much longer than the MPI layer. The hang also occurs maybe 1 out of 10 times.

Thanks @minitu. 
That's strange. It might be a weird messaging race condition that probably rarely gets triggered with PAMILRTS.

minitu:
Just fixed segfault due to missing `CpvInitialize` in `randomttl` as well.

Testing `randomttl` with the same configurations as `kneighbors`, MPI and UCX layers work without any issues.
But PAMILRTS has the same slowdown and occasional hang issue.

Commands are as follows:
MPI & UCX: `jsrun -n4 -a1 -c20 -K1 -r2 ./randomttl 50 1000 +ppn 19 +pemap L4-76:4,84-156:4 +commap L0,80`
PAMILRTS: `jsrun -n4 -a1 -c20 -K1 -r2 ./randomttl 50 1000 +ppn 20 +pemap L0-76:4,80-156:4`

PR #2799, opened 2020-04-10
UCX: Polling perf improvement

brminich:
This is experimental patch trying to fix slowness in #2716 
Please do not merge it until verified in #2716 

The ideas is to shorten `ucp_tag_probe_nb` time, which may be quite big when:
- lots of small messages arriving
- probe is done with partial mask

In this case every probe call would go thru all unexpected messages list trying to find the one, matching the tag. But when probing with full tag mask, just one particular unexpected queue is examinated.


trquinn:
I did a cherry-pick of this PR on to v6.10.1, and tested it with ChaNGa on Frontera.  Here is the timeline for the domain decomposition of "dwf1.6144" on 2 nodes/112 cores.  Compare this with 
[TimelineScreenshot_new.pdf](https://github.com/UIUC-PPL/charm/files/4465126/TimelineScreenshot_new.pdf)
Compare this with [the timeline](https://github.com/UIUC-PPL/charm/files/4451964/TimelineScreenshot.pdf) posted in #2716, which takes an order of magnitude more total time.  The domain decomposition time is now comparable to the MPI build.
I also tested with the full dwf1.6144 benchmark out 40 nodes, with no issues.

I then tested a larger simulation (10x more particles) on 24, 32, 40, 48, and 64 nodes.  On 64 nodes there is occasionally a hang during load balancing.
So it appears this patch helps, but doesn't fix everything.  I'll continue to investigate.

brminich:
@trquinn, thanks a lot for checking the patch!
@nitbhat, is there a way to know the maximal possible size of `NcpyOperationInfo`?
To implement this patch cleanly, I have to make sure that all auxiliary messages (i.e. sending `NcpyOperationInfo`) are sent with eager protocol.

src/arch/ucx/machine.C
```diff
@@ -279,9 +280,9 @@ void LrtsInit(int *argc, char ***argv, int *numNodes, int *myNodeID)
     }
 
     // Any RNDV thresh is allowed
-    if (!CmiGetArgInt(*argv, "+ucx_rndv_thresh", &ucxCtx.eagerSize)) {
-        ucxCtx.eagerSize = UCX_MSG_PROBE_THRESH;
-    }
+    int thresh = 0;
+    CmiGetArgInt(*argv, "+ucx_rndv_thresh", &ucxCtx.eagerSize);
+    ucxCtx.eagerSize = std::max(UCX_MSG_PROBE_THRESH, thresh);
```
PhilMiller, line 285: This can't be right - `thresh` is never re-assigned from its initial `0`, the command line argument is now entirely ignored, and `ucxCtx.eagerSize` is always set to 32768.

PhilMiller:
```
  std::max(UCX_MSG_PROBE_THRESH, thresh);
= std::max(32768, 0);
= 32768
```

brminich: yes, you are right. Would need to `CmiGetArgInt(*argv, "+ucx_rndv_thresh", &thresh);`
Waiting for a an answer on maximal size of `NcpyOperationInfo` to implement clean patch. Will apply this comment as well then

nitbhat:
> @trquinn, thanks a lot for checking the patch!
> @nitbhat, is there a way to know the maximal possible size of `NcpyOperationInfo`?
> To implement this patch cleanly, I have to make sure that all auxiliary messages (i.e. sending `NcpyOperationInfo`) are sent with eager protocol.

Yeah, it shouldn't be too hard to add a thin LRTS API to determine the max size of the `NcpyOperationInfo` object. 

The dynamic fields only includes `CkCallback`, which is defined in the Charm++ layer. For that reason, I just need to add a minor API to return `sizeof(CkCallback)` from the charm layer.

PhilMiller:
Suggestion: have the Charm++ layer set `sizeof(CkCallback)` in a machine layer variable, maybe by calling some machine layer function, rather than creating a circular dependency by making a call upward.

Also, why should `sizeof(CkCallback)` ever be relevant in the machine layer? Is that so you can send one in a message to use as the 'target-side notification'? If it's initiator-side, it seems like a pointer or table index that can be returned to the caller suffices.

nitbhat:
> Suggestion: have the Charm++ layer set `sizeof(CkCallback)` in a machine layer variable, maybe by calling some machine layer function, rather than creating a circular dependency by making a call upward.

Yeah, that's definitely a cleaner implementation. 

> 
> Also, why should `sizeof(CkCallback)` ever be relevant in the machine layer? Is that so you can send one in a message to use as the 'target-side notification'? If it's initiator-side, it seems like a pointer or table index that can be returned to the caller suffices.

Yes, it's to avoid sending a separate callback message. By doing this, we can initiate a reverse operation (GET instead of a PUT     *or*    a PUT instead of a GET) and send the acknowledgement object (`CkCallback`) in one message.

PhilMiller:
Just because I've been away since before UCX stuff really got moving, and I'm curious:
- Does it implement PUT-with-notify directly, or as "tell the target to GET and then fire the included callback"?
- Is there a use-case for GET-with-notified-target?
Notification on the initiator as always seemed like the easy case.

nitbhat:
> Just because I've been away since before UCX stuff really got moving, and I'm curious:
> 
> * Does it implement PUT-with-notify directly, or as "tell the target to GET and then fire the included callback"?

So, the lower layer implementation of the Zcpy API supports all these kind of operations:
1. PUT with notify directly (everything on the initiator)
2. GET with notify directly (everything on the initiator)
3. Tell the target to PUT and then notify from target 
4. Tell the target to GET and then notify from target 

3 and 4 require the `CkCallback` object to be sent over and then called from the target. 

> * Is there a use-case for GET-with-notified-target?
>   Notification on the initiator as always seemed like the easy case.

The use case is always usually notification on the initiator. 

However, for certain specific conditions, it is required to implement the RDMA operation as a reverse operation i.e. the mode listed in points 3 and 4. 

These conditions include: 
1. On gni, GETs need the length, local_addr, and remote_addr to be 4-byte aligned. (For such cases, we revert to a remote PUT instead of a local GET). 

2. We support a user-controlled registration mode called `CK_BUFFER_UNREG` where the buffer is not registered until is it required to be registered (for an RDMA transfer). So, memcpy and CMA transfers can happen without registration. 

For these conditions, since I need to send a message to the remote side to implement a reverse operation, I am sending the `CkCallback` object as well so that it can fire a target PUT (instead of an initiator GET) or a target GET (instead of an initiator PUT) and then on completion, enqueue the callback. (*Note that at the machine layer level, RDMA GET/PUT completion is triggered only on the side that called the RDMA GET/PUT for all machine layers*)

nitbhat:
> @trquinn, thanks a lot for checking the patch!
> @nitbhat, is there a way to know the maximal possible size of `NcpyOperationInfo`?
> To implement this patch cleanly, I have to make sure that all auxiliary messages (i.e. sending `NcpyOperationInfo`) are sent with eager protocol.

@brminich I've added this PR (https://github.com/UIUC-PPL/charm/pull/2805) to get the max size of `NcpyOperationInfo` by calling `LrtsGetMaxNcpyOperationInfoSize()` from the machine layer. Let me know if you have any issues when using it.

brminich:
@nitbhat, thanks! I'll give it a try and will get back to you

brminich:
@nitbhat, updated the patch with `LrtsGetMaxNcpyOperationInfoSize()` usage. Can you please review?

src/arch/ucx/machine.C
```diff
@@ -614,8 +619,11 @@ void LrtsAdvanceCommunication(int whileidle)
     do {
        cnt = ucp_worker_progress(ucxCtx.worker);
 
+       // Probe with full tag mask to avoid long traversing thru unexpected
+       // queue of eager messages (messages with non-full mask added to the
+       // same unexpected queue)
        msg = ucp_tag_probe_nb(ucxCtx.worker, UCX_MSG_TAG_PROBE,
-                              UCX_MSG_TAG_MASK, 1, &info);
+                              UCX_MSG_TAG_MASK_FULL, 1, &info);
```
nitbhat, line 626: The overall change looks good. 

I have one question: Before this change, which messages were not being caught by this probe call? (i.e. messages with what tags were having to go through the unexpected queue?). I'm asking this because I see that the earlier tag `UCX_MSG_TAG_MASK` (00001111) was catching all messages except the UCX_RMA tags right?

brminich: @nitbhat, previously probe was done with non-full mask matching the messages sent with `PROBE` bit (0x2).  This bit is used for sending long messages (regardless whether it is auxiliary zcopy message or payload). The problem is with probing using non-full tag mask. In this case UCX will go thru the list of all unexpected messages before it finds the match (or does not find if none such messages). With Changa there were a lot of small messages sent with eager protocol (i.e. which are not mathcing the probe for long messages). But with every LrtsAdvance call UCX tried to find long message going thru all the list of small messages. 
Now, when we do probe for the full tag, UCX search for the message just in a single queue intended specifically for this  `PROBE` tag. And if there is no any long message arrived probe function will return immediately.

Also to probe with full mask I have to restrict auxilliary messages to be sent with eager protocol only, because I use some high bits of the tag marking the type of the message.

nitbhat: Oh okay, I understand. Thanks for explaining that. It will be useful to add some of this information in the commit message. Something like:

`This commit changes the ucp_tag_probe_nb call to probe with a full tag mask instead of a non-full tag mask. This is done in order to search for PROBE tag messages (used for large messages) in a separate queue and avoid iterating over the small message queues for every LrtsAdvanceCommunication call, hence optimizing receives, especially for applications with a large number of short messages.`

brminich:
@nitbhat, updated

evan-charmworks:
@matthiasdiener It looks like the GitHub CI is glitched for this PR.

matthiasdiener:
> @matthiasdiener It looks like the GitHub CI is glitched for this PR.

Yeah, they are having lots of trouble today: https://www.githubstatus.com/
I'll set the github build as optional.

PR #2801, opened 2020-04-11
Shrink the size of NcpyOperationInfo structs

stwhite91:
Rearranging member variables based on their natural
alignment requirements and shrinking some types decreases
the size of the struct from 152 to 112 bytes.

Use 'clang -cc1 -fdump-record-layouts src/util/cmirdmautils.C'
to see record layouts.


stwhite91:
I changed the following from 'unsigned short int' to 'unsigned char': srcRegMode, srcDeregMode, isSrcRegistered, destRegMode, destDeregMode, isDestRegistered.

And the following from 'int' to 'unsigned short int': srcLayerSize, srcAckSize, destLayerSize, destAckSize.

I am not sure that these are all safe transformations, and asserts should probably be added to check that, but I also wanted to ask if there are other structs that are sent over in metadata messages which we should be optimizing layout for similarly?

src/util/cmirdmautils.h
```diff
@@ -13,28 +13,28 @@ typedef struct ncpystruct{
...
-  unsigned short int srcDeregMode;
-  unsigned short int isSrcRegistered;
   const void *srcRef;
+  int srcPe;
+  int srcSize;
+  short int srcLayerSize;
+  short int srcAckSize;
+  unsigned char srcRegMode;
+  unsigned char srcDeregMode;
+  unsigned char isSrcRegistered;
```
nitbhat, line 25: Since we've moved to all C++ now, we can change `isSrcRegistered` and `isDestRegistered` to bools (for easier usage, but that can happen in a separate patch as well since that'll need machine layer changes).

Pes and size variables can be made unsigned too. (doesn't include rootNode though).

nitbhat:
Thanks for looking into this @stwhite91. 

Other structures that are sent in the metadata messages include `CkNcpyBuffer` and `CmiNcpyBuffer`.

src/util/cmirdmautils.h
stwhite91: PEs are signed ints everywhere else in the runtime, and size variables are also signed ints in most of the communication layers AFAIK, so making them unsigned here won't really get us anything. Changing to bool would be good for those that can be

src/ck-core/ckrdma.h
```diff
@@ -413,11 +410,6 @@ int getRootNode(envelope *env);
 // Function declaration for EM Ncpy Ack handler initialization
 void initEMNcpyAckHandler(void);
 
-struct zcPupIncompleteInfo {
```
stwhite91, line 416: This type was unused

evan-charmworks:
We should replace any uses of `bool` in structs that go over the network with `char`.

src/ck-core/ckrdma.h
nitbhat: Yeah, looks like this was added during the ZC Pup patch, but was never used.

nitbhat:
Looks like you need to include <limits> in cmitrdmautils.C

```
cmirdmautils.C:78:34: error: ‘std::numeric_limits’ has not been declared
   CmiAssert(srcLayerSize <= std::numeric_limits::max<unsigned short int>());
                                  ^
converse.h:1147:13: note: in definition of macro ‘CmiAssert’
   ((void) ((expr) ? 0 :                   \
             ^
```

PR #2802, opened 2020-04-13
cdash/ctest: Fix test directories on Windows

matthiasdiener:
On Windows systems, `./build` does not create the usual top-level `bin`, `tmp`, etc. symlinks, which ctest used to find the directory of tests to run. This patch changes this to determine the name of the directory where the `include/VERSION` file is located, and use that directory to locate the `tests/`, `examples/`, and `benchmarks/` directories. 

Note that this might fail when there are multiple builds within the same charm directory, as there will be multiple `include/VERSION` files. Since this is meant just for automated test builds, we can ignore this limitation for now.

### DIFF
--- a/cdash/CTestTestfile.cmake
+++ b/cdash/CTestTestfile.cmake
@@ -1,12 +1,17 @@
+# Determine where the tests/ examples/ benchmarks/ directories are.
+# Necessary for Windows, which does not have the usual tmp etc. symlinks.
+file(GLOB_RECURSE mydir FOLLOW_SYMLINKS  ../*/include/VERSION) # The '*' is the dir we want.
+get_filename_component(mydir ${mydir} DIRECTORY)
+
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Another option would be to create the same top-level symlinks (bin lib tmp ...) on windows as on the other architectures. I’m not sure why this wasn’t done originally.

matthiasdiener:
> Another option would be to create the same top-level symlinks (bin lib tmp ...) on windows as on the other architectures. I’m not sure why this wasn’t done originally.

The reasons for the lack of top-level symlinks seem to be:
- Our createlink.cpp creates Windows hard links, which are only supported for files and not directories.
- Cygwin supports Unix-like symlinks with `ln -s`, which appear to work fine but some create weird interactions with Windows binaries (e.g. `cl.exe`).

PR #2804, opened 2020-04-13
Fix broken BlockMap

epmikida:
By overriding populateInitial but not procNum, BlockMap was broken
and resulted in an influx of updateLocation messages upon array
creation. Adding a procNum function to fix this would have made it
equivalent to DefaultArrayMap.

Change-Id: Id3a53d7dc47d98d106eeda728c3d714bf723373d

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -940,55 +940,11 @@ public:
   }
 };
 
-class BlockMap : public RRMap
-{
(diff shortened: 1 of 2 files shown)


epmikida:
Array maps in general probably need some TLC, but this patch should at least be a quick fix for the bug seen in ChaNGa as part of issue #2716.

PR #2816, opened 2020-04-22
Clean up various uses of std::unordered_map

stwhite91:


evan-charmworks:
Is `std::make_pair` necessary? With C++11 perfect forwarding I thought emplace could take the two parts of the pair directly.

stwhite91:
> Is `std::make_pair` necessary? With C++11 perfect forwarding I thought emplace could take the two parts of the pair directly.

That's right, fixed in the latest commit

PR #2820, opened 2020-04-23
Ccd: Added support for CcdPROCESSOR_LONG_IDLE

nitbhat:


nitbhat:
@trquinn: As suggested by @lvkale, I have implemented this hook that you can use with the Ccd framework to call your function when the hang occurs. You can see the usage in the test that I have written in this PR. Make sure that you build charm/ChaNGa with `--enable-error-checking` to enable this feature. Additionally, you can use the runtime flag `+longIdleThresh` to supply the time (in seconds) that you would want the RTS to wait before triggering the LONG_IDLE condition, calling your supplied function. (default wait time right now is 10s)

src/conv-core/convcore.C
```diff
@@ -1632,12 +1663,22 @@ void CsdBeginIdle(void)
 #else
   CpvAccess(cmiMyPeIdle) = 1;
 #endif // CMK_SMP
+#if CMK_ERROR_CHECKING
+  CpvAccess(idleBeginWalltime) = CmiWallTimer();
```
stwhite91, line 1667: CcdCallBacks(), called above, itself calls CmiWallTimer(), as does CcdRaiseCondition(), called below. Both of these currently return void, but we could change them to return the time as a double, and remove the need to have new/extra calls to CmiWallTimer()

stwhite91: Of course, those times will be slightly stale by the time they are passed out of those routines, but it won't matter for this use case, and we can warn potential users of that in comments

nitbhat: That's a good point. In the latest patch, I've made CcdRaiseCondition return curWallTime and used it to set idleBeginWallTime.

src/conv-core/convcore.C
```diff
@@ -979,6 +984,10 @@ void CmiTimerInit(char **argv)
   CpvAccess(inittime_wallclock) = times(&temp);
   CpvAccess(inittime_virtual) = temp.tms_utime + temp.tms_stime;
   CpvAccess(clocktick) = 1.0 / (sysconf(_SC_CLK_TCK));
+
+#if CMK_ERROR_CHECKING
```
stwhite91, line 988: Is there some other place in convcore.C (or maybe in conv-conds.C) that we could initialize this once, rather than the 8 separate times we have here?

nitbhat: Yeah, I think I can move it to `CcdModuleInit` since it's used for Ccd functionality (instead of timers).

doc/converse/manual.rst
```diff
@@ -1926,6 +1926,17 @@ CcdPROCESSOR_STILL_IDLE
...
+CcdPROCESSOR_LONG_IDLE
+   This is an extension of CcdPROCESSOR_STILL_IDLE for a relatively longer
+   period of time. It is raised when the scheduler finds that it doesn't
+   have any messages to execute for a long period of time. The default
+   LONG_IDLE time is 10 seconds. However, it is customizable using a user
+   passed runtime flag ``+longIdleThresh <long idle time in seconds>``.
+   This feature is useful for debugging hangs in applications. It can allow
+   the user to add a user defined function as a hook to execute when the program
+   goes into a long idle state, typically seen during hangs. The test program
+   ``tests/charm++/longIdle`` illustrates the usage of CcdPROCESSOR_LONG_IDLE
```
stwhite91, line 1938: Should mention that this feature is only enabled for builds with error checking enabled.

nitbhat: Ah, right. Done.

src/conv-core/convcore.C
```diff
@@ -1632,12 +1637,22 @@ void CsdBeginIdle(void)
 #else
   CpvAccess(cmiMyPeIdle) = 1;
 #endif // CMK_SMP
-  CcdRaiseCondition(CcdPROCESSOR_BEGIN_IDLE) ;
+  double curWallTime = CcdRaiseCondition(CcdPROCESSOR_BEGIN_IDLE) ;
```
minitu, line 1640: Should we change the function signature or should we just have a separate CmiWallTimer call?
If we are going with the former, the Converse programming manual should also be updated.

nitbhat: Oh yeah, good point. I can update the converse programming manual. 

I think it's better to go ahead with CcdRaiseCondition returning the current wall time since it can save a couple of additional timer calls. (and esp since it was returning void before this).

doc/converse/manual.rst
```diff
@@ -2014,12 +2028,15 @@ callbacks from within ccd callbacks.
...
+  double CcdRaiseCondition(int condNum)
 
 
 When this function is called, it invokes all the functions whose
 pointers were registered for the ``condNum`` via a *prior* call to
-``CcdCallOnCondition`` or ``CcdCallOnConditionKeep``.
+``CcdCallOnCondition`` or ``CcdCallOnConditionKeep``. The function
+internally calls ``CmiWallTimer`` and returns this value. When using
+``CcdRaiseCondition``, the return value can be used to determine the
+current walltime avoiding an additional call to ``CmiWallTimer``.
```
stwhite91, line 2039: You should mention here that the wall time returned may be stale by the time it is returned, because any functions that were registered for that condition will be called between the timer call and the return. So users shouldn't depend on it being exact or current.

nitbhat: Done.

PR #2821, opened 2020-04-24
Iterative LB example to demonstrate hang with CMK_GLOBAL_LOCATION_UPDATE

minitu:
The added example (`examples/charm++/load_balancing/iterative`) demonstrates a hang with LB when Charm++ is compiled with `CMK_GLOBAL_LOCATION_UPDATE`. Also confirmed that it doesn't hang without `CMK_GLOBAL_LOCATION_UPDATE`. Tested on LLNL Lassen.

Charm++ build command: `./build charm++ ucx-linux-ppc64le smp ompipmix -j -g -O0 --basedir=$HOME/openmpi-4.0.1/install -DCMK_GLOBAL_LOCATION_UPDATE`

Test command: `jsrun -n2 -a1 -c2 -g1 -K1 -r2 ./iterative -i 100 +ppn 1 +pemap L0,80 +commap L4,84 +balancer RotateLB +LBDebug 1`

Output:
```
Charm++> Running in SMP mode: 2 processes, 1 worker threads (PEs) + 1 comm threads per process, 2 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: v6.11.0-devel-208-g4ad39a43d
Isomalloc> Synchronized global address space.
CharmLB> Verbose level 1, load balancing period: -1 seconds
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled.
Charm++> cpuaffinity PE-core map (logical indices): 0,80
HWLOC> [0] Thread 0x200001331640 bound to cpuset: 0x00000100
Charm++> Running on 1 hosts (2 sockets x 20 cores x 4 PUs = 160-way SMP)
Charm++> cpu topology info is gathered in 0.003 seconds.
CharmLB> RotateLB created.
[Load balancing itertions test]
Iters: 100, Data size: 128 bytes
Testing chare array...
CharmLB> RotateLB: PE [0] step 0 starting at 0.012207 Memory: 236.687500 MB
CharmLB> RotateLB: PE [0] strategy starting at 0.012472
CharmLB> RotateLB: PE [0] Memory: LBManager: 37 KB CentralLB: 1 KB
CharmLB> RotateLB: PE [0] #Objects migrating: 2, LBMigrateMsg size: 0.00 MB
CharmLB> RotateLB: PE [0] strategy finished at 0.012478 duration 0.000007 s
CharmLB> RotateLB: PE [0] step 0 finished at 0.012588 duration 0.000380 s
```

PR #2827, opened 2020-04-29
LocMgr: Fix IDs for CMK_GLOBAL_LOCATION_UPDATE

rbuch:
CkLocMgr deals only with element IDs. This fixes the pathway where
CMK_GLOBAL_LOCATION_UPDATE is enabled so that it does not pass IDs
with collection information to CkLocMgr functions

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2513,7 +2513,7 @@ int CkLocMgr::deliverMsg(CkArrayMessage *msg, CkArrayID mgr, CmiUInt8 id, const 
     // LB deals in IDs with collection information only when CMK_GLOBAL_LOCATION_UPDATE
     // is enabled, so add the group information if so.
 #if CMK_GLOBAL_LOCATION_UPDATE
-    const CmiUInt8 lbObjId = ck::ObjID(thisgroup.idx, id).getID();
+    const CmiUInt8 lbObjId = ck::ObjID(getGroupID(), id).getID();
 #else
     const CmiUInt8 lbObjId = id;
 #endif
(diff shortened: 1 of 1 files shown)


rbuch:
Fixes hang observed in #2821.

minitu:
I'm getting a build error

src/ck-core/cklocation.C
```diff
@@ -2509,13 +2510,15 @@ int CkLocMgr::deliverMsg(CkArrayMessage *msg, CkArrayID mgr, CmiUInt8 id, const
 #if CMK_LBDB_ON
   if ((idx || compressor) && type==CkDeliver_queue && !(opts & CK_MSG_LB_NOTRACE) && lbmgr->CollectingCommStats())
   {
+    // LB deals in IDs with collection information only when CMK_GLOBAL_LOCATION_UPDATE
+    // is enabled, so add the group information if so.
 #if CMK_GLOBAL_LOCATION_UPDATE
-    CmiUInt8 locMgrGid = thisgroup.idx;
-    id = ck::ObjID(id).getElementID();
-    id |= locMgrGid << ck::ObjID().ELEMENT_BITS;
+    const CmiUInt8 lbObjId = ck::ObjID(getGroupID(), id).getID();
```
rbuch, line 2516:
```suggestion
    const CmiUInt8 lbObjId = ck::ObjID(thisgroup, id).getID();
```

rbuch: I think it is slightly clearer if the variable is used directly so it's obvious that you're passing in the ID of this location manger group.

PR #2829, opened 2020-04-30
Map refactor

epmikida:
Just a draft PR for now to get feedback on which maps are important to keep


stwhite91:
It seems like we could preserve the old map API to not break existing user-defined maps, possibly by

stwhite91:
Oops, ignore that

PR #2833, opened 2020-05-02
Remove static function and callback registrations in LB

rbuch:


kavithachandrasekar:
Just making a note here that we would want to replace other static functions as well that are related to LB - staticMigrate(), staticMetaLBResumeWaitingChares() and staticMetaLBCallLBOnChares(). Maybe those can go in a separate PR.

PR #2834, opened 2020-05-04
Converse randomttl benchmark: Move variable init to main function

nitbhat:
This fixes the crash in the program caused because of accessing
variables before their initialization. Previously, the vector 'peVector'
was being accessed by a message handler function before the values in
'peVector' were set.

### DIFF
--- a/benchmarks/converse/randomttl/randomttl.C
+++ b/benchmarks/converse/randomttl/randomttl.C
@@ -166,7 +166,7 @@ CmiStartFn mymain(int argc, char *argv[])
   std::srand (unsigned(std::time(0)) + CmiMyPe());
 
   // set values to Pes
-  for (int i=0; i <  CmiNumPes(); i++) CpvAccess(peVector).push_back(i);
+  for (int i=0; i < CmiNumPes(); i++) CpvAccess(peVector).push_back(i);
 
   // using built-in random generator
   std::random_shuffle(CpvAccess(peVector).begin(), CpvAccess(peVector).end());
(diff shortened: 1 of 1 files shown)


benchmarks/converse/randomttl/randomttl.C
```diff
@@ -169,6 +163,14 @@ CmiStartFn mymain(int argc, char *argv[])
       CmiAbort("Usage: ./randomttl <k> <ttl>\n");
   }
 
+  std::srand (unsigned(std::time(0)) + CmiMyPe());
+
+  // set values to Pes
+  for (int i=0; i <  CmiNumPes(); i++) CpvAccess(peVector).push_back(i);
```
minitu, line 169:
```suggestion
  for (int i=0; i < CmiNumPes(); i++) CpvAccess(peVector).push_back(i);
```

nitbhat: Thanks @minitu

benchmarks/converse/randomttl/randomttl.C
```diff
@@ -169,6 +163,14 @@ CmiStartFn mymain(int argc, char *argv[])
       CmiAbort("Usage: ./randomttl <k> <ttl>\n");
   }
 
+  std::srand (unsigned(std::time(0)) + CmiMyPe());
+
+  // set values to Pes
+  for (int i=0; i < CmiNumPes(); i++) CpvAccess(peVector).push_back(i);
```
stwhite91, line 169: It'd be more efficient to replace this loop with the following:

CpvAccess(peVector).resize(CmiNumPes());
std::iota(CpvAccess(peVector).begin(), CpvAccess(peVector).end(), 0);

nitbhat: Good point, done!

PR #2836, opened 2020-05-05
Shrink the size of some enum types in LRTS

stwhite91:


evan-charmworks:
Looks like further cleanup is needed for this to work, since the headers containing these structs are still included by C files. The simplest thing to try would be surrounding the structs with `#ifdef __cplusplus`.

evan-charmworks:
What was the reason we don't compile charmxi as C++11?

matthiasdiener:
> What was the reason we don't compile charmxi as C++11?

FWIW, charmxi is built as C++11 in the cmake build system.

PhilMiller:
> What was the reason we don't compile charmxi as C++11?

Seemingly neglect, since the C++ parts of the runtime were switched to compiling as C++11 while I was still there. Where/when there's a difference, the host compiler that builds charmxi has pretty much always been ahead of the target machine compiler on standards compliance.

matthiasdiener:
Out of curiosity, does this make any difference in practice?

evan-charmworks:
This PR needs build system changes so that charmxi's `NATIVE` compiler is passed the parameter needed to activate C++11.

PR #2846, opened 2020-05-09
cmake: use RULE_LAUNCH_COMPILE instead of CMAKE_*_COMPILER_LAUNCHER

matthiasdiener:
RULE_LAUNCH_COMPILE is supported by older cmake versions (up to 3.0 at least).
Allows Fortran support with the same cmake versions as C/C++ support.


stwhite91:
Do we no longer need CMake v3.11 when building with Fortran?

matthiasdiener:
> Do we no longer need CMake v3.11 when building with Fortran?

Yes, 3.4 is the minimum for all features now.

PR #2847, opened 2020-05-10
cmake: add gni build

matthiasdiener:


matthiasdiener:
> Some places use `gni-*` (with asterisk) followed by `mpi-crayx` with the `MATCHES` operator, while others use `gni-` (no asterisk) and `-crayx`. Could it be made consistent?

Good point, should be fixed now.

PR #2854, opened 2020-05-13
Util: Replace dummy cache padding with alignas

rbuch:


rbuch:
> We should add a configure check for support of C++17's std::hardware_destructive_interference_size and set CMI_CACHE_LINE_SIZE to that when it is defined. IIRC we hardcoded 64 as the cache line size for all architectures except ppc64le which is 128.

Okay, I've added this. I was unable to actually use this on any compiler I tried (gcc 7.5.0, 10.0.1, clang 6.0.0, 11.0.0). According to [this](https://stackoverflow.com/questions/60004804/where-is-stdhardware-destructive-interference-size), it seems like it's not implemented in either libstdc++ or libc++, so I supposed that's expected. I think this implementation should work when it's added, though.

As a sidenote, according to http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0154r1.html, Google seems to use the following for cache alignment stuff, which may improve our alignment on ARM:

```
// Cache line alignment
#if defined(__i386__) || defined(__x86_64__)
#define CACHELINE_SIZE 64
#elif defined(__powerpc64__)
// TODO(dougkwan) This is the L1 D-cache line size of our Power7 machines.
// Need to check if this is appropriate for other PowerPC64 systems.
#define CACHELINE_SIZE 128
#elif defined(__arm__)
// Cache line sizes for ARM: These values are not strictly correct since
// cache line sizes depend on implementations, not architectures.  There
// are even implementations with cache line sizes configurable at boot
// time.
#if defined(__ARM_ARCH_5T__)
#define CACHELINE_SIZE 32
#elif defined(__ARM_ARCH_7A__)
#define CACHELINE_SIZE 64
#endif
#endif

#ifndef CACHELINE_SIZE
// A reasonable default guess.  Note that overestimates tend to waste more
// space, while underestimates tend to waste more time.
#define CACHELINE_SIZE 64
#endif

#define CACHELINE_ALIGNED __attribute__((aligned(CACHELINE_SIZE)))
```

stwhite91:
We might as well use the google code for setting ARM cache line size

rbuch:
> We might as well use the google code for setting ARM cache line size

On digging a bit deeper, I think we are close enough to what they have already. The `defined(__arm__)` style checks are not portable (seemingly gcc specific), and we otherwise have no good way of detecting the distinctions between different generations of ARM (and even then, it's no guarantee that it actually reflects the hardware). Defaulting to 64 for everything except PPC seems like the right move for now.

PR #2858, opened 2020-05-14
Fix UCX CI hang

matthiasdiener:
Updates Open MPI to 4.0.3 (from 4.0.1) and installs a package dependency of 4.0.3. Not sure what made 4.0.1 suddenly stop working. 

Some data points:
- Hangs appear randomly when starting a Charm++ program with charmrun/mpirun.
- Hangs were reproduced on other machines, not only on CI.
- I was able to see the hangs with 4.0.1 on Charm++ versions several months old, with both build systems.

The hangs looked like this:
```
../../../../../../bin/testrun  +p2 ./simple_get 20  
real	0m0.073s
user	0m0.025s
sys	0m0.049s

Running on 2 processors:  ./simple_get 20 
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 2  ./simple_get 20
```


matthiasdiener:
Needs another review.

PR #2861, opened 2020-05-14
Use CkReferenceMsg to avoid duplicating TreeLB messages

### LINKED ISSUE #2857: Remove unnecessary message copying in TreeLB
Currently, `TreeLevel.h` makes copies of the decision message when sending it to children. Change this so that `CmiReference` or some other scheme is used to avoid these unnecessary copies when possible while still allowing the sends in `TreeLB` to work with arrays of messages and ensure that they get properly freed.

rbuch:
Closes #2857.

### DIFF
--- a/src/ck-ldb/TreeLevel.h
+++ b/src/ck-ldb/TreeLevel.h
@@ -552,7 +552,9 @@ class RootLevel : public LevelLogic
       decisions.resize(num_children);
       decisions[0] = migMsg;
       for (int i = 1; i < num_children; i++)
-        decisions[i] = (TreeLBMessage*)CkCopyMsg((void**)&migMsg);
+      {
+        decisions[i] = CkReferenceMsg(migMsg);
+      }
     }
     else
     {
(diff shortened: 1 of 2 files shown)


src/ck-ldb/TreeLevel.h
```diff
@@ -552,7 +552,10 @@ class RootLevel : public LevelLogic
       decisions.resize(num_children);
       decisions[0] = migMsg;
       for (int i = 1; i < num_children; i++)
-        decisions[i] = (TreeLBMessage*)CkCopyMsg((void**)&migMsg);
+      {
+        CmiReference(UsrToEnv(migMsg));
```
stwhite91, line 556: We should really add a CkMsgReference() API (and perhaps a CkMsgSetUsed) like this:

void *CkMsgReference(void *msg) {
    CmiReference(UsrToEnv(msg));
    return msg;
}

Then this can become  'decisions[i] = CkMsgReference(migMsg);'

Can be done in a separate patch

rbuch: Yeah, good idea, I've made #2862.

rbuch:
Will be broken until #2862 is merged into `master`.

stwhite91:
> Will be broken until #2862 is merged into `master`.

You could rebase on top of #2862 using the 'chain' app Nitin added. Either way the commit msg should be updated to say CkReferenceMsg instead of CmiReference

PR #2862, opened 2020-05-14
Add CkReferenceMsg function

rbuch:
A common paradigm in the codebase is to increment the refcount of a
message by using CmiReference(UsrToEnv(msg)), this creates a function
to simplify this usage.

### DIFF
--- a/src/ck-core/debug-charm.C
+++ b/src/ck-core/debug-charm.C
@@ -66,8 +66,7 @@ void CpdBeforeEp(int ep, void *obj, void *msg) {
     entry.memoryBackup = NULL;
     entry.obj = obj;
     if (msg != NULL) {
-      entry.msg = msg;
-      CkReferenceMsg(msg);
+      entry.msg = CkReferenceMsg(msg);
     }
     else entry.msg = NULL;
     _debugData.push(entry);
--- a/src/ck-core/debug-charm.C
+++ b/src/ck-core/debug-charm.C
@@ -67,7 +67,7 @@ void CpdBeforeEp(int ep, void *obj, void *msg) {
     entry.obj = obj;
     if (msg != NULL) {
       entry.msg = msg;
-      CmiReference(UsrToEnv(msg));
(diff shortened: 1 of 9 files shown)


matthiasdiener:
Why does this return the message?

stwhite91:
> Why does this return the message?

That keeps the function signature the same as CkCopyMsg() and can be useful, as in PR #2861

epmikida:
> May be a good idea to have a PR which cleans up usage where relevant now that we have this. A search/replace for `CmiReference(UsrToEnv(*))` -> `CkReferenceMsg(*)` shouldn't be too hard to do.

Nevermind...you already did that. I guess I was just expecting more instances of it.

src/ck-core/debug-charm.C
```diff
@@ -67,7 +67,7 @@ void CpdBeforeEp(int ep, void *obj, void *msg) {
     entry.obj = obj;
     if (msg != NULL) {
       entry.msg = msg;
```
stwhite91, line 69: The body of this if-statement can be simplified to 'entry.msg = CkReferenceMsg(msg);' now

src/xlat-i/xi-Parameter.C
```diff
@@ -712,7 +712,7 @@ void ParamList::beginUnmarshallSDAGCall(XStr& str, bool usesImplBuf) {
     if (hasArray || hasRdma()) {
       if (!usesImplBuf) {
         str << "  genClosure->_impl_marshall = impl_msg_typed;\n";
```
stwhite91, line 714: These two lines could be combined as well

src/ck-core/debug-charm.C
rbuch: Good call, done.

src/xlat-i/xi-Parameter.C
rbuch: Doing that causes a type error due to it returning a `void*`, so I'm leaving this one as it is.

PR #2865, opened 2020-05-15
Create VERSION file in bin/ instead of include/

### LINKED ISSUE #2844: Conflict between LLVM's version file and our VERSION file
On MacOS, which generally has case-insensitive file systems, the `VERSION` file that gets generated while building Charm++ conflicts with a `version` header file provided by LLVM, leading to compilation errors:

```
clang++ -fPIC -dynamic -fno-common -stdlib=libc++ -DCMK_GFORTRAN -mmacosx-version-min=10.7 -D_DARWIN_C_SOURCE -I../include -I. -c test.cpp -o test.o -std=c++11
In file included from test.cpp:3:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/map:479:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/__tree:14:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/iterator:421:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/__functional_base:14:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/type_traits:417:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/cstddef:37:
../include/version:1:1: error: expected unqualified-id
61001
^
```

See https://github.com/llvm-mirror/libcxx/blob/master/include/cstddef#L37

This is with llvm 10.0 (homebrew) on MacOS 10.15.4, both old and new build systems are affected.
llvm 9 is probably affected too.

Potential fixes:
- rename our `VERSION` file.
- make our `VERSION` file C-compliant (e.g. by adding `//` before the version number). This would avoid the compilation error, but LLVM's version file would still not be included (which is not a critical error I think).
- remove `VERSION` completely.

evan-charmworks:
Fixes #2844


matthiasdiener:
Am I right in seeing that with the old build system, there are two different `VERSION` files created, `include/VERSION` and `tmp/VERSION`:

```
$ cat ./netlrts-linux-x86_64/include/VERSION
61001
$ cat ./netlrts-linux-x86_64/tmp/VERSION
v6.11.0-devel-244-g9215bc0a8
```

evan-charmworks:
I don't think this is the right solution.

PR #2867, opened 2020-05-16
CI: Use UCX v1.8.0

evan-charmworks:


evan-charmworks:
Not ready for merge.

evan-charmworks:
@nitbhat Was there something you found in your experiments that would suggest we should keep the CI on UCX 1.7.0 until 1.9.0 comes out?

nitbhat:
> @nitbhat Was there something you found in your experiments that would suggest we should keep the CI on UCX 1.7.0 until 1.9.0 comes out?

Not really, we can move on to UCX 1.8.0. There was a bug that I was facing while running with `+trackMsgs` (the message tracking infrastructure) and that was supposedly fixed after v1.8.0. However, I don't think that should hold us back to UCX v1.7.0 (since that might have the bug as well).

PR #2874, opened 2020-05-22
ampiCC: Accept arguments from environment variable

evan-charmworks:
Exporting AMPI_BUILD_FLAGS with values such as `-charm-shared` will
allow passing options without modifying an MPI program's build scripts.

### DIFF
--- a/src/libs/ck-libs/ampi/ampiCC
+++ b/src/libs/ck-libs/ampi/ampiCC
@@ -80,7 +80,7 @@ do
 done
 }
 
-eval processArgs "$@"
+eval processArgs "$AMPI_BUILD_FLAGS" "$@"
 
 # If there are no filename arguments specified, just pass through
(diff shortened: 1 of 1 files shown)


matthiasdiener:
out of curiosity, do we already honor `CFLAGS` etc?

evan-charmworks:
To my knowledge, CFLAGS is something accepted (or not) by any given project's Makefile. Moreover, flags like `-charm-shared` and `-fsglobals` are not C flags, they are charmc or ampiCC flags.

matthiasdiener:
> To my knowledge, CFLAGS is something accepted (or not) by any given project's Makefile. Moreover, flags like `-charm-shared` and `-fsglobals` are not C flags, they are charmc or ampiCC flags.
> 
> Part of the goal here is something that works 100% reliably.

Right, makes sense.

PR #2878, opened 2020-05-23
ampiCC: Handle OpenMPI's -showme parameter

evan-charmworks:


src/libs/ck-libs/ampi/ampiCC
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-AMPICC="$0"
+[ -z "$AMPICC" ] && AMPICC="$(basename $0)"
```
matthiasdiener, line 2: could be `AMPICC=${AMPICC:-$(basename $0)}`

PR #2880, opened 2020-05-25
CMake: Fix CMake variable reference to use braces instead of parens

rbuch:


rbuch:
> Are you sure this is correct? The idea here was, I believe, to pass through `VERBOSE` at the time the top-level `make` is called with (e.g.) `make VERBOSE=1`. In other words, `VERBOSE` shouldn't be set by `cmake` when configuring Charm++, but through `make` at build time. ~It might be that the correct command here is `make V=$VERBOSE`.~
> 
> Edit:
> After some more tests, I think the original behavior is correct. `VERBOSE` is a reference to bash variable, not a cmake variable. Compare the output difference between enabling it or not:
> 
> ```shell
> $ mkdir mybuild && cd mybuild && cmake ..
> $ make hwloc VERBOSE=1
> $ make clean
> $ make hwloc
> ```

Ah, I see.  I'll close this for the time being, then.

PR #2108, opened 2019-05-08
Cleanup #1314: Replace PUPed raw arrays in src/ck-ldb/ with std::vector

evan-charmworks:
*Original date: 2019-05-07 17:34:22*
*Original PR: https://charm.cs.illinois.edu/gerrit/5171*

---

Skipped:
LBMigrateMsg

Change-Id: Ifd5a7cc5e01b13982be3469eb214b85a7e9fe9cc

PR #2886, opened 2020-05-26
Remove CkVec from ck-ldb

rbuch:
- [x] Test with NAMD
- [x] Test with ChaNGa
- [x] Verify that stats message is being cleared/there is no memory leak in `CentralLB` and `HybridBaseLB`


src/ck-ldb/CentralLB.C
```diff
@@ -419,35 +419,24 @@ void CentralLB::MissMigrate(int waitForBarrier)
...
+       statsData->objData.insert(statsData->objData.end(), msg->objData, msg->objData + msg->n_objs);
+       statsData->from_proc.insert(statsData->from_proc.end(), msg->n_objs, pe);
+       statsData->to_proc.insert(statsData->to_proc.end(), msg->n_objs, pe);
+       statsData->commData.insert(statsData->commData.end(), msg->commData, msg->commData + msg->n_comm);
        // free the memory
        delete msg;
        statsMsgsList[pe]=0;
     }
-    statsData->n_migrateobjs = nmigobj;
+    statsData->n_migrateobjs = count_if(statsData->objData.begin(),
```
stwhite91, line 437: Maybe more consistent to do "std::count_if" here or to do "count_if" below

src/ck-ldb/CentralLB.C
```diff
@@ -883,29 +876,26 @@ void CentralLB::removeNonMigratable(LDStats* stats, int count)
...
+  new_to_proc.reserve(stats->n_migrateobjs);
+  for (i=0; i<stats->objData.size(); i++)
   {
     LDObjData &odata = stats->objData[i];
     if (odata.migratable) {
-      nonmig[n_objs] = odata;
-      new_from_proc[n_objs] = stats->from_proc[i];
-      new_to_proc[n_objs] = stats->to_proc[i];
-      n_objs ++;
+      nonmig.push_back(odata);
```
stwhite91, line 896: I think this is better as emplace_back?

src/ck-ldb/CentralLB.C
```diff
@@ -944,20 +932,17 @@ void CentralLB::removeNonMigratable(LDStats* stats, int count)
     case LD_OBJLIST_MSG:    // object message FIXME add multicast
       break;
     }
-    newCommData[n_comm] = cdata;
-    n_comm ++;
+    newCommData.push_back(cdata);
```
stwhite91, line 935: emplace_back

src/ck-ldb/HybridBaseLB.h
```diff
@@ -379,10 +379,10 @@ class HybridBaseLB : public CBase_HybridBaseLB
     int info_recved;		// for CollectInfo()
     int vector_expected, vector_completed;
     int resumeAfterMigration;
-    CkVec<MigrationRecord> outObjs;
-    //CkVec<Location> unmatchedObjs;
+    std::vector<MigrationRecord> outObjs;
+    //std::vector<Location> unmatchedObjs;
     std::map< LDObjKey, int >  unmatchedObjs;
```
stwhite91, line 384: Could this be unordered_map?

src/ck-ldb/CentralPredictor.C
```diff
@@ -202,26 +202,20 @@ void CentralLB::FuturePredictor(BaseLB::LDStats* stats) {
...
   } else {
-
+    const int n_objs = stats->objData.size();
     if (predicted_model->parameters == NULL) {     // time to create the new prediction model
       // allocate parameters
-      predicted_model->model_valid = new bool[stats->n_objs];
-      predicted_model->parameters = new double*[stats->n_objs];
-      for (i=0; i<stats->n_objs; ++i) predicted_model->parameters[i] = new double[predicted_model->predictor->num_params];
-      for (i=0; i<stats->n_objs; ++i) {
+      predicted_model->model_valid = new bool[n_objs];
```
stwhite91, line 215: Would be nicer to have these as vectors as well

src/ck-ldb/HybridBaseLB.C
```diff
@@ -707,22 +702,19 @@ void HybridBaseLB::CreateMigrationOutObjs(int atlevel, LDStats* stats,
...
   if (statsData != NULL) {
-    CkVec<LDObjData> &oData = statsData->objData;
-
     for (i=0; i<m; i++)
     {
       // copy into LDStats
       LDObjData &data = datas[i];
-      oData.push_back(data);
-      statsData->n_objs++;
+      statsData->objData.push_back(data);
```
stwhite91, line 717: emplace_back

src/ck-ldb/HybridBaseLB.C
```diff
@@ -822,11 +810,8 @@ void HybridBaseLB::ObjMigrated(LDObjData data, LDCommData *cdata, int n, int atl
   LDStats *statsData = lData->statsData;
 
   if (statsData != NULL) {
-    CkVec<LDObjData> &oData = statsData->objData;
-
     // copy into LDStats
-    oData.push_back(data);
-    statsData->n_objs++;
+    statsData->objData.push_back(data);
```
stwhite91, line 814: emplace_back

evan-charmworks:
Related to #2108 which is probably out of date after the new framework merges.

src/ck-ldb/HybridBaseLB.h
rbuch: Yeah, but that's out of scope for this change, I think; we'd need to add hashing for `LDObjKey`, so it's not a trivial change.

src/ck-ldb/CentralLB.C
rbuch: Agreed, added `std::` to all uses.

src/ck-ldb/CentralLB.C
rbuch: Since this is adding an object that already exists, I don't think there would be any difference between `emplace_back` and `push_back` in this case. I'd prefer to keep it as `push_back` to make it clear that it's going to be calling a copy constructor (see [this](https://abseil.io/tips/112) for more rationale).

stwhite91: I think push_back will result in calling the copy constructor to create a new object, which will then be moved into the vector, whereas emplace_back will call the copy constructor in place, avoiding the move into the vector. The extent of the performance difference depends on LDObjData, its size, whether it has a move constructor that is different from its copy constructor.

src/ck-ldb/BaseLB.h
```diff
@@ -127,11 +125,11 @@ class BaseLB: public CBase_BaseLB
     int getRecvHash(LDCommData &cData);
     void clearCommHash();
     void clear() {
-      n_objs = n_migrateobjs = n_comm = 0;
-      objData.free();
-      commData.free();
-      from_proc.free();
-      to_proc.free();
+      n_migrateobjs = 0;
+      objData.clear();
```
stwhite91, line 129: Should we be calling shrink_to_fit() after each of these clear calls? Or do we expect these vectors to be reused after this with roughly the same size as before it?

src/ck-ldb/BaseLB.C
```diff
@@ -147,14 +147,13 @@ void BaseLB::LDStats::makeCommHash() {
   if (objHash) return;
    
   int i;
-  hashSize = n_objs*2;
-  hashSize = primeLargerThan(hashSize);
+  hashSize = primeLargerThan(objData.size() * 2);
   objHash = new int[hashSize];
```
stwhite91, line 151: Could be vector<int>

src/ck-ldb/CentralLB.C
rbuch: I wrote a test program to check this, and both `push_back` and `emplace_back` just call a single copy constructor per invocation, directly constructing into the target location in the vector in the case where the argument is of their template type `T`. Behavior is only different in the case where it can use perfect forwarding to avoid having to do a premature construction when arguments to a constructor are passed in, in which case it can defer calling the constructor until it can be done in place.

For example, as [seen in libc++](https://github.com/llvm/llvm-project/blob/master/libcxx/include/vector), both `push_back` and `emplace_back` are essentially calls to `__construct_one_at_end` with arguments passed in in different ways (either directly as a `const T&` or with `std::move` for `push_back` and via `std::forward` for `emplace_back`). In the case of `nonmig.push_back(odata)`, `odata` gets passed as a `const_reference` until `__construct_one_at_end`, where it is then used for a copy construction. For `nonmig.emplace_back(odata)`, `odata` gets passed as a perfectly forwarded argument until `__construct_one_at_end`, where it is then used for a copy construction in the same way. In both cases, there is a single copy construction.

If you want to avoid calling the copy constructor completely when adding an existing object `T obj` to a `vector<T>`, you can use either `push_back(std::move(obj))` or `emplace_back(std::move(obj))`, both are identical.

src/ck-ldb/CentralPredictor.C
rbuch: Done.,

src/ck-ldb/BaseLB.C
rbuch: Done.

src/ck-ldb/BaseLB.h
rbuch: No, the expectation is that these should be reused with similarly sized data, so we should leave them as is.

stwhite91:
Does this still need to be tested with NAMD and ChaNGa as the commit msg body says?

rbuch:
Fix for NAMD: https://gitlab.com/tcbgUIUC/namd/-/merge_requests/51
Fix for ChaNGa: https://github.com/N-BodyShop/changa/pull/93

rbuch:
> Does this still need to be tested with NAMD and ChaNGa as the commit msg body says?

It's ready for merge now, it's been tested with NAMD and ChaNGa and they've both merged patches to support this change.

evan-charmworks (approved):

stwhite91 (approved):

PR #2888, opened 2020-05-28
Remove deprecated CK_DEFAULT_BITWISE_PUP and PUPmarshall

stwhite91:


matthiasdiener:
There is still `#define PUPmarshallBytes(type)` in src/util/pup.h

stwhite91:
> There is still `#define PUPmarshallBytes(type)` in src/util/pup.h

Removed

PR #2895, opened 2020-06-02
Add CcdSCHEDLOOP and move HAPI cleanup

minitu:
The following changes contribute to proper layering of HAPI and enable compilation of Converse-only programs.
- Adds `CcdEVERY` for functions that should be invoked at every scheduler loop, such as `hapiPollEvents`
- Moves HAPI cleanup code from Converse to Charm++ exit


minitu:
> Overall change looks good.
> 
> About CcdEVERY, I was wondering how the behavior will be when it is called with `CcdCallOnCondition` instead of `CcdCallOnConditionKeep`. In that case, will it be called only once? (In which case, it is not 'EVERY' scheduler loop anymore). So, I would be in favor of renaming `CcdEVERY` to something like `CcdSCHEDLOOP`, since `CcdEVERY` is quite generic.

I'm guessing that's the behavior based on the documentation, although I've never used `CcdCallOnCondition`. Good point on the naming, will submit an update.

minitu:
@stwhite91 @nitbhat Please review again

minitu:
@stwhite91 @nitbhat Added CcdSCHEDLOOP to the Converse manual, please review again

PR #2896, opened 2020-06-04
Intra-node GPU messaging #1: Add DeviceManager, buddy allocator and CUDA IPC

minitu:
First PR out of 4 planned PRs for intra-node GPU messaging.
Following additions and modifications were made:

- Added `DeviceManager`, a per-device entity
- Added buddy allocator which uses the buddy allocation algorithm to handle possible concurrent allocation and deallocation requests from PEs mapped to the same device, minimizing external fragmentation
- Added inter-process GPU messaging using POSIX shared memory and CUDA IPC
- ~Moved HAPI initialization code to before CPU topology initialization to avoid the hang caused by a `CmiBarrier` call~
- Replace repeated uses of `CsvAccess` and `CpvAccess` with a reference variable (e.g. `GPUManager& csv_gpu_manager = CsvAccess(gpu_manager)`)

PR sequence: #2896 -> #2901 -> #2903 -> #2904


src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -519,6 +627,171 @@ hapiWorkRequest::hapiWorkRequest() :
 #endif
   }
 
+// Create POSIX shared memory region accessible to all processes on the same host
+// Invoked by PE rank 0 of each process (no locking needed for SMP)
+void shmCreate() {
+  struct stat shm_file_stat;
+
+  // Create the shared memory file
+  CsvAccess(gpu_manager).shm_name = new char[16];
```
stwhite91, line 636: This string should be longer (24?), it will overflow in the case that we ever have >99 PEs per node.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -519,6 +627,171 @@ hapiWorkRequest::hapiWorkRequest() :
...
+      CsvAccess(gpu_manager).shm_chunk_size * CpvAccess(my_device));
+  void* device_ptr = (void*)(CsvAccess(gpu_manager).device_managers[CpvAccess(my_device)].comm_buffer->base_ptr);
+  hapiCheck(cudaIpcGetMemHandle(shm_mem_handle, device_ptr));
+
+  // Create CUDA IPC events and corresponding handles
+  cuda_ipc_event_shared* shm_event_shared = (cuda_ipc_event_shared*)((char*)shm_mem_handle + sizeof(cudaIpcMemHandle_t));
+  int device_index = CsvAccess(gpu_manager).device_count * CmiMyNodeRankLocal() + CpvAccess(my_device);
+  cuda_ipc_device_info& my_device_info = CsvAccess(gpu_manager).cuda_ipc_device_infos[device_index];
+
+  my_device_info.event_pool_flags = new int[CsvAccess(gpu_manager).ipc_event_pool_size];
```
stwhite91, line 735: Is there a good reason for all of gpu_manager's arrays to be bare pointers rather than std::string and std::vector?

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Good point, will update

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Not really, at least the ones in this patch are mostly artifacts from referring to MVAPICH's C-based code.
Will go through them and C++-fy them if I can.

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Changed to use `std::string`

minitu:
@stwhite91 Made changes to use `std::string` and `std::vector` instead of pointers as much as possible. I also wanted to have a structure that encapsulates `cudaIpcMemHandle_t` and multiple `cuda_ipc_event_shared` objects to avoid more pointer calculations, but it seems infeasible because the number of IPC events is a runtime variable and the size of shared memory that stores them is static (cannot be resized).

src/arch/cuda/hybridAPI/buddy_allocator.h
```diff
@@ -0,0 +1,247 @@
...
+      buckets[bucket_count-1].emplace_back(base_ptr, limit);
+    }
+
+    ~allocator() {
+      // Free GPU memory
+      cudaFree(base_ptr);
+      delete[] buckets;
+    }
+
+    void* malloc(size_t request) {
```
nitbhat, line 134: I would suggest changing the name of `request` to `requestedSize` or something that gives and an indication that it is a size of some sort. The same suggestion for `limit`. (And you can change the `size` parameter in `allocator`'s constructor to maybe `deviceAllocationSize`).

nitbhat: Additionally, based on the suggestion for variable names in the style guide (https://github.com/UIUC-PPL/charm/wiki/Charm-Runtime-Style-Guide), it'll be good to skip the underscore in variable names and use the capitalization format.

src/arch/cuda/hybridAPI/buddy_allocator.h
```diff
@@ -0,0 +1,247 @@
...
+      int bucket = get_bucket(alloc_size);
+      int original_bucket = bucket;
+
+      // Find an empty bucket
+      while (buckets[bucket].empty() && bucket < bucket_count) {
+        bucket++;
+      }
+
+      // All buckets were empty!
+      if (bucket == bucket_count) {
```
nitbhat, line 150: Should the comment read "// All buckets were full!"?

nitbhat: In the case where no free blocks are available, I'm wondering if we should allocate more device memory using `cudaMalloc` and keep all the allocator objects in a list? I'm guessing if multiple consecutive allocations are requested and all of them wait to be freed (because of transfers in progress), then more memory cannot be requested since the buckets will be full?

src/arch/cuda/hybridAPI/buddy_allocator.h
```diff
@@ -0,0 +1,247 @@
+#ifndef __BUDDY_ALLOCATOR_H_
```
nitbhat, line 1: There seems to be quite a bit of code in this header file. It'll be good to just limit small functions and data structure definitions to headers and move other code to a `.C`.

src/ck-core/init.C
```diff
@@ -1627,6 +1629,50 @@ void _initCharm(int unused_argc, char **argv)
 	}	
 #endif
 
+#if CMK_CUDA
+  // Perform HAPI initialization for GPU support
```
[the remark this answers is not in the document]
minitu: Agreed, will make that change

src/ck-core/init.C
```diff
@@ -1627,6 +1629,50 @@ void _initCharm(int unused_argc, char **argv)
...
+    }
+
+    CmiNodeBarrier();
+
+    ipcHandleCreate(); // Create CUDA IPC handles
+  }
+
+  // Ensure CUDA IPC handles are available for all processes
+  // Note: Causes a hang when this barrier is placed after CPU topology initialization
+  // FIXME: This only needs to be a host-wide synchronization
```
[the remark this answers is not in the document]
minitu: Not in this patch, planning to update it later. The problem is AFAIK we currently don't have a host-wide barrier (`CmiNodeBarrier` is a process-wide barrier).

src/arch/cuda/hybridAPI/buddy_allocator.h
minitu: Agreed on `request` and `limit`. But as for the style, HAPI is mostly based on the Google style (at least my additions), so will keep it that way for now.

src/arch/cuda/hybridAPI/buddy_allocator.h
minitu: > Should the comment read "// All buckets were full!"?

You're right, will change this

> In the case where no free blocks are available, I'm wondering if we should allocate more device memory using `cudaMalloc` and keep all the allocator objects in a list? I'm guessing if multiple consecutive allocations are requested and all of them wait to be freed (because of transfers in progress), then more memory cannot be requested since the buckets will be full?

Currently more memory is not allocated when there are no free blocks because the underlying memory is mmap'ed (for use of CUDA IPC) and cannot be dynamically resized. I'm planning to update this so that we can allocate a larger backing store once it runs out and mmap it again (albeit costly, because it has to be mmap'ed on all processes).

src/arch/cuda/hybridAPI/buddy_allocator.h
minitu: I was originally thinking of a header-only library but now I agree that it's better to move the implementation to a `.C`, will update

minitu:
> * ~Moved HAPI initialization code to before CPU topology initialization to avoid the hang caused by a `CmiBarrier` call~

HAPI actually uses `CmiNumPhysicalNodes` in the PE-device mapping phase, which only works (returns the correct value) after CPU topology initialization.
So the hang which is observed when `CmiBarrier` is used after CPU topology init needs to be fixed after all, as `CmiBarrier` (or a host-wide barrier instead) is necessary to enforce all CUDA IPC handles are created across the host before they are opened.

minitu:
@nitbhat Addressed your reviews

src/arch/cuda/hybridAPI/Makefile
```diff
@@ -14,11 +14,14 @@ libs: libcudahybridapi.a
...
 	-rm -f $@
-	ar q $@ hybridapi.o
+	ar q $@ hybridapi.o buddy_allocator.o
 
-hybridapi.o: hapi_impl.cpp hapi_impl.h hapi.h hapi_nvtx.h
+hybridapi.o: hapi_impl.cpp hapi_impl.h gpumanager.h devicemanager.h buddy_allocator.h hapi.h hapi_nvtx.h
+	$(CHARMC) $(FLAGS) $(INC) -o $@ $<
+
+buddy_allocator.o : buddy_allocator.cpp buddy_allocator.h
 	$(CHARMC) $(FLAGS) $(INC) -o $@ $<
```
matthiasdiener, line 26: Do the changes need to be reflected in the CMake build?

src/arch/cuda/hybridAPI/buddy_allocator.cpp
```diff
@@ -0,0 +1,187 @@
...
+  int allocator::get_bucket(size_t size) {
+    return (int)std::ceil(std::log2((double)size)) - 2;
+  }
+
+  int allocator::get_block_index(uint8_t* ptr, size_t size) {
+    return (size_t)(ptr - base_ptr) / size;
+  }
+
+  allocator::allocator(size_t size) : min_size(4), base_ptr(NULL) {
+    if (size <= 0) {
```
matthiasdiener, line 55: Can size_t be less than 0?

src/arch/cuda/hybridAPI/buddy_allocator.cpp
```diff
@@ -0,0 +1,187 @@
...
+      fprintf(stderr, "Allocator size has to be larger than 0 bytes\n");
+      abort();
+    }
+
+    // Request GPU memory (closest power of 2)
+    int total_size_log2 = std::ceil(std::log2((double)size));
+    total_size = (size_t)std::pow(2, total_size_log2);
+    cudaMalloc(&base_ptr, total_size);
```
matthiasdiener, line 63: Should this check for an error?

matthiasdiener:
Can you link to the other 3 PRs in the first comment in this PR?

src/arch/cuda/hybridAPI/buddy_allocator.cpp
minitu: No, I'll change it to `size == 0`

src/arch/cuda/hybridAPI/buddy_allocator.cpp
minitu: Good point, will add error checking for `cudaMalloc` and `cudaFree` calls

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -89,15 +93,89 @@ CsvDeclare(GPUManager, gpu_manager);
 CpvDeclare(int, my_device); // GPU device that this thread is mapped to
 CpvDeclare(bool, device_rep); // Is this PE a device representative thread? (1 per device)
 
+// Returns the local rank of the logical node (process) that the given PE belongs to
```
nitbhat, line 96: Should we move these to machine-common-core.C? That's where the other similar functions live.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -89,15 +93,89 @@ CsvDeclare(GPUManager, gpu_manager);
...
+
+// Called by all PEs in Charm++ layer init
+void hapiInit(char** argv) {
+  if (!CmiInCommThread()) {
+    if (CmiMyRank() == 0) {
+      hapiInitCsv(); // Initialize per-process variables (GPUManager)
+    }
+    hapiInitCpv(); // Initialize per-PE variables
+
+    CmiNodeBarrier();
```
nitbhat, line 129: Can you add comments as to why a NodeBarrier is needed after a set of operations? (Pardon my lack of within-node/device level stuff).

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -199,6 +274,12 @@ void hapiMapping(char** argv) {
...
     }
 
+    // Create a DeviceManager per GPU device
+    std::vector<DeviceManager>& device_managers = CsvAccess(gpu_manager).device_managers;
+    for (int i = 0; i < device_count; i++) {
+      device_managers.emplace_back(i, device_count * CmiMyNodeRankLocal() + i);
+    }
+
     // Count number of PEs per device
     CsvAccess(gpu_manager).pes_per_device = CmiNodeSize(CmiMyNode()) / device_count;
```
nitbhat, line 284: Since this code is called by each PE, it looks like each PE of the node will be writing to this `csv` variable. Shouldn't it be written to by only PE 0?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -231,8 +312,97 @@ void hapiMapping(char** argv) {
...
+
+  // Process device communication buffer parameters (in MB)
+  int input_comm_buffer_size;
+  if (CmiGetArgIntDesc(argv, "+gpucommbuffer", &input_comm_buffer_size,
+        "GPU communication buffer size (in MB)")) {
+    if (CmiMyRank() == 0) {
+      // Round up size to the closest power of 2
+      size_t comm_buffer_size = (size_t)input_comm_buffer_size * 1024 * 1024;
+      int size_log2 = std::ceil(std::log2((double)comm_buffer_size));
+      CsvAccess(gpu_manager).comm_buffer_size = (size_t)std::pow(2, size_log2);
```
nitbhat, line 334: What happens to this value when the user doesn't pass `+gpucommbuffer`? Is there a reasonable default chosen?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -231,8 +312,97 @@ void hapiMapping(char** argv) {
...
+    }
+  }
+
+  if (CmiMyPe() == 0) {
+    CmiPrintf("HAPI> GPU communication buffer size: %lu MB "
+        "(rounded up to the nearest power of two)\n",
+        CsvAccess(gpu_manager).comm_buffer_size / (1024 * 1024));
+  }
+
+  CmiNodeBarrier();
```
nitbhat, line 344: Could you add a comment for why this barrier is needed?

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -519,6 +689,173 @@ hapiWorkRequest::hapiWorkRequest() :
...
+  for (int i = 0; i < CmiNumNodes() / CmiNumPhysicalNodes(); i++) {
+    if (i == CmiMyNodeRankLocal()) continue;
+
+    // Loop through GPU devices per process
+    for (int j = 0; j < CsvAccess(gpu_manager).device_count; j++) {
+      int device_index = CsvAccess(gpu_manager).device_count * i + j;
+      cuda_ipc_device_info& cur_device_info = CsvAccess(gpu_manager).cuda_ipc_device_infos[device_index];
+
+      // Open memory handle
+      cudaIpcMemHandle_t* shm_mem_handle =
```
nitbhat, line 833: If I'm understanding correctly, is the POSIX shm stuff setup for cuda IPC?

src/arch/cuda/hybridAPI/devicemanager.h
```diff
@@ -0,0 +1,65 @@
...
+
+  void* alloc_comm_buffer(size_t size) {
+    return comm_buffer->malloc(size);
+  }
+
+  void free_comm_buffer(size_t offset) {
+    comm_buffer->free((void*)(comm_buffer->base_ptr + offset));
+  }
+
+  size_t comm_buffer_free_size() {
```
nitbhat, line 53: You could add a `get` to this method to since it returns a value. It gives an impression to be some sort of a `free` method.

src/arch/cuda/hybridAPI/buddy_allocator.cpp
```diff
@@ -0,0 +1,195 @@
+#include "buddy_allocator.h"
```
nitbhat, line 1: Can you add a high level summary of how the buddy allocator works?

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: I think these can live here for now since HAPI is the only one using them.

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: This block of code is only called by rank 0 PEs (so once per process).

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: It's 64MB by default (see `gpumanager.h`).

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: This ensures that `comm_buffer_size` is set before PEs (device representative PEs to be exact, one PE per GPU) create device communication buffers on their respective GPUs with `dm->create_comm_buffer` .

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Yes, POSIX shared memory is used to exchange CUDA IPC handles between processes on the same physical node.

src/arch/cuda/hybridAPI/devicemanager.h
minitu: Will do

src/arch/cuda/hybridAPI/buddy_allocator.cpp
minitu: Will do

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Will do

minitu:
@nitbhat Addressed your comments

src/arch/cuda/hybridAPI/hapi_impl.cpp
nitbhat: Okay, sounds good.

src/arch/cuda/hybridAPI/hapi_impl.cpp
nitbhat: Ah okay, I see the `if (CmiMyRank() == 0) {` guard now. Thanks!

src/arch/cuda/hybridAPI/gpumanager.h
```diff
@@ -100,6 +148,19 @@ struct GPUManager {
     // Number of PEs mapped to each device
     pes_per_device = -1;
 
+    // Device communication buffer
+    comm_buffer_size = 1 << 26; // 64MB by default
```
stwhite91, line 152: Is there a way for users to override this?

minitu: Yes there is the `+gpucommbuffer` command line argument

minitu:
@stwhite91 Can I get your approval?

src/arch/cuda/hybridAPI/buddy_allocator.cpp
```diff
@@ -0,0 +1,195 @@
...
+    size_t allocated = 0;
+    size_t used = 0;
+    for (const auto& elem : alloc_map) {
+      const auto& block = elem.second;
+      allocated += block.size;
+      used += block.requested;
+      printf("ptr: %p, size: %lu, req: %lu\n", elem.first, block.size, block.requested);
+    }
+
+    printf("(fragmentation) free: %lu, allocated: %lu, used: %lu\n", free, allocated, used);
```
evan-charmworks, line 32: Here and elsewhere, the format specifier for `size_t` is `%zu` rather than `%lu`.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -231,8 +312,97 @@ void hapiMapping(char** argv) {
...
+  // Set device and store PE-device mapping
   hapiCheck(cudaSetDevice(CpvAccess(my_device)));
+#if CMK_SMP
+  CmiLock(CsvAccess(gpu_manager).device_mapping_lock);
+#endif
+  CsvAccess(gpu_manager).device_map.emplace(CmiMyPe(),
+      &(CsvAccess(gpu_manager).device_managers[CpvAccess(my_device)]));
+#if CMK_SMP
+  CmiUnlock(CsvAccess(gpu_manager).device_mapping_lock);
+#endif
```
evan-charmworks, line 324: Here and elsewhere, is there a drawback to adding `GPUManager & my_gpu_manager = CsvAccess(gpu_manager);` near the top of functions and replacing uses of `CsvAccess(gpu_manager)` with this new local variable? Similarly for other globals with multiple uses in a function.

src/arch/cuda/hybridAPI/buddy_allocator.cpp
minitu: Good point, will fix them

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: Not really, it just seemed like convention to use `CsvAccess` explicitly similar to how it is used elsewhere in the Charm++ source code. I guess it does have the benefit of reminding us that the variable is process-shared, but I am for your suggestion (it was cumbersome to add `CsvAccess` all the time when writing the code).

minitu:
@nitbhat @evan-charmworks @stwhite91 Addressed Evan's comments, please review

PR #2902, opened 2020-06-08
Remove duplication of load balancer count

rbuch:


src/ck-ldb/BaseLB.h
```diff
@@ -151,9 +151,8 @@ class BaseLB: public CBase_BaseLB
   BaseLB(CkMigrateMessage *m):CBase_BaseLB(m) {
     lbmgr = CProxy_LBManager(_lbmgr).ckLocalBranch();
   }
-  virtual ~BaseLB();
+  virtual ~BaseLB() {};
```
epmikida, line 154: remove semicolon

rbuch: Done.

PR #2906, opened 2020-06-09
Adding CkChannel for Go-like Channel Functionality

Wingpad:
First swipe at adding channels to Charm++, adding two constructs: `CkChannel` and `CkMultiChannel`. The first of which is a simple channel, with non-blocking send and blocking receive functionality. The second is, effectively, a 1D array of channels with the aforementioned functionality but adds mechanisms to get responses from multiple channels. Both objects are migratable, singleton chares.

TODOs:
- Add `pup` to channel objects.
- Write documentation for new features/examples.

Making a PR to start getting feedback. Open questions include which other API functions these objects should provide, such as:
- Should these have a "wait for a single value without saving the response" function?
- Should these have a channel closing mechanism?

### DIFF
--- a/examples/channels/multi/multi.C
+++ b/examples/channels/multi/multi.C
@@ -1,0 +1,51 @@
+#include "multi.h"
+#include <cstdlib>
+#include <ctime>
+#include <numeric>
+
+Main::Main(CkArgMsg *m) {
+  mainProxy = thisProxy;
+  numWorkers = CkNumPes();
+  CkPrintf("[MAIN] Creating channel, producer, and consumer.\n");
+  channel = CProxy_CkMultiChannel<2, int>::ckNew();
+  CProxy_JobConsumer consumer = CProxy_JobConsumer::ckNew(channel);
+  CProxy_JobProducer producer = CProxy_JobProducer::ckNew(channel);
+  // Run the consumers and producer
+  consumer.run();
+  producer.run();
+}
+
+void Main::stageI() {
+  CkPrintf("[MAIN] Setting a callback for data from any channel.\n");
+  CkCallback cb(CkIndex_Main::stageII(std::make_pair(0, 0)), mainProxy);
+  int channels[2];
+  std::iota(channels, channels + 2, 0);
+  channel.receiveFromAny(channels, 2, cb);
+  CkPrintf("[MAIN] Sending data to a random channel...\n");
+  std::srand(std::time(nullptr));
+  int ch = std::rand() % 2;
+  channel.send(ch, ch);
+}
+
+void Main::stageII(std::pair<int, int> response) {
+  CkPrintf("[MAIN] Received data from channel %d...\n", response.first);
+  CkExit();
+}
+
+void JobConsumer::run() {
+  CkPrintf("[CONSUMER%d] Received job %d from channel 0.\n", thisIndex, channel.receive(0));
+  channel.send(1, thisIndex);
+}
+
+void JobProducer::run() {
+  CkPrintf("[PRODUCER] Putting %d jobs into channel 0.\n", numWorkers);
+  for (int i = 0; i < numWorkers; i++) {
+    channel.send(0, 2 * (i + 1));
+  }
+  CkPrintf("[PRODUCER] Awaiting completion notices from consumers (on channel 1).\n");
+  channel.waitN(1, numWorkers);
+  mainProxy.stageI();
+}
+
+#include "multi.def.h"
+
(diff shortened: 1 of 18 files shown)


src/libs/ck-libs/channel/channel.h
```diff
@@ -0,0 +1,180 @@
...
+      if (!data[channel].empty()) {
+        return channel;
+      }
+    }
+    return -1;
+  }
+
+  std::pair<int, T> receiveFromAny(int* channels, int numChannels) {
+    int which = checkChannels(channels, numChannels);
+    if (which >= 0) goto END;
```
stwhite91, line 158: Probably nicer to change this to `if (which < 0) { ... }` and get rid of the goto

examples/channels/multi/multi.C
```diff
@@ -0,0 +1,51 @@
+#include "multi.h"
+#include <cstdlib>
+#include <ctime>
+#include <numeric>
+
+Main::Main(CkArgMsg *m) {
```
stwhite91, line 6: Either make this [nokeep] or delete the msg, same for the other example.

src/libs/ck-libs/channel/channel.h
Wingpad: Alright, I put the GOTO there to avoid repeating the following block of code, but if that's not a concern then I'll duplicate it and delete the GOTO.

    auto front = data[which].front();
    data[which].pop();
    return std::make_pair(which, front);

examples/channels/multi/multi.C
Wingpad: I didn't know `nokeep` was an option for the main entry point, I'll go ahead and do that.

src/libs/ck-libs/channel/channel.h
stwhite91: Isn't this equivalent?
```
std::pair<int, T> receiveFromAny(int* channels, int numChannels) {
    int which = checkChannels(channels, numChannels);
    if (which < 0) {
       waiting.waitForAny(channels, numChannels, CthSelf());
       CthSuspend();
       which = checkChannels(channels, numChannels);
    }
    auto front = data[which].front();
    data[which].pop();
    return std::make_pair(which, front);
  }
```

examples/channels/multi/multi.C
```diff
@@ -0,0 +1,51 @@
...
+  consumer.run();
+  producer.run();
+}
+
+void Main::stageI() {
+  CkPrintf("[MAIN] Setting a callback for data from any channel.\n");
+  CkCallback cb(CkIndex_Main::stageII(std::make_pair(0, 0)), mainProxy);
+  int channels[2];
+  std::iota(channels, channels + 2, 0);
+  channel.receiveFromAny(channels, 2, cb);
```
rbuch, line 23: Does this not block?

examples/channels/multi/multi.C
```diff
@@ -0,0 +1,51 @@
...
+  CProxy_JobProducer producer = CProxy_JobProducer::ckNew(channel);
+  // Run the consumers and producer
+  consumer.run();
+  producer.run();
+}
+
+void Main::stageI() {
+  CkPrintf("[MAIN] Setting a callback for data from any channel.\n");
+  CkCallback cb(CkIndex_Main::stageII(std::make_pair(0, 0)), mainProxy);
+  int channels[2];
```
rbuch, line 21: You can use brace initialization instead of `iota` here if you want.

src/libs/ck-libs/channel/channel.C
```diff
@@ -0,0 +1,3 @@
+#include "charm++.h"
```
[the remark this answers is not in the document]
Wingpad: Since the code is heavily templated, most code needs to be in the header at present. `*.C` was included to future-proof/foster possible additions of non-templated code in the correct location. That being said, it can safely be removed from now.

src/libs/ck-libs/channel/channel.h
```diff
@@ -0,0 +1,180 @@
...
+#include <utility>
+#include <memory>
+
+template<typename T>
+class CkChannel : public CBase_CkChannel<T> {
+public:
+  CkChannel() {}
+  CkChannel(CkMigrateMessage*) {}
+
+  void send(T value) {
```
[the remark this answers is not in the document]
Wingpad: Is only one entry method within a chare allowed to run at a time? If so, it should be thread-safe already as these methods are all entry methods.

examples/channels/multi/multi.C
Wingpad: Since it's requesting a callback and not receiving the value in place, it does not block. Perhaps `receiveFromAny` and `receive` when invoking a callback to distinguish them from a blocking call? `receiveViaCallback` gets a bit verbose when it comes to `receiveFromAnyChannelViaCallback`... any other suggestions for a name?

Wingpad: Maybe `receiveAsync` and `receiveAnyChannelAsync`?

PR #2911, opened 2020-06-12
Fix LB invocation

rbuch:
In the LB refactor, LBManager started to call `InvokeLB` directly, but this doesn't work for all existing LBs, so switch back to using barrier receivers to invoke LBs. Also, correct LB creation order and enabling/disabling LBs.


rbuch:
Eventually all of the LB invocation stuff should be handled directly in LBManager. I plan to extricate LocalBarrier from LBManager (they are currently intertwined), and then LBManager itself will have a barrier receiver, which, when called, will call into the appropriate LB. This change is merely an intermediate step to get the LBs working in the meantime.

kavithachandrasekar:
I may have renamed AtSync -> InvokeLB incorrectly in DistBaseLB.C. Instead renaming ProcessAtSync -> InvokeLB might fix the issue with invoking DistributedLB. If this is a simpler fix, we may not need to switch back to local barrier receivers.

rbuch:
> I may have renamed AtSync -> InvokeLB incorrectly in DistBaseLB.C. Instead renaming ProcessAtSync -> InvokeLB might fix the issue with invoking DistributedLB. If this is a simpler fix, we may not need to switch back to local barrier receivers.

I've done this in 9dc6e4d.

PR #2922, opened 2020-06-25
TRAM: Receive messages at logical node level

minitu:
This PR changes adds a message delivery mechanism in TRAM such that messages are delivered to the **nodegroup** on intermediate and destination logical nodes, and data items are delivered locally on the destination **nodegroup** via the `localDeliver` method (instead of the destination **group**). The user can opt in to this node-level implementation at compile time using the `nodeLevel` aggregate attribute parameter, i.e. `[aggregate (nodeLevel : 1)]`

The most significant changes in this PR are:

- Creation of `MeshStreamerNG`, which sets the foundation for a node-level implementation of TRAM
- Messages sent by the existing group-level TRAM (`MeshStreamer`) are now intercepted by `MeshStreamerNG` in the `receiveAlongRoute` entry method. This also facilitates the local delivery (i.e. `localDeliver`) of the data items to the destination PEs as hops between PEs in the same logical node can be avoided.
- charmxi changes needed to accommodate the above changes
- Update to the varTRAM benchmark (largely readability and cleanup)

Message storage and send operations are still performed at group-level (e.g. `storeMessage` and subsequent send methods such as `sendMeshStreamerMessage`), as well as the user interface (i.e. `GroupMeshStreamer` and `ArrayMeshStreamer`).

Sanjay and I have discussed the future implementation of node-level TRAM, which includes 1) moving the message buffers to the nodegroup (`MeshStreamerNG`), and 2) parallelizing the nodegroup's `receiveAlongRoute` function so that multiple data items can be processed concurrently.


src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
```diff
@@ -32,19 +33,27 @@ module NDMeshStreamer {
     entry void receiveAtDestination(MeshStreamerMessageV *msg);
   };
 
+  template<class dtype, class RouterType>
+  nodegroup MeshStreamerNG {
+    entry MeshStreamerNG();
+    entry [immediate] void setGroupProxy(CProxy_MeshStreamer<dtype, RouterType> groupProxy);
```
minitu, line 39: This entry method had to be marked `immediate` to prevent `MeshStreamerNG::receiveAlongRoute` from getting invoked by `MeshStreamer::sendMeshStreamerMessage` before the group proxy is stored in the nodegroup. Although `MeshStreamerNG::setGroupProxy` is invoked right after the nodegroup is created via `ckNew` (see line 1207 of `xi-Entry.C`), without `immediate` it is not ensured that this entry method is executed before other methods of the nodegroup (e.g. `receiveAlongRoute`).

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -785,23 +830,18 @@ receiveAlongRoute(MeshStreamerMessageV *msg) {
   lastDestinationPe = -1;
   for (int i = 0; i < msg->numDataItems; i++) {
     destinationPe = msg->destinationPes[i];
-    if (destinationPe == myIndex_) {
-      //dtype dataItem = msg->getDataItem<dtype>(i);
-      localDeliver(msg->dataItems+msg->template getoffset<dtype>(i),
-                   msg->template getoffset<dtype>(i+1)-msg->template getoffset<dtype>(i),
-                   msg->destObjects[i], msg->sourcePes[i]);
-    }
```
minitu, line 793: This part has been effectively moved to `MeshStreamerNG`.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
```diff
@@ -13,6 +13,7 @@ module NDMeshStreamer {
 
   template<class dtype, class RouterType>
   group [migratable] MeshStreamer {
+    entry void localDeliver(size_t size, const char data[size], CkArrayIndex arrayId, int sourcePe);
```
minitu, line 16: Making `localDeliver` an entry method causes `size` and `data` parameters to switch places (since `size` needs to come before `data`), which can be seen in multiple places of this PR.

minitu:
Ready for review.

src/libs/ck-libs/NDMeshStreamer/DataItemTypes.h
```diff
@@ -2,6 +2,7 @@
 #define DATA_ITEM_TYPES_H
 
 #include <utility>
+#include "ckarray.h"
```
sanjaykale, line 5: What specifically do you need from ckarray.h? (list/state in comments). A pure Charm++ library shouldn't need that, but of course TRAM may need low level access.

minitu: It was for `CkArrayIndex` in `DataItemHandle`. Actually it only needs `ckarrayindex.h` instead of `ckarray.h`, so fixed that and added a comment.

minitu:
Converting to draft to implement a user option to use group-based or nodegroup-based TRAM, as well as performance analysis.

minitu:
Added `nodeLevel` parameter to `[aggregate]` attribute (i.e. `[aggregate (nodeLevel:1)]`) to allow users to opt in to node-level mechanism at compile time. `nodeLevel:0` will use the default group-based mechanism.

minitu:
node-level_1st_n1

The above plot shows results from `examples/charm++/TRAM/aggregateRandomAccessGroup`, run on a single node of OLCF Summit with 6 processes and 7 PEs per process (total 42 PEs = max number of non-SMT cores, .and a global table size of 2^14 * 42 = 688,128 words.

Node-level TRAM currently performs about 10x slower than PE-level TRAM, need to explore why.

minitu:
Nodegroup IDs are not generated from the same stream as chare arrays (their array managers?) and groups:

a1f1: 10, 11
a1f2: 10, 12
a2f1: 15, 16
a2f2: 15, 17
g1f1: 18, 19
g1f2: 18, 20
g2f1: 21, 22
g2f2: 21, 23
g3f1: 24, 25
g3f2: 24, 26
ng1f1: 2, 3 (should be 27)
ng1f2: 2, 4 (should be 28)
ng2f1: 3, 4 (should be 29)
ng2f2: 3, 5 (should be 30)
ng1f1 TRAM: 27
ng1f2 TRAM: 28
ng2f1 TRAM: 29
ng2f2 TRAM: 30

minitu:
The source of the segfault issue that occurs with nodegroup TRAM clients is the following code block (lines 1143-1146 of `src/xlat-i/xi-Entry.C`):
```
void Entry::genTramDefs(XStr& str) {
  ...
  str << "    CkGroupID tramGid;\n"
      << "    tramGid.idx = gId.idx + " << tramInstanceIndex << ";\n"
      << "    " << aggregatorName() << " = (" << aggregatorType() << "*)"
      << " CkLocalBranch(tramGid);\n  }\n";
  ...
}
```

This results in the TRAMized entry method using a simple arithmetic to determine the associated TRAM instance's group ID from the client nodegroup's group ID, as follows:
```
void CProxyElement_Updater::insertData(const CmiUInt8 &key) {
  if (insertData_marshall3TramAggregator == NULL) {
    CkGroupID gId = ckGetGroupID();
    CkGroupID tramGid;
    tramGid.idx = gId.idx + 1; // Arithmetic to determine TRAM instance's group ID from the client's group ID
    insertData_marshall3TramAggregator = (GroupMeshStreamer<CmiUInt8, Updater, SimpleMeshRouter,
        CkIndex_Updater::_callmarshall_insertData_marshall3>*) CkLocalBranch(tramGid);
  }
  insertData_marshall3TramAggregator->insertData(key, ckGetGroupPe());
}
```

While this 'works' for chare array and chare group clients, it doesn't for nodegroups since their group IDs are generated from a different stream. A segfault occurs in the `CkLocalBranch(tramGid)` call above since `tramGid.idx` is invalid.
I tried generating nodegroups' group IDs from the same stream as groups, but that caused a lot of errors including node broadcasts that occur during init.
We agreed that storing the TRAM instance's group ID when it gets created and somehow making that available to the TRAM client is the way to go, with a broadcast being one option.

A TRAM instance gets created through `ckNew` in the `ckNew` call of the client, as follows:
```
CkGroupID CProxy_Updater::ckNew(const CkEntryOptions *impl_e_opts) {
...
    CProxy_GroupMeshStreamer<CmiUInt8, Updater, SimpleMeshRouter,
        CkIndex_Updater::_callmarshall_insertData_marshall3> tramProxy =
    CProxy_GroupMeshStreamer<CmiUInt8, Updater, SimpleMeshRouter,
        CkIndex_Updater::_callmarshall_insertData_marshall3>::ckNew(
            tramNodeProxy, nDims, dims, gId, tramBufferSize, false, 0.01, maxItemsBuffered, thresholdFractionNum,
            thresholdFractionDen, cutoffFractionNum, cutoffFractionDen, nodeLevel);
...
```

Please let me know if you have any ideas on this.

minitu:
@rbuch Any thoughts on this?

rbuch:
It looks like there isn't really a good way of avoiding a broadcast (assuming that using the same stream for groups and nodegroups isn't workable, as you said) with the current schemes of group/nodegroup creation.

One option is to make a custom group (I use that term without loss of generality, all of this equally applies to nodegroups) creation method that lets you explicitly provide the ID to use when creating a group and use some kind of ID space partitioning scheme to ensure that certain segments are dedicated to groups created in service of other groups (right now the `idx` field of `CkGroupID` is a 32 bit int consisting of the concatenation of the PE index and a counter that increments with every group creation, it could be made into a 64 bit value where the top 32 bits are 0 and the bottom 32 what they are right now in the regular case, but a copy of the creating group's ID in this new case and the lower 32 bits a counter incremented by the creating group). This would allow the same kind of arithmetic as is done now, but added to a shifted version of the base ID, e.g. `tramGid.idx = gId.idx << 32 + 1;`.

minitu:
Closing as for nodegroup chares we are moving on to https://github.com/Wingpad/hypercomm-aggregation.

PR #2931, opened 2020-07-02
Add flags for indicating checkpoints and migrations within PUP

rbuch:
Add flags and isCheckpoint() and isMigration() functions to PUP::er so users can conditionally do things based on the purpose of the PUP.


src/util/pup.h
```diff
@@ -229,10 +237,16 @@ class er {
   void becomeUserlevel(void) {PUP_er_state|=IS_USERLEVEL;}
   bool isUserlevel(void) const {return (PUP_er_state&IS_USERLEVEL)!=0?true:false;}
   
-  //This indicates that the pup routine is restoring from a checkpoint.
-  void becomeRestarting(void) {PUP_er_state|=IS_RESTARTING;}
-  bool isRestarting(void) const {return (PUP_er_state&IS_RESTARTING)!=0?true:false;}
-  
+  bool isCheckpoint(void) const
```
stwhite91, line 240: I would think it might be valuable to have separate isRestarting and isCheckpointing, like packing/unpacking. Is isCheckpoint set when restarting from a checkpoint?

rbuch: Yes, isCheckpoint is set both when making a checkpoint and when restarting from a checkpoint. I only found one instance of isRestarting being used in the RTS code, I replaced that with isCheckpoint (in `liveViz`).

evan-charmworks: I like how `isCheckpoint` is orthogonal to which way the data is going. `isPacking && isRestarting` would never happen, and neither would `isUnpacking && isCheckpoint` if the former existed.

rbuch:
This change was spurred by a discussion we had with @nilsdeppe; I want to wait to merge this until he or @moxcodes (who is working on the LB stuff in Spectre has tested this and verified that it works for their use case.

moxcodes:
Hi @rbuch! Thanks for working on this, it seems like a very useful feature for customizing the behavior of the pup functions. I'm working on incorporating a proof-of-concept checkpoint into the system we're using to test load balancing in spectre so that I can verify that each of the cases work as I understand they should. I'll check back in soon, once I've constructed a working test in spectre based on this branch.

moxcodes:
I've verified that this works well in our charm object wrappers, and that I'm able to transmit differing information and differing sizes in the checkpoint and migration scenarios. Just a couple of questions so that we can better design our user code:
- Is it true that the `isMigration` and `isCheckpoint` are universally mutually exclusive, or is that something we should check for safety in our code?
- Will there be any cases where the charm runtime calls the pup when both are false (i.e. would it be safe to assume that if the pup is called when not migrating and not checkpointing that it was called from within our user code?)
Thanks again!

rbuch:
> I've verified that this works well in our charm++ object wrappers, and that I'm able to transmit differing information and differing sizes in the checkpoint and migration scenarios. Just a couple of questions so that we can better design our user code:
> 
>     * Is it true that the `isMigration` and `isCheckpoint` are universally mutually exclusive, or is that something we should check for safety in our code?
> 
>     * Will there be any cases where the charm runtime calls the pup when both are false (i.e. would it be safe to assume that if the pup is called when not migrating and not checkpointing that it was called from within our user code?)
>       Thanks again!

Great, thanks for the feedback.

- As implemented currently, yes, they are mutually exclusive. I think it's safe to say that the runtime code itself will never make calls where both are enabled, and I can't think of any cases where the runtime system would have to break that invariant.
- Yes, it may make calls where both are set to false in a variety of situations, e.g. when marshalling to send an object of that type as an argument to an entry method, when reading/writing from swap for out of core execution with prefetching, etc. These cases likely won't come up unless you explicitly enable them, but I'd be wary to guarantee that it will never happen, so it might be worth doing an assert or something to check for that case or setting your own user flag independently to indicate that it was called from within your user code.

PR #2940, opened 2020-07-03
CI: Add CUDA

evan-charmworks:


.github/workflows/cuda.yml
```diff
@@ -0,0 +1,19 @@
+name: NetLRTS Linux CUDA
```
matthiasdiener, line 1: Maybe add 'buildold' at the end here.

minitu:
@evan-charmworks Thanks for adding this in!

Issue #2536, opened 2019-09-26
RTS flags are incorrectly handled for AMPI progs when run in SMP mode


nitbhat:
It looks like Charm RTS parses RTS flags only on CmiMyRank 0 and not on other ranks. This causes the RTS flag to show up as the application input arg in AMPI for ranks other than 0. (Not AMPI Rank 0, but CmiMyRank 0).  

Test program:
```c
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>

int main(int argc, char **argv)
{
  int my_id; /* process id */
  int p; /* num procs */
  MPI_Init( &argc, &argv );

  MPI_Comm_rank( MPI_COMM_WORLD, &my_id );
  MPI_Comm_size( MPI_COMM_WORLD, &p );

  printf("[Rank: %d][Num procs: %d] argc = %d\n", my_id, p, argc);

  MPI_Finalize();
  return 0;
}
```

Buggy cases:
AMPI Rank 0 (CmiMyRank 0) gets argc as 1, but AMPI Rank 1 (CmiMyRank 1) gets argc as 2. (since +noCMAForZC is not parsed by the RTS). 

```
c455-022[knl](1038)$ ../../../bin/testrun  +p2 ./ampi_arg_test +noCMAForZC ++ppn 2 +pemap 0,1 +commap 2
[Rank: 0][Num procs: 2] argc = 1
[Rank: 1][Num procs: 2] argc = 2
[Partition 0][Node 0] End of program
```

All ranks except AMPI Rank 0 (CmiMyNode 0, CmiMyRank 0) and AMPI Rank 9 (CmiMyNode 1, CmiMyRank 0) get incorrect args i.e argc as 2. 
```
c455-004[knl](1005)$ ../../../bin/testrun  +p18 ./ampi_arg_test +noCMAForZC ++ppn 9 +pemap 0-8,10-18 +commap 9,19
[Rank: 0][Num procs: 18] argc = 1
[Rank: 1][Num procs: 18] argc = 2
[Rank: 2][Num procs: 18] argc = 2
[Rank: 3][Num procs: 18] argc = 2
[Rank: 4][Num procs: 18] argc = 2
[Rank: 5][Num procs: 18] argc = 2
[Rank: 6][Num procs: 18] argc = 2
[Rank: 7][Num procs: 18] argc = 2
[Rank: 8][Num procs: 18] argc = 2
[Rank: 9][Num procs: 18] argc = 1
[...]

PR #2944, opened 2020-07-06
TCharm: Move +vp handling to nodeInit

evan-charmworks:
The old way is a vestige of AMPI multiple worlds, removed in commit
ab97e47ca7fa1819bea179fc4c1cfc2c6676e386.


stwhite91 (dismissed):

matthiasdiener (dismissed):

evan-charmworks:
I need to test if this is a regression regarding #2536.

evan-charmworks:
PR fixed and result confirmed experimentally.

stwhite91 (approved):

src/libs/ck-libs/tcharm/tcharm.C
```diff
@@ -71,6 +76,11 @@ void TCharm::nodeInit()
   // Assumes no anytime migration and only static insertion
   _isAnytimeMigration = false;
   _isStaticInsertion = true;
+
+  char **argv = CkGetArgv();
+  nChunks = CkNumPes();
+  CmiGetArgIntDesc(argv,"-vp",&nChunks,"Set the total number of virtual processors");
+  CmiGetArgIntDesc(argv,"+vp",&nChunks,NULL);
```
matthiasdiener, line 83:
```suggestion
  CmiGetArgIntDesc(argv, "-vp", &nChunks, "Set the total number of virtual processors");
  CmiGetArgIntDesc(argv, "+vp", &nChunks, NULL);
```

src/libs/ck-libs/tcharm/tcharm.C
```diff
@@ -11,7 +11,7 @@ Orion Sky Lawlor, olawlor@acm.org, 11/19/2001
 
 CtvDeclare(TCharm *,_curTCharm);
 
-static int lastNumChunks=0;
+static int nChunks;
```
matthiasdiener, line 14: Any particular reason why this variable and the static variable below are no initialized?

evan-charmworks: They get implicitly set to zero by the BSS segment.

matthiasdiener: Sure, but why not write it down explicitly?

evan-charmworks: Doesn't that increase the size of the .data segment?

matthiasdiener: I think it does, but no idea if that's important or not. This PR is fine for me either way.

matthiasdiener (approved):

stwhite91 (approved):

PR #2946, opened 2020-07-10
Clean up Fortran build targets

evan-charmworks:
Clean up Fortran build targets

Now systems that support Fortran will not ignore failures, and systems
that do not will print fewer "error (ignored)" messages.

- CMake: Copy Fortran .mod files to include/
- f90charm: Always build if Fortran is available
- Add ampi/fallreduce to automated testing


examples/charm++/Makefile
```diff
@@ -33,6 +35,10 @@ DIRS = \
   user-driven-interop \
   zerocopy \
 
+ifeq (1,$(CMK_CAN_LINK_FORTRAN))
+#  DIRS += f90charm
```
stwhite91, line 39: Why is this commented out?

matthiasdiener, line 39: Is this commented intentionally?

evan-charmworks:
The f90charm examples are commented out because they are broken.

Compilation is broken in a way it was not when I worked on them previously. If I understand correctly, Fortran 90 does not support functions with variadic arguments, such as CkPrintf which is used in the examples. The compiler interprets the first argument count and type arrangement as the only valid one for CkPrintf and later uses are considered errors.

If I hack around this and build the examples, they do not complete successfully.

It also looks like I need to update this PR regarding `testp`.

PR #2951, opened 2020-07-13
CI: add shellcheck test and fix several scripts

matthiasdiener:


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,19 @@
...
+    timeout-minutes: 45
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: ShellCheck
+      run: |
+        # TODO: add more scripts
+        shellcheck build buildcmake package-tarball.sh relink.script src/scripts/commitid.sh src/scripts/testrun src/arch/uth/charmrun src/arch/sim/charmrun src/arch/multicore/charmrun src/arch/mpi-win-x86_64/charmrun
```
evan-charmworks, line 19:
```suggestion
        shellcheck \
          build \
          buildcmake \
          package-tarball.sh \
          relink.script \
          src/scripts/commitid.sh \
          src/scripts/testrun \
          src/arch/uth/charmrun \
          src/arch/sim/charmrun \
          src/arch/multicore/charmrun \
          src/arch/mpi-win-x86_64/charmrun \
```

package-tarball.sh
```diff
@@ -62,13 +62,13 @@ git add -f contrib/hwloc/{,src/,include/}Makefile.in
...
 # Get an identifier for the current state of the code
-object_id=`git write-tree`
+object_id=$(git write-tree)
 
 # Construct the target file/folder name
 version="charm-$(cat src/scripts/VERSION)"
 
 # Generate the distribution tarball
-git archive --format=tar.gz --prefix="$version/" -o $version.tar.gz $object_id
+git archive --format=tar.gz --prefix="$version/" -o "$version".tar.gz "$object_id"
```
evan-charmworks, line 71:
```suggestion
git archive --format=tar.gz --prefix="$version/" -o "$version.tar.gz" "$object_id"
```

src/arch/mpi-win-x86_64/charmrun
```diff
@@ -91,7 +89,7 @@ test $QUIET -eq 1 && charm_args=$charm_args" "++quiet
 
 args=$args" "$charm_args
 
-if [ "$DEBUG" = '1' -o "$DEBUG_NO_PAUSE" = '1' ]
+if [ "$DEBUG" = '1' ] || [ "$DEBUG_NO_PAUSE" = '1' ]
```
evan-charmworks, line 92: What is the linting tool's rationale for this? It causes two process launches instead of one.

src/arch/multicore/charmrun
```diff
@@ -73,11 +73,11 @@ done
 
 test $QUIET -eq 1 && args=$args" "++quiet
 
-if [ "$DEBUG" = '1' -o "$DEBUG_NO_PAUSE" = '1' ]
+if [ "$DEBUG" = '1' ] || [ "$DEBUG_NO_PAUSE" = '1' ]
```
evan-charmworks, line 76: Another location of the above

src/arch/mpi-win-x86_64/charmrun
matthiasdiener: The reason is here: https://github.com/koalaman/shellcheck/wiki/SC2166. This does indeed launch two processes instead of one, which is something we could fix with `[[` instead of `[`, but I think that might be better in another PR.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,29 @@
+name: ShellCheck
```
evan-charmworks, line 1: Would it make sense to name this job and yml file something more generic like `Linting` in case we want to add other tools as additional `steps`? It wouldn't make sense to add more CI jobs for them since this one says `Successful in 5s`.

matthiasdiener: We could do that, but there wouldn't be many advantages imho. Adding short jobs is better for parallelism.

evan-charmworks: There is a finite limit on the number of jobs, correct? That is the dominating factor in my mind. I'm not sure how parallelism comes into play when the Shellcheck step only takes five seconds.

matthiasdiener: Yes, there is a limit of concurrent jobs (20 on GitHub). Note that the check takes 5 secs at the moment because there are very few files checked. Checking all our internal scripts (not including hwloc and romio) as I hope to do in the future will probably take several mins on CI. In that case, running (e.g.) cpplint concurrently makes sense I think.

nitbhat:
It looks like the multicore autobuild failure (http://charm.cs.illinois.edu/autobuild/old.2020_07_16__01_01/multicore-linux-x86_64.txt) was introduced by the change in `src/arch/multicore/charmrun` (https://github.com/UIUC-PPL/charm/commit/96f4c66b68ccd1dad530bd2375d3a70893d34caa#diff-daf2ca0c35a49d0eb9cad51795b80fdb)

Previously with `exec $cmd $args` arguments gets passed as:
```
----- Charm++> before ppnSet argc=3, arv[0]=./check, argv[1]=+p4, argv[2]=4
----- Charm++> before plusPSet argc=3, arv[0]=./check, argv[1]=+p4, argv[2]=4

```

With `exec "$cmd" "$args"` arguments gets passed as:
```
----- Charm++> before ppnSet argc=2, arv[0]=./check, argv[1]= +p4 4, argv[2]=(null)
----- Charm++> before plusPSet argc=2, arv[0]=./check, argv[1]= +p4 4, argv[2]=(null)
```

This causes the +p parsing to be ignored from what it looks like. Was that change introduced for any specific reason? Can we revert it?

matthiasdiener:
> It looks like the multicore autobuild failure (http://charm.cs.illinois.edu/autobuild/old.2020_07_16__01_01/multicore-linux-x86_64.txt) was introduced by the change in `src/arch/multicore/charmrun` ([96f4c66#diff-daf2ca0c35a49d0eb9cad51795b80fdb](https://github.com/UIUC-PPL/charm/commit/96f4c66b68ccd1dad530bd2375d3a70893d34caa#diff-daf2ca0c35a49d0eb9cad51795b80fdb))
> 
> Previously with `exec $cmd $args` arguments gets passed as:
> 
> ```
> ----- Charm++> before ppnSet argc=3, arv[0]=./check, argv[1]=+p4, argv[2]=4
> ----- Charm++> before plusPSet argc=3, arv[0]=./check, argv[1]=+p4, argv[2]=4
> ```
> 
> With `exec "$cmd" "$args"` arguments gets passed as:
> 
> ```
> ----- Charm++> before ppnSet argc=2, arv[0]=./check, argv[1]= +p4 4, argv[2]=(null)
> ----- Charm++> before plusPSet argc=2, arv[0]=./check, argv[1]= +p4 4, argv[2]=(null)
> ```
> 
> This causes the +p parsing to be ignored from what it looks like. Was that change introduced for any specific reason? Can we revert it?

This was done to fix a shellcheck warning (which in this case was overeager). I'll prepare a PR.

PR #2954, opened 2020-07-13
Charm++ Benchmarks: Add a testp target for scalable testing on P pes

nitbhat:


nitbhat:
Right now, `P` needs to be passed to the `testp` target through command line. 
Something like `make test P=4` (just like TESTOPTS). 

Is it worth having a default P value?

benchmarks/charm++/zerocopy/Makefile
```diff
@@ -23,6 +31,8 @@ $(foreach i,$(DIRS),build-$i):
 
 $(foreach i,$(DIRS),test-$i):
 	$(MAKE) -C $(subst test-,,$@) test OPTS='$(OPTS)' TESTOPTS='$(TESTOPTS)'
+$(foreach i,$(TESTPDIRS),testp-$i):
```
evan-charmworks, line 34: missing a newline

nitbhat: Good catch, fixed.

PR #2955, opened 2020-07-13
Extract atsync barrier from LBManager

rbuch:


src/ck-core/cksyncbarrier.h
```diff
@@ -0,0 +1,83 @@
...
+
+  void CallReceivers(void);
+  void CheckBarrier(bool flood_atsync = false);
+  void ResumeClients(void);
+
+private:
+  std::list<LBClient*> clients;
+  std::list<LBReceiver*> receivers;
+
+  std::list<int> local_pes_to_notify;
```
stwhite91, line 68: Why a list versus a vector?

rbuch: For `clients` (and `receivers` to some degree), because of random deletion (e.g. when the load balancer migrates AtSync elements away from the PE, the clients are removed and then readded on the receive side). Really, this should be changed to use something similar to what I implemented for the LBDatabase (sort of a linked list embedded inside a vector that keeps track of empty slots and avoids compaction when items are removed from the middle). I'm planning on making that implementation more generic, moving it into `util/` and then using it here.

Semantically, this is actually a set, but I think we can get better performance via a vector.

I've changed what was `local_pes_to_modify` to be a `vector<bool>` now.

src/ck-core/init.C
```diff
@@ -1495,6 +1498,7 @@ void _initCharm(int unused_argc, char **argv)
 		*/
 		_registerCkFutures();
 		_registerCkArray();
+                _registerCkSyncBarrier();
```
minitu, line 1501: Not sure if you want to match the surrounding indentation here

rbuch: Good call, fixed.

PR #2965, opened 2020-07-16
Add testp-tests, testp-examples, and testp-benchmarks targets 

nitbhat:


evan-charmworks:
Both files have this target:

```
test syncfttest mpisyncfttest: LIBS
	$(MAKE) -C ../tests $@
	$(MAKE) -C ../examples $@
	$(MAKE) -C ../benchmarks $@
```

We should add `testp` here, right?

nitbhat:
> Both files have this target:
> 
> ```
> test syncfttest mpisyncfttest: LIBS
> 	$(MAKE) -C ../tests $@
> 	$(MAKE) -C ../examples $@
> 	$(MAKE) -C ../benchmarks $@
> ```
> 
> We should add `testp` here, right?
> 
> ```
> test syncfttest mpisyncfttest testp: LIBS
> ```

Yes, I missed that. Thanks for catching it @evan-charmworks. Fixed in the latest commit.

evan-charmworks:
Should it go in src/scripts/Makefile too?

matthiasdiener:
> Should it go in src/scripts/Makefile too?

I think so, yes. Another option would be to ìnclude` `Makefile.tests` in `src/scripts/Makefile`.

nitbhat:
> > Should it go in src/scripts/Makefile too?
> 
> I think so, yes. Another option would be to `include` `Makefile.tests` in `src/scripts/Makefile`.

This is good idea. I created `cmake/Makefile.tests.common` that contains the common test targets of both the build systems. It is included by both `cmake/Makefile.tests` and `src/scripts/Makefile`. This will avoid the copying of targets to maintain consistency.

src/scripts/Makefile
```diff
@@ -38,32 +38,8 @@ CXXFLAGS=-I.
...
-
-all-test-AMPI-tests: AMPI
-	$(MAKE) -C ../tests/ampi all
-
-all-test-AMPI-examples: AMPI
-	$(MAKE) -C ../examples/ampi all
-
-all-test-AMPI-benchmarks: AMPI
-	$(MAKE) -C ../benchmarks/ampi all
+include ../../cmake/Makefile.tests.common
```
evan-charmworks, line 42: I believe this will fail if `make` is called from the `tmp` symlink folder made directly in `charm`. Is it possible to copy or symlink the file to `tmp` under the old build system instead?

nitbhat: That's a good point. 
However, on testing this, it works fine. I think that is primarily because `tmp` is symlinked to `<arch>/tmp`, `../../cmake/Makefile.tests.common` is still valid.

cmake/Makefile.tests.common
```diff
@@ -0,0 +1,94 @@
+# This file contains common targets for building and testing tests, examples and benchmarks
```
matthiasdiener, line 1: Could add this file to .gitattributes for highlighting.

nitbhat: Done.

PR #2971, opened 2020-07-20
AMPI: Improve ampirun's argument parsing

evan-charmworks:
- Avoid testing "-x" on the command, because it fails on $PATH commands
- Improve argument quoting safety
- Error out if invalid mpirun argument is passed
- Add "--" to stop parsing for mpirun arguments
- Call charmrun from "$(dirname $0)/" and not "./" in case run from $PATH


matthiasdiener:
Shellcheck has a few suggestions: (I would ignore SC1081, but SC2221/SC2222 seem particularly relevant)

```

In ampirun line 5:
progname=$(basename $0)
                    ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
progname=$(basename "$0")


In ampirun line 6:
progdir=$(dirname $0)
                  ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
progdir=$(dirname "$0")


In ampirun line 79:
        cmd="$@ "
            ^---^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In ampirun line 83:
      -*|--*)
      ^-- SC2221: This pattern always overrides a later one on line 83.
         ^-^ SC2222: This pattern never matches because of a previous pattern on line 83.


In ampirun line 89:
        cmd="\"$arg\" $@ "
            ^------------^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In ampirun line 116:
Do() {
^-- SC1081: Scripts are case sensitive. Use 'do', not 'Do' (or quote if literal).


In ampirun line 117:
  [[ $verbose == 1 ]] && echo "$progname: Executing $@" 1>&2
                                                    ^-- SC2145: Argument mixes string and array. Use * or separate argument.


In ampirun line 121:
Do "$progdir/charmrun" $charmrun_args $cmd
^-- SC1081: Scripts are case sensitive. Use 'do', not 'Do' (or quote if literal).
                       ^------------^ SC2086: Double quote to prevent globbing and word splitting.
                                      ^--^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
Do "$progdir/charmrun" "$charmrun_args" "$cmd"

For more information:
  https://www.shellcheck.net/wiki/SC1081 -- Scripts are case sensitive. Use '...
  https://www.shellcheck.net/wiki/SC2145 -- Argument mixes string and array. ...
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
```

matthiasdiener:
> * Improve argument quoting safety
> * Call charmrun from "$(dirname $0)/" and not "./" in case run from $PATH

I agree with these changes.

> * Avoid testing "-x" on the command, because it fails on $PATH commands
> * Error out if invalid mpirun argument is passed
> * Add "--" to stop parsing for mpirun arguments

I'm not so sure about these changes. How are other charmrun arguments (e.g. `+balancer RotateLB`) handled now?

evan-charmworks:
> I'm not so sure about these changes. How are other charmrun arguments (e.g. `+balancer RotateLB`) handled now?

Any Charmrun option not understood by ampirun such as `++debug` will be passed in place to Charmrun, where it will work correctly. Charmrun parses the entire command line and eats any arg it understands, unlike mpirun. (`+balancer RotateLB` is a Charm program arg, not a Charmrun arg, so it is also passed in place.)

> > * Avoid testing "-x" on the command, because it fails on $PATH commands
> 
> Imho, it would be easier to add a check if a command is in $PATH to the -x check (which seems to be the main reason for the three changes above).

I can retry with a `command -v` check to see if this fixes what I encountered.

> > * Error out if invalid mpirun argument is passed
> > * Add "--" to stop parsing for mpirun arguments
> 
> "--" is non-standard and does not work with e.g. mpich's mpirun.

These match Open MPI's mpirun.

matthiasdiener:
> > I'm not so sure about these changes. How are other charmrun arguments (e.g. `+balancer RotateLB`) handled now?
> 
> Any Charmrun option not understood by ampirun such as `++debug` will be passed in place to Charmrun, where it will work correctly. Charmrun parses the entire command line and eats any arg it understands, unlike mpirun. (`+balancer RotateLB` is a Charm program arg, not a Charmrun arg, so it is also passed in place.)

Ok, after testing a few examples, it seems that the behavior is mostly the same. A minor difference is e.g.

```
$ ./ampirun -n 4 -v ++debug --help ./pgm
```

which shows ampirun's help in the old version, but executes ./pgm in the new one. I'm not sure it is that important though.

> I can retry with a command -v check to see if this fixes what I encountered.

No need I think. This code is simpler.

> > > * Error out if invalid mpirun argument is passed
> > > * Add "--" to stop parsing for mpirun arguments
> > 
> > 
> > "--" is non-standard and does not work with e.g. mpich's mpirun.
> 
> These match Open MPI's mpirun.

Right. What I wanted to express here was that we can't rely on "--" being present, as it is non-standard (and seems to be undocumented even for OpenMPI). I think it is good to support it.

evan-charmworks:
> Right. What I wanted to express here was that we can't rely on "--" being present, as it is non-standard (and seems to be undocumented even for OpenMPI). I think it is good to support it.

I think I see now: There is not much point in implementing `--` in ampirun if we can't guarantee that we can pass what follows to e.g. mpirun without also needing `--`.

matthiasdiener:
> > Right. What I wanted to express here was that we can't rely on "--" being present, as it is non-standard (and seems to be undocumented even for OpenMPI). I think it is good to support it.
> 
> I think I see now: There is not much point in implementing `--` in ampirun if we can't guarantee that we can pass what follows to e.g. mpirun without also needing `--`.

Right, but I still think this patch is worthwhile by explicitly filtering the `--` before passing it to charmrun or mpirun.

evan-charmworks:
Okay, sounds good. Is there anything else we should change in this PR?

matthiasdiener:
> Okay, sounds good. Is there anything else we should change in this PR?

I think not. If you are ok with the slight behavior change that I mentioned above, this patch looks ok to me.

PR #2979, opened 2020-07-22
cmake: implement all-test

matthiasdiener:


.github/workflows/multicore_darwin.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: build
-      run: ./build all-test multicore-darwin-x86_64 -g -j3 --with-production --enable-tracing
+      run: ./build LIBS multicore-darwin-x86_64 -g -j3 --with-production --enable-tracing
```
rbuch, line 14: Why is this changed to LIBS?

matthiasdiener: Because we need to add "-tracemode projections" to the build command for the tests, which the build command doesn't do (and the make commands below wouldn't rebuild the tests since all-tests was done already).

CMakeLists.txt
```diff
@@ -151,7 +151,14 @@ endif()
 set(CMK_REPLAYSYSTEM ${REPLAY})
 
 if(${TARGET} STREQUAL "all-test")
-  message(WARNING "Setting TARGET to LIBS instead of all-test.")
+  add_custom_target(all-test ALL COMMAND make -C ${CMAKE_BINARY_DIR}/tmp all-test)
+  add_dependencies(all-test moduleampi ampi-compat ck modulecompletion
+        conv-partition conv-util hwloc conv-static conv-core ldb-rand
+        memory-os-isomalloc memory-default threads-default ckmain moduletcharmmain
+        conv-machine tmgr conv-ldb ckqt tcharm-compat moduleNDMeshStreamer create_symlinks moduleCkCache trace-converse moduleCommonLBs moduleTreeLB moduleCkMulticast moduleCkIO conv-cpm memory-os-wrapper threads-default-tls ldb-neighbor ldb-workstealing modulearmci modulecollidecharm modulecollide memory-os)
+  if(${CMK_CAN_LINK_FORTRAN})
```
evan-charmworks, line 159: Should this be `if(CMK_CAN_LINK_FORTRAN)`?

CMakeLists.txt
```diff
@@ -151,7 +151,14 @@ endif()
 set(CMK_REPLAYSYSTEM ${REPLAY})
 
 if(${TARGET} STREQUAL "all-test")
-  message(WARNING "Setting TARGET to LIBS instead of all-test.")
+  add_custom_target(all-test ALL COMMAND make -C ${CMAKE_BINARY_DIR}/tmp all-test)
```
evan-charmworks, line 154: Does this have the drawback that it forces the dependent make to run as `-j1`?

matthiasdiener: No, this will run in parallel.

CMakeLists.txt
matthiasdiener: Fixed.

evan-charmworks:
Is it possible for the normal make output to be printed so there is some indication of what is going on while the tests are building?

matthiasdiener:
> Is it possible for the normal make output to be printed so there is some indication of what is going on while the tests are building?

In the current implementation, only when running verbosely (i.e., `make VERBOSE=1` or `./build ... -v`)

matthiasdiener:
> > Is it possible for the normal make output to be printed so there is some indication of what is going on while the tests are building?
> 
> In the current implementation, only when running verbosely (i.e., `make VERBOSE=1` or `./build ... -v`)

Please dont merge this yet, Ill try to find a better solution for this.

matthiasdiener:
> > > Is it possible for the normal make output to be printed so there is some indication of what is going on while the tests are building?
> > 
> > 
> > In the current implementation, only when running verbosely (i.e., `make VERBOSE=1` or `./build ... -v`)
> 
> Please dont merge this yet, Ill try to find a better solution for this.

Couldn't find a better solution unfortunately.

evan-charmworks:
Does all-test work on Windows?

matthiasdiener:
> Does all-test work on Windows?

Not sure, since we haven't merged cmake Windows support yet, but I would expect it to be supported.

.travis.yml
```diff
@@ -20,16 +20,18 @@ matrix:
     dist: xenial
     script:
     - sudo apt-get update && sudo apt-get install -y mpich libmpich-dev
-    - ./build LIBS mpi-linux-x86_64 smp -j3 --with-production --enable-error-checking --enable-lbuserdata -g -Werror=vla || exit 1
-    - make -C mpi-linux-x86_64-smp/tests TESTOPTS="+setcpuaffinity"
```
matthiasdiener, line 24: seems like we were never actually testing this build...

matthiasdiener:
Ready for review.

PR #2981, opened 2020-07-22
github-autobuild: change to build once a day

matthiasdiener:


evan-charmworks:
Should this be at 01:00 to match current autobuild?

matthiasdiener:
> Should this be at 01:00 to match current autobuild?

Not sure. We might want to run it at a slightly different time in order not to overlap with the current autobuild.

.github/workflows/autobuild.yml
```diff
@@ -2,7 +2,7 @@ name: Autobuild
 
 on:
   schedule:
-    - cron: '59 * * * *'
+    - cron: '01 05 * * *' # UTC, corresponds to 00:01 Central Time
```
rbuch, line 5: To be pedantic, if this is UTC, then it should vary between 23:01 and 00:01 (CST and CDT, respectively).

matthiasdiener: good point, fixed.

PR #2984, opened 2020-07-23
TRAM: Restore support for chare groups

minitu:
Restores `GroupMeshStreamer` to revive support for chare groups, which was removed in #2775.


minitu:
Moved common code in `insertData()` methods of `ArrayMeshStreamer` and `GroupMeshStreamer` that increments completion detection and quiescence detection into `createDetectors()`.

minitu:
Confirmed that both `aggregateRandomAccessArray` and `aggregateRandomAccessGroup` works now.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
```diff
@@ -33,14 +33,24 @@ module NDMeshStreamer {
   };
 
   template<class dtype, class ClientType, class RouterType, int (*EntryMethod)(char *, void *) = defaultMeshStreamerDeliver<dtype,ClientType> >
-    group [migratable] ArrayMeshStreamer :
-    MeshStreamer<dtype, RouterType> {
+  group [migratable] GroupMeshStreamer : MeshStreamer<dtype, RouterType> {
+    entry GroupMeshStreamer(
+        int numDimensions, int dimensionSizes[numDimensions],
```
rbuch, line 38: Perhaps this and the corresponding block for `ArrayMeshStreamer` can be unified as a #define or something since they seem to be identical. There may also be a way to bake these default params into the parent class, but perhaps that gets wonky since this is a CI.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -78,6 +78,8 @@ template <class dtype>
 struct DataItemHandle {
   CkArrayIndex arrayIndex;
   const dtype *dataItem;
+
+  DataItemHandle(dtype* _ptr) : dataItem(_ptr) {}
```
rbuch, line 82: Maybe just reverse the order of the parameters of the other constructor and add a default parameter for `_idx`?

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
```diff
@@ -631,15 +634,18 @@ storeMessage(int destinationPe, const Route& destinationRoute,
...
-    detectorLocalObj_->produce();
-  }
+  // Increment completion detection and quiescence detection
+  if (useCompletionDetection_) detectorLocalObj_->produce();
+  QdCreate(1);
+}
+
+template <class dtype, class RouterType>
+inline void MeshStreamer<dtype, RouterType>::
+insertData(const DataItemHandle<dtype> *dataItemHandle, int destinationPe) {
```
rbuch, line 648: Does this function need to call `if (useCompletionDetection_) detectorLocalObj_->produce();`? It seems like the old array code would call that once in its own `insertData` and then again in this one.

minitu: So the old code had this weird way of creating completion and quiescence detections for group chares in `MeshStreamer::insertData` (which is eventually invoked by `GroupMeshStreamer` as it didn't have its own `insertData` function), whereas for chare arrays it was done in `ArrayMeshStreamer::insertData`. Since I added the `GroupMeshStreamer::insertData` in this PR (which creates the detectors like `ArrayMeshStreamer::insertData`), `MeshStreamer::insertData` shouldn't create the detectors anymore.

minitu:
@rbuch Addressed your comments

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.h
rbuch: I'm still a bit confused. In the old version, it looks like `ArrayMeshStreamer::insertData()` called `produce()`, then it calls `MeshStreamer::insertData()` (on line 1190), which called `produce()` again. This PR would reduce the number of calls to `produce()` but not to `consume()` for that path way, it seems, so they may not match anymore. Am I missing something here?

minitu: Sorry by the old code I meant the version before Venkat's varTRAM patch (e.g. v6.9.0) that removed `GroupMeshStreamer`. Before the varTRAM patch there were two `MeshStreamer::insertData` functions (one that takes `const void *dataItemHandle` and another that takes `const dtype& dataItem`), and the second one called `produce()` and `QdCreate()` and then called the first `MeshStreamer::insertData` again. `ArrayMeshStreamer::insertData` called the first `MeshStreamer::insertData` directly (which doesn't call `produce()`, etc.) so only  one call to `produce()` and `QdCreate()` was made, which is correct. So Venkat's patch actually changed the `ArrayMeshStreamer::insertData` to call `produce()` two times, which is incorrect. FYI, `GroupMeshStreamer` called the second `MeshStreamer::insertData` before the varTRAM patch to call `produce()` and `QdCreate()`, and my PR effectively does the same by adding a separate GroupMeshStreamer::insertData that calls `createDetectors`.

rbuch: Okay, gotcha, thanks for the explanation.

src/libs/ck-libs/NDMeshStreamer/NDMeshStreamer.ci
minitu: I should mention that the default parameters are determined in the charmxi code (`xi-Entry.C`, IIRC) so having them in the CI file and `NDMeshStreamer.h` is redundant. These are removed in this PR.

PR #2986, opened 2020-07-23
added a message about break/continue

Wingpad:
A minor doc update to make it explicit that SDAG loops do not have a notion of break/continue at this current time. Another natural spot to add this may be under the FAQs section, a la:
“Can I use break/continue within SDAG loops?”
“at this time, break and continue are not supported within SDAG loops. Using them, either within SDAG or “serial” blocks, will result in a compilation error. In the future, these may be added as a function call (e.g. CkSdagBreak or CkSdagContinue) to be called within serial blocks or, explicitly, added to charmxi’s syntax.”


doc/charm++/manual.rst
```diff
@@ -1923,7 +1923,9 @@ instead use a for loop:
 Note that ``int iter;`` is declared in the chare’s class definition and
 not in the ``.ci`` file. This is necessary because the Charm++ interface
 translator does not fully parse the declarations in the ``for`` loop
-header, because of the inherent complexities of C++.
+header, because of the inherent complexities of C++. Also, ``break`` and
+``continue`` statements are not supported within SDAG loops, either
+within ``serial`` blocks or otherwise.
```
stwhite91, line 1928: Within serial blocks, users are free to write whatever valid C++ they want, including continue and break statements. We just want to clarify that there is no breaking or continuing from an SDAG loop.

Issue #2980, opened 2020-07-22
Jacobi2D AMR example failed to run


ouankou:
Hello! I built Charm++, AMR lib, and Jacobi2D in order from this repo on Ubuntu 18.04 and the compilation works well.
The building command is `./build charm++ multicore-linux-x86_64`.
However, executing `./charmrun ./jacobi` under the example folder leads to the following error.

```
Running command: ./jacobi

Charm++> No provisioning arguments specified. Running with a single PE.
         Use +auto-provision to fully subscribe resources or +p1 to silence this message.
Charm++: standalone mode (not using charmrun)
Charm++> Running in Multicore mode: 1 threads (PEs)
Converse/Charm++ Commit ID: v6.11.0-devel-317-gfe43a4dcd
CharmLB> Load balancer assumes all CPUs are same.
Charm++> Running on 1 hosts (1 sockets x 6 cores x 2 PUs = 12-way SMP)
Charm++> cpu topology info is gathered in 0.006 seconds.
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: Cannot insert array element twice!
[0] Stack Traceback:
  [0:0] jacobi 0x55dc0ee36aec CmiAbortHelper(char const*, char const*, char const*, int, int)
  [0:1] jacobi 0x55dc0ee36bc6 CmiGetNonLocal
  [0:2] jacobi 0x55dc0ed12435 CkLocMgr::addElementToRec(CkLocRec*, CkArray*, CkMigratable*, int, void*)
  [0:3] jacobi 0x55dc0ed12399 CkLocMgr::addElement(CkArrayID, CkArrayIndex const&, CkMigratable*, int, void*)
  [0:4] jacobi 0x55dc0ece6110 CkArray::insertElement(CkArrayMessage*, CkArrayIndex const&, int*)
  [0:5] jacobi 0x55dc0ece5f95 CkArray::insertElement(CkMarshalledMessage&&, CkArrayIndex const&, int*)
  [0:6] jacobi 0x55dc0ecead6f CkIndex_CkArray::_call_insertElement_marshall2(void*, void*)
  [0:7] jacobi 0x55dc0eccfb26 CkDeliverMessageFree
  [0:8] jacobi 0x55dc0eccfd11
  [0:9] jacobi 0x55dc0eccfd3c
  [0:10] jacobi 0x55dc0ecd0dce
  [0:11] jacobi 0x55dc0ecd245d CkSendMsgBranchInline
[...]

PR #2989, opened 2020-07-23
AMR: Fix hashing and update style of custom array index

rbuch:
Use placement new to put custom BitVec array index in the index member
of CkArrayIndex.


rbuch:
Fixes #2980.

Issue #2839, opened 2020-05-07
FIx CMake errors/unsupported configs


matthiasdiener:
- `gni-*`
- `*-crayx?`
- Windows

evan-charmworks:
Passing `smp` as part of the build tuplet instead of as a separate word is silently accepted but does not actually build in SMP mode. For example, `./build AMPI mpi-linux-x86_64-smp -j4 -g3` and `./build AMPI netlrts-linux-x86_64-smp -j4 -g3`.

matthiasdiener:
> Passing `smp` as part of the build tuplet instead of as a separate word is silently accepted but does not actually build in SMP mode. For example, `./build AMPI mpi-linux-x86_64-smp -j4 -g3` and `./build AMPI netlrts-linux-x86_64-smp -j4 -g3`.

This seems to be an undocumented feature.
Are there any users/use cases for this?

epmikida:
The `-no-use-new-std` (and probably the `-use-new-std` arguments to build seem to be ignored. I'm working on a project that uses a std later than c++11 but when I build charm with the cmake script, it results in `charmc` always using the `--std=c++11` flag which crashes my build.

Here's the build command I used:
`./build charm++ mpi-linux-x86_64 --build-shared -no-use-new-std -j8`

Whereas if I switch to `./buildold` with the same command I get a `charmc` that works.

matthiasdiener:
> The `-no-use-new-std` (and probably the `-use-new-std` arguments to build seem to be ignored. I'm working on a project that uses a std later than c++11 but when I build charm with the cmake script, it results in `charmc` always using the `--std=c++11` flag which crashes my build.
> 
> Here's the build command I used:
> `./build charm++ mpi-linux-x86_64 --build-shared -no-use-new-std -j8`
> 
> Whereas if I switch to `./buildold` with the same command I get a `charmc` that works.

I can't reproduce the issue. When using the old build system with `./buildold charm++ netlrts-linux-x86_64 -j8 -g --with-production  -no-use-new-std`, I get the following `charmc` behavior (gcc 5.4.0):

```bash
$ charmc -c hello.C -verbose
charmc: Executing g++ -DCMK_GFORTRAN -I../../../bin/../include -D__CHARMC__=1
-U_FORTIFY_SOURCE -fno-stack-protector -std=c++11 -fno-lifetime-dse 
-c hello.C -o hello.o

$ charmc -c hello.C -verbose -no-use-new-std
[...]

PR #2992, opened 2020-07-24
Support: Warn not to use unoptimized build for benchmarking

rbuch:


rbuch:
Adding the color codes may add some extra noise when output is redirected, but I think it's still worth it to make sure it's easily seen by users (and it still shows up when `cat`ing the file). We can also do it selectively be detecting if `stdout` is a tty or not, but making that fully portable isn't entirely trivial.

matthiasdiener:
Imho, since we are not using color anywhere else in Charm++, it might not be necessary here either. How does this handle running with ++quiet?

rbuch:
> Imho, since we are not using color anywhere else in Charm++, it might not be necessary here either. How does this handle running with ++quiet?

It doesn't get printed with `++quiet` (neither does the rest of the preamble).

I'm not married to the color idea, I just thought it'd be good to make it as evident as possible so that people would actually notice it.

matthiasdiener:
> > Imho, since we are not using color anywhere else in Charm++, it might not be necessary here either. How does this handle running with ++quiet?
> 
> It doesn't get printed with `++quiet` (neither does the rest of the preamble).

Ok.

> I'm not married to the color idea, I just thought it'd be good to make it as evident as possible so that people would actually notice it.

Imho, we can't use color here. Not everyone uses white-on-black terminals, and depending on the color scheme the user has chosen that warning might be almost invisible. It is also not super portable (color in Windows terminals is weird, and we have seen issues in the cmake build where the color deactivation failed for some reason, which leads to *all* text printed in yellow).

I'd also argue that the warning is a bit too verbose, and would prefer just attaching `(debug build)` or `(production build)` to the end of the Charm++ version printed just before these warnings, but I'm happy to be overruled on that.

rbuch:
> > > Imho, since we are not using color anywhere else in Charm++, it might not be necessary here either. How does this handle running with ++quiet?
> > 
> > 
> > It doesn't get printed with `++quiet` (neither does the rest of the preamble).
> 
> Ok.
> 
> > I'm not married to the color idea, I just thought it'd be good to make it as evident as possible so that people would actually notice it.
> 
> Imho, we can't use color here. Not everyone uses white-on-black terminals, and depending on the color scheme the user has chosen that warning might be almost invisible. It is also not super portable (color in Windows terminals is weird, and we have seen issues in the cmake build where the color deactivation failed for some reason, which leads to _all_ text printed in yellow).

That's a fair concern, I'll take the color stuff out.
 
> I'd also argue that the warning is a bit too verbose, and would prefer just attaching `(debug build)` or `(production build)` to the end of the Charm++ version printed just before these warnings, but I'm happy to be overruled on that.

I agree that the current warning is on the verbose side, I was coming from the position that if our concern was people using these builds for benchmarks, then we should be specific about telling them not to and giving them instructions on how to build to do so. That being said, perhaps just saying whether the build is a Debug or Release build is sufficient. I'm interested in hearing other opinions on this as well.

minitu:
I think the outputs are still hard to catch, can we add some prefix like `WARNING:` to make it more visible?

evan-charmworks:
@matthiasdiener It looks like CMake does not enable error checking by default like the old build system does.

matthiasdiener:
> @matthiasdiener It looks like CMake does not enable error checking by default like the old build system does.

CC: #2839

PR #2994, opened 2020-07-29
cmake: add amr

matthiasdiener:
Optional target, not built unless you do `cd builddir && make amr` (similar behavior as the old build system).


matthiasdiener:
Ignore, AMR isn't build with the old build system either.

matthiasdiener:
Reopening as optional target.

matthiasdiener:
Needs another review.

PR #2996, opened 2020-07-29
cmake: fix version check for 3.11

matthiasdiener:


CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ message("# with the below configuration information.                     #")
 message("# You can run './buildold' to use the previous build system.    #")
 message("#################################################################")
 
-if(CMAKE_VERSION VERSION_GREATER 3.11)
+if(CMAKE_VERSION VERSION_GREATER 3.12 OR CMAKE_VERSION VERSION_EQUAL 3.12)
```
rbuch, line 24: Why is this necessary?

matthiasdiener: 3.12.0 does not match `VERSION GREATER 3.12`, only 3.12.1+ do. `VERSION_GREATER_EQUAL` is only available in cmake 3.7+, while we support 3.4+. 🤷

PR #3000, opened 2020-07-30
Make relink.script more resilient

rbuch:
If one of subdirs was missing, rm would fail to remove any of
them, so now rm each member of subdirs individually.

### DIFF
--- a/relink.script
+++ b/relink.script
@@ -23,11 +23,12 @@ fi
 
 subdirs="bin include lib lib_so tmp"
 
+# remove symbolic link
+rm -f $subdirs
+
 for i in $subdirs
 do
-  # remove symbolic link
-  rm -f "$i"
-  # skip nonexistent subdir
+# skip nonexistent subdir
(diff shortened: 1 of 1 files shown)


relink.script
```diff
@@ -23,12 +23,11 @@ fi
 
 subdirs="bin include lib lib_so tmp"
 
-# remove symbolic link
-rm -f "$subdirs"
```
matthiasdiener, line 27: I think it is easier to just remove the `"` from here. This is a mistake introduced by #2951 .

rbuch: Done.

rbuch: It seems like doing this causes it to fail the ShellCheck test. Is there a simple way around this?

matthiasdiener: Can you do the same fix as in https://github.com/UIUC-PPL/charm/commit/c30789cae98a71582c20edf10fc29b3497d3fdf2 ? I'm without internet at the moment, so I can't add this myself.

PR #3002, opened 2020-08-02
Verbs Machine Layer: Add incorrectly removed rdma ack handling code

nitbhat:
This code is exercised when sending large messages through the regular API (from LrtsSendFunc)


jcphill:
This one actually fixes the leak in NAMD.

PR #3007, opened 2020-08-04
Intra-node GPU messaging: Clean up stray shared memory file at init, add +gpushm option

minitu:
- If an execution using POSIX shared memory fails, subsequent runs failed as well because the shared memory file was not properly deleted. This PR addresses this issue by checking if such a file exists and deleting it before creating a new file.

- Changed POSIX shared memory optimizations to **not** be applied by default, as it takes up memory space on the GPU.

- Added the `+gpushm` option, which allows the user to opt in to using POSIX shared memory optimizations for inter-process GPU messaging. Otherwise the inter-process messaging mechanism with `nocopydevice` falls back to a naive host-staged method, which has been utilized for inter-node messaging.

- Changed CUDA IPC event pool to be exclusive to each PE (which was the intended design), instead of it being shared by multiple PEs.


src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -131,28 +136,13 @@ void hapiInit(char** argv) {
...
   }
 
-  // Ensure CUDA IPC handles are available for all processes
-  // Note: Causes a hang when this barrier is placed after CPU topology initialization
-  // FIXME: This only needs to be a host-wide synchronization
-  CmiBarrier();
-
-  if (CmiMyRank() == 0) {
-    ipcHandleOpen(); // Open CUDA IPC handles for accessing other processes' device memory
-  }
```
minitu, line 155: Moved to `shmInit`, so that these won't be executed when the user passes `+gpunoshm`.

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -329,58 +319,78 @@ static void hapiMapping(char** argv) {
...
   }
 
-  CmiNodeBarrier(); // Ensure device communication buffer size is set
+  CmiNodeBarrier();
 
-  // Create device communication buffers
-  // Should only be done by device representative threads
-  if (cpv_device_rep) {
-    DeviceManager* dm = csv_gpu_manager.device_map[CmiMyPe()];
+  if (csv_gpu_manager.use_shm) {
```
minitu, line 339: Device communication buffers and CUDA IPC event pool are only created when the user does not opt out of shared memory optimizations.

src/ck-core/ckrdmadevice.C
```diff
@@ -95,59 +96,45 @@ bool CkRdmaDeviceIssueRgets(envelope *env, int numops, void **arrPtrs, int *arrS
...
-      default:
-        {
-          CkAbort("Invalid mode");
-          break;
-        }
+    if (mode == CkNcpyModeDevice::MEMCPY) {
+      // Directly invoke memcpy from source buffer to destination buffer
+      hapiCheck(cudaMemcpyAsync((void*)dest.ptr, source.ptr, dest.cnt,
+            cudaMemcpyDeviceToDevice, postStructs[i].cuda_stream));
+    } else if (mode == CkNcpyModeDevice::IPC && csv_gpu_manager.use_shm) {
```
minitu, line 103: POSIX shared memory is only used when the user doesn't opt out of it.

src/ck-core/ckrdmadevice.C
```diff
@@ -178,13 +165,14 @@ int CkRdmaGetDestPEChare(int dest_pe, void* obj_ptr) {
...
-  int pool_size = CsvAccess(gpu_manager).cuda_ipc_event_pool_size;
+  int pool_size = CsvAccess(gpu_manager).cuda_ipc_event_pool_size_pe;
+  int pool_start = CkMyRank() * pool_size;
   int device_index = dm->global_index;
   cuda_ipc_device_info& my_device_info = CsvAccess(gpu_manager).cuda_ipc_device_infos[device_index];
 
   // Free IPC events that are complete
   // TODO: Don't do this every time but only when the event pool is somewhat empty
-  for (int i = 0; i < pool_size; i++) {
+  for (int i = pool_start; i < pool_start + pool_size; i++) {
```
minitu, line 175: This changes the CUDA IPC event pool to be exclusive to be PE, instead of the pool being shared by multiple PEs. This should reduce contention, but also limits the number of IPC events that can be used by a PE. This shouldn't be an issue unless the number of messages being sent by each PE are very disproportionate.

PR #3012, opened 2020-08-10
Initial cleanup of LGTM alerts

AjayMT:
This is a first pass at resolving the LGTM errors reported [here](https://lgtm.com/projects/g/UIUC-PPL/charm?mode=list). Still in progress.


src/conv-ldb/generate.C
```diff
@@ -122,7 +122,7 @@ n -= (V-1);
 }
 #endif
 
-static void AddEdges(VerticesListType *graph, EdgeListType *EdgeList, int V, int n)
+static void AddEdges(VerticesListType *graph, EdgeListType *EdgeList, int v, int n)
```
rbuch, line 125: Since we're bulk renaming stuff here anyway, it might be worth changing their names to things that make more semantic sense, like changing this from `V` to `numvertices` or something of that ilk.

src/ck-core/ckcheckpoint.C
```diff
@@ -214,7 +214,7 @@ public:
 };
 
 // broadcast
-void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback cb, bool _requestStatus){
+void CkCheckpointMgr::Checkpoint(const char *dirname, CkCallback _cb, bool _requestStatus){
```
evan-charmworks, line 217: In this file, maybe the cb local variables can stay named what they are, (without underscore), since naming a local CkCallback this way is a common convention throughout the code. Instead, the global variable `cb` should be renamed to something else. I suggest `restartMainCB`.

src/conv-core/threads.C
```diff
@@ -456,7 +457,9 @@ static void *CthAllocateStack(CthThreadBase *th, int *stackSize, int useMigratab
   if (*stackSize==0) *stackSize=CthCpvAccess(_defaultStackSize);
   th->stacksize=*stackSize;
   if (!useMigratable || !CmiIsomallocEnabled()) {
-    ret=malloc(*stackSize); 
+    CmiEnforce(*stackSize <= std::numeric_limits<size_t>::max());
```
evan-charmworks, line 460: Unfortunately this line is a tautological comparison. A `size_t` will always be <= the max value possible for the type. LGTM did not consider this to fix the alert. I suggest we use the max value of `uint32_t` here instead of `size_t`, since a 4 GB stack is absurdly large.

src/util/pup_stl.h
```diff
@@ -127,7 +128,9 @@ namespace PUP {
     size_t nChar=v.length();
     p|nChar;
     if (p.isUnpacking()) { //Unpack to temporary buffer
+      CmiEnforce(nChar <= std::numeric_limits<size_t>::max());
```
evan-charmworks, line 131: Somehow LGTM _did_ consider this to fix the alert, even though the malloc above did not work. This confuses me, but we can leave it.

src/conv-core/threads.C
evan-charmworks: Actually, I see now that `stackSize` is an `int`. If you change `size_t` to `int` the warning should go away.

evan-charmworks:
```suggestion
    CmiEnforce(*stackSize <= std::numeric_limits<int>::max());
```

src/conv-core/msgmgr.C
```diff
@@ -168,7 +169,9 @@ CmmTable CmmPup(pup_er p, CmmTable t, CmmPupMessageFn msgpup)
       int ntags, *tags;
       void *msg;
       pup_int(p, &ntags);
-      tags = (int*) malloc(ntags*sizeof(int));
+      CmiEnforce(ntags <= std::numeric_limits<int>::max());
```
matthiasdiener, line 172: Would this condition ever not be true? (same for threads.C, pup_stl.h)

src/conv-core/msgmgr.C
```diff
@@ -168,7 +169,9 @@ CmmTable CmmPup(pup_er p, CmmTable t, CmmPupMessageFn msgpup)
...
+      tags = (int*) malloc((size_t)ntags*sizeof(int));
```
matthiasdiener, line 173: Can this be a std::vector? (same for threads.C)

src/util/charmrun-src/charmrun.C
```diff
@@ -4367,16 +4367,16 @@ static void start_nodes_daemon(std::vector<nodetab_process> & process_table)
   {
     const nodetab_host * h = p.host;
 
-    char *arg_currdir_r = pathfix(arg_currdir_a, h->pathfixes);
-    strcpy(task.cwd, arg_currdir_r);
-    free(arg_currdir_r);
-    char *arg_nodeprog_r = pathextfix(arg_nodeprog_a, h->pathfixes, h->ext);
-    strcpy(task.pgm, arg_nodeprog_r);
+    char *_arg_currdir_r = pathfix(arg_currdir_a, h->pathfixes);
```
rbuch, line 4370: This should ideally be something more descriptive.

src/conv-core/msgmgr.C
evan-charmworks: It looks like LGTM no longer identifies alerts at these locations, so I've removed the numeric limits checks.

src/conv-core/msgmgr.C
evan-charmworks: This CmmTable code does highly suspicious things like allocating these tags as subregions of a single CmiAlloc when you create them, but separate mallocs when unpacking. It should probably be on the chopping block entirely. I don't want to touch it now, and I've removed the limit check anyway.

In threads.C, the pointer to the buffer that is allocated is the return value of the function.

src/util/charmrun-src/charmrun.C
evan-charmworks: Done.

matthiasdiener (approved):

rbuch (approved):

Issue #1505, opened 2017-04-15
Deprecate hapi_pool* in favor of hapi_host*


mprobson:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1505*

---
This is pending Jaemin's patch "#2275":~~https://charm.cs.illinois.edu/gerrit/#/c/2275/~~ https://github.com/UIUC-PPL/charm/commit/1e117f16a5e82401a7c4940b5fb5cd195184cfca and that patch that introduces the malloc versions of these functions "#2407":~~https://charm.cs.illinois.edu/gerrit/#/c/2407/~~ https://github.com/UIUC-PPL/charm/commit/7401869eab4fd0d7192edc73d975011817bbc5dd.

This makes the API more extensible and allows the user to choose what memory mode they want to use when compiling. 

Places that need the change
- charm examples
- ChaNGa (already has equivalent change for the hapi_poolFree to hapi_hostFree)

minitu:
The function signature of `hapiPoolMalloc` does not match `cudaMallocHost`, which means the current macro definitions in `hapi.h` for `hapiHostMalloc` will not work. Need to fix `hapiPoolMalloc` to have the same signature as `cudaMallocHost`, and then ChaNGa (and other apps) can use `hapiHostMalloc` and control its behavior through compile-time flags such as `HAPI_USE_CUDAMALLOCHOST` and `HAPI_MEMPOOL`.

minitu:
Should be resolved by https://github.com/UIUC-PPL/charm/pull/3017 and https://github.com/N-BodyShop/changa/pull/79.

minitu:
Resolved.

PR #3017, opened 2020-08-13
Update hapiPoolMalloc to match signature of cudaMallocHost

minitu:
Related to #1505.
`hapiHostMalloc` defined in `hapi.h` does not work properly since it is defined as either `void* hapiPoolMalloc(size_t size)` or `cudaError_t cudaMallocHost(void** ptr, size_t size)` depending on the compile-time flags.
This PR updates `hapiPoolMalloc` to have the same signature as `cudaMallocHost` to resolve this issue.

### DIFF
--- a/examples/charm++/cuda/matmul/matmul.C
+++ b/examples/charm++/cuda/matmul/matmul.C
@@ -135,14 +135,14 @@ class Workers : public CBase_Workers {
     NVTXTracer nvtx_range("Workers::~Workers", NVTXColor::WetAsphalt);
 #endif
 
-    hapiFreeHost(h_A);
-    hapiFreeHost(h_B);
-    hapiFreeHost(h_C);
+    hapiCheck(cudaFreeHost(h_A));
+    hapiCheck(cudaFreeHost(h_B));
+    hapiCheck(cudaFreeHost(h_C));
     hapiCheck(cudaStreamDestroy(stream));
 #ifndef USE_WR
-    hapiFree(d_A);
-    hapiFree(d_B);
-    hapiFree(d_C);
+    hapiCheck(cudaFree(d_A));
+    hapiCheck(cudaFree(d_B));
+    hapiCheck(cudaFree(d_C));
 
     if (useCublas) cublasDestroy(handle);
 #endif
--- a/src/arch/cuda/hybridAPI/hapi.h
+++ b/src/arch/cuda/hybridAPI/hapi.h
@@ -5,15 +5,6 @@
 /* See hapi_functions.h for the majority of function declarations provided
  * by the Hybrid API. */
 
-// HAPI wrappers for pinned host memory allocation
-#ifdef HAPI_MEMPOOL
-#define hapiMallocHost hapiPoolMalloc
-#define hapiFreeHost   hapiPoolFree
-#else
-#define hapiMallocHost cudaMallocHost
-#define hapiFreeHost   cudaFreeHost
-#endif // HAPI_MEMPOOL
-
 #ifdef __cplusplus
 
 #include <cstring>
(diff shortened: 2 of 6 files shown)


minitu:
Also removed `HAPI_USE_CUDAMALLOCHOST` as `hapiHostMalloc` cannot be a wrapper for `malloc` due to its different signature, and ChaNGa uses this flag to call `malloc` on its own anyways (so no impact on ChaNGa).

src/arch/cuda/hybridAPI/hapi_functions.h
```diff
@@ -43,7 +43,7 @@ AMPI_CUSTOM_FUNC(cudaError_t, hapiFreeHostPool, void*)
 AMPI_CUSTOM_FUNC(cudaError_t, hapiMemcpyAsync, void*, const void*, size_t, enum cudaMemcpyKind, cudaStream_t)
 
 // Explicit memory allocations using pinned memory pool.
-AMPI_CUSTOM_FUNC(void*, hapiPoolMalloc, size_t)
+AMPI_CUSTOM_FUNC(void, hapiPoolMalloc, void**, size_t)
```
evan-charmworks, line 46: `cudaMallocHost` returns `cudaError_t`. Should `hapiPoolMalloc` do the same? Same question for `hapiPoolFree`.

minitu: Good point, I think they should for `hapiCheck` to work (which checks the `cudaError_t` return value).

minitu:
Also removed unused CUDA wrappers, especially `hapiMallocHost` and `hapiFreeHost` as they collided with the name changes.

src/arch/cuda/hybridAPI/hapi.h
```diff
@@ -5,22 +5,6 @@
...
-#    define hapiHostFree   hapiPoolFree
-#  else
-#    define hapiHostMalloc cudaMallocHost
-#    define hapiHostFree   cudaFreeHost
-#  endif // HAPI_MEMPOOL
-#else
-#  define hapiHostMalloc malloc
-#  define hapiHostFree   free
-#endif // HAPI_USE_CUDAMALLOCHOST
-
```
minitu, line 23: This was retained because of ChaNGa, but ChaNGa will be updated such that this is not required (https://github.com/N-BodyShop/changa/pull/79).

src/arch/cuda/hybridAPI/hapi_functions.h
```diff
@@ -38,13 +38,11 @@ AMPI_CUSTOM_FUNC(cudaError_t, hapiMalloc, void**, size_t)
 AMPI_CUSTOM_FUNC(cudaError_t, hapiFree, void*)
 AMPI_CUSTOM_FUNC(cudaError_t, hapiMallocHost, void**, size_t)
 AMPI_CUSTOM_FUNC(cudaError_t, hapiFreeHost, void*)
-AMPI_CUSTOM_FUNC(cudaError_t, hapiMallocHostPool, void**, size_t)
-AMPI_CUSTOM_FUNC(cudaError_t, hapiFreeHostPool, void*)
```
minitu, line 42: These wrappers to `hapiPool*` functions are no longer needed as `hapiPool*` functions also return `cudaError_t` and have the same function prototypes as the CUDA variants.

examples/charm++/cuda/matmul/matmul.C
```diff
@@ -135,14 +135,14 @@ class Workers : public CBase_Workers {
...
+    hapiCheck(cudaFreeHost(h_B));
+    hapiCheck(cudaFreeHost(h_C));
     hapiCheck(cudaStreamDestroy(stream));
 #ifndef USE_WR
-    hapiFree(d_A);
-    hapiFree(d_B);
-    hapiFree(d_C);
+    hapiCheck(cudaFree(d_A));
+    hapiCheck(cudaFree(d_B));
+    hapiCheck(cudaFree(d_C));
```
minitu, line 145: These were changed to match the allocations (which are done through `cudaMallocHost`)

minitu:
Ready for review, decided to go with `cudaError_t hapiMallocHost(void** ptr, size_t size, bool pool)` for the user to choose whether or not to use the memory pool for pinned host memory.

PR #2950, opened 2020-07-13
AMPI Benchmarks: Cleanup and add untested tests

nitbhat:
The major change in this PR is the addition of the following tests to the `test` target:
alltoall  isomalloc onesided speed

The minor change involves cleanup of the trailing white spaces and indentation from source code files. I also ran clang-format on the source code files.

PR #3018, opened 2020-08-14
Collision detection library and examples: Formatting/Stylistic cleanup 

nitbhat:
Ran two vim commands on all files:

1. gg=G (Fix indentation and remove tabs)
2. %s/\s\+$//e (To remove trailing whitespaces)

### DIFF
--- a/src/libs/ck-libs/collide/bbox.h
+++ b/src/libs/ck-libs/collide/bbox.h
@@ -1,10 +1,10 @@
 /*
-Orion's Standard Library
-written by 
-Orion Sky Lawlor, olawlor@acm.org, 7/18/2001
+   Orion's Standard Library
+   written by
+   Orion Sky Lawlor, olawlor@acm.org, 7/18/2001
 
-Rather complete interval and box classes.
-*/
+   Rather complete interval and box classes.
+   */
 #ifndef __OSL_BBOX_H
 #define __OSL_BBOX_H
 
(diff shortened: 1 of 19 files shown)


src/libs/ck-libs/collide/bbox.h
```diff
@@ -14,181 +14,181 @@ Rather complete interval and box classes.
...
-		segs[1].init(a[1],b[1],c[1]);
-		segs[2].init(a[2],b[2],c[2]);
-	}
-	
-	void print(const char *desc=NULL) const;
+  rSeg1d segs[3];//Spans for x (0), y (1), and z (2)
+  public:
+  bbox3d() {}
+  bbox3d(const rSeg1d &x,const rSeg1d &y,const rSeg1d &z)
+  {segs[0]=x; segs[1]=y; segs[2]=z;}
```
evan-charmworks, line 114: Did something go wrong with the whitespace cleanup here?

rbuch:
How does this compare to the results of `clang-format` (or anything else that has come up in Core).

src/libs/ck-libs/collide/bbox.h
nitbhat: It looks like `gg=G` indents `{` right under the function definition start and that's what was done here. It just happened to have a one-line body causing the weirdness. I fixed this in the next commit.

nitbhat:
> How does this compare to the results of `clang-format` (or anything else that has come up in Core)?

This was minor in terms of cleanup (and change) as compared to what `.clang-format` would've done. For eg, `.clang-format` works on a lot of minor spacing changes related to operators as seen in this file (https://github.com/UIUC-PPL/charm/pull/2950/files#diff-6d0bcf1168484d5075fef2b388a926ab). Another thing that `.clang-format` does IIRC is that it transforms all small function bodies to one line definition format. So far, we didn't come up with any standard formatting practices (Evan is going to be exploring some alternatives to clang-format that are more flexible).

Issue #3024, opened 2020-08-18
new build system fails with ChaNGa target on frontera with ROMIO


trquinn:
This may be similar to #2648, except when building the ChaNGa target there is no need to build ROMIO (or AMPI for that matter.)
The build command is:
`./build ChaNGa ucx-linux-x86_64 smp --with-production --basedir=$HOME/ucx/build_master -j16`
and the final errors are:
```[ 98%] No patch step for 'romio'
[100%] No update step for 'romio'
[100%] Performing configure step for 'romio'
configure: WARNING: Your make does not allow comments in target code.
Using this make may cause problems when building programs.
You should consider using gnumake instead.
configure: WARNING: File locks may not work with NFS.  See the Installation and
users manual for instructions on testing and if necessary fixing this
configure: error: Unable to compile a simple MPI program.
Use environment variables to provide the location of MPI libraries and
include directories
make[2]: *** [src/libs/ck-libs/ampi/romio-prefix/src/romio-stamp/romio-configure] Error 1
make[1]: *** [src/libs/ck-libs/ampi/CMakeFiles/romio.dir/all] Error 2
make: *** [all] Error 2
```
And to be clear: I don't want to build an MPI program!

matthiasdiener:
This error doesn't necessarily mean that there is anything wrong with the AMPI or romio compilation; most likely a library failed to build or can't be found at link time (romio configure is the first time any binary is linked using charmc).

Can you please try to run `make` in `tests/charm++/simplearrayhello/` and report any errors that might appear?

trquinn:
You nailed it.  The UCX libraries have an undefined reference.  That's probably causing the problem.

matthiasdiener:
> You nailed it. The UCX libraries have an undefined reference. That's probably causing the problem.

Ok, that's good to know. Do you know whether this undefined reference is due to a build error in ucx or in charm++?

trquinn:
Probably user error in building ucx.h

matthiasdiener:
Thanks. I've added a patch that will make build issues like this easier to find and debug (https://github.com/UIUC-PPL/charm/issues/3025).

PR #3025, opened 2020-08-18
build: make sure we can link executables

matthiasdiener:
Makes build errors such as https://github.com/UIUC-PPL/charm/pull/3024 a bit easier to find and debug.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -861,6 +861,12 @@ if(${TARGET} STREQUAL "charm4py")
   add_dependencies(charm hwloc)
 endif()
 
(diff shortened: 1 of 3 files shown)


matthiasdiener:
> Do you foresee any issue with hello.{decl,def}.h being generated in `include/`?

Not really, no. There is no other file that generates hello.{decl,def}.h currently, so nothing is overwritten. Also, building tests/charm++/simplearrayhello in its folder will rebuild hello.{decl,def}.h in the folder, so changes in the hello.ci file there are reflected when building tests/charm++/simplearrayhello/hello.

PR #3027, opened 2020-08-19
Fix userDriven exit in MPI builds

epmikida:


evan-charmworks:
Do other machine layers need this as well, like how GNI calls `PMI_Finalize();` before `exit()`, and OFI and UCX call `runtime_fini();`?

epmikida:
> Do other machine layers need this as well, like how GNI calls `PMI_Finalize();` before `exit()`, and OFI and UCX call `runtime_fini();`?

Yes that's a good point. I'll go through those and make changes, and see if I see any others that seem in need of change.

src/ck-core/mpi-interoperate.C
```diff
@@ -164,10 +164,10 @@ void CharmLibInit(MPI_Comm userComm, int argc, char **argv) {
 }
 
 // CharmLibExit is called for both forms of interop when the application is
-// done with Charm. In userDrivenMode, this does a full exit and kills the
-// application, just like CkExit(). In MPI interop, it just kills the Charm
-// instance, but allows the outside application and other Charm instances to
-// continue.
+// done with Charm. It does all necessary cleanup for Charm++ and return
```
rbuch, line 167:
```suggestion
// done with Charm. It does all necessary cleanup for Charm++ and returns
```

Issue #2060, opened 2019-03-20
Communication performance degrades after successive load balancing when migrating many objects


juanjgalvez:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/2060*

---
This was observed on Blue Waters, running on 128 physical nodes, with chares communicating in a stencil-like pattern with large messages sizes (80,000 bytes). Doing AtSync load balancing, with +LBSyncResume and a barrier after resumeFromSync, and measuring the time it takes for the program to complete a series of iterations between load balancing steps, the time increases after each lb step (see attached plot). As far as I can tell, the performance keeps degrading and I haven't seen it reach a ceiling yet (I have run up to 300 lb steps). Also see the attached test program to replicate the issue (I ran with `+balancer GreedyLB +LBPeriod 0.001 +LBObjOnly +LBSyncResume +LBCommOff`). Issue is observed in both non-SMP and SMP mode.

The issue appears to be in location management, since that is the only outstanding process after a load balancing step (in this test computation resumes only after all migrations have completed). Also, in a separate test I implemented an application-level mechanism to force the update of locations before resuming computation, and that restores the original performance. And finally, if I "reset" the location tables on every PE like so:

```c++
void CkLocMgr::resetLocationTableEntries()
{
  std::vector<CmiUInt8> ids;
  for (auto itr = id2pe.begin(); itr != id2pe.end(); itr++)
    ids.push_back(itr->first);

  for (auto &id : ids) {
    if (homePe(id) == CkMyPe()) continue;
    updateLocation(id, homePe(id));
  }
}
```

the original performance is also restored, and slowly degrades again afterwards unless I keep "resetting" the tables in this fashion.
The reasoning behind the above code snippet is that the best performance was obtained at the beginning, when location managers assumed that chares were in their home PE. But still not sure where the exact bug lies.

![loc-bug-charm++-non-smp](https://user-images.githubusercontent.com/1772435/63629804-2bce6600-c5da-11e9-9073-6173da9fe896.png)
![loc-bug-charm++-smp](https://user-images.githubusercontent.com/1772435/63629805-2c66fc80-c5da-11e9-94ec-3098c7810a0f.png)

[lbtest.tar](https://github.com/UIUC-PPL/charm/files/3536678/lbtest.tar.gz)

juanjgalvez:
*Original date: 2019-03-20 15:28:23*

---
[...]

PR #3031, opened 2020-08-20
CkLocMgr: Bound number of hops that messages can take

epmikida:
In cases with frequent load balancing, chains of stale location table entries
can cause messages with extremely high numbers of hops to reach their
destination. This change limits the impact of these chains by routing messages
through the home PE as soon as a potential chain (hops > 2) is detected.
Further work should also go towards clearing out location tables to remove
the existence of these chains entirely.

#2060

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2536,6 +2536,11 @@ int CkLocMgr::deliverMsg(CkArrayMessage *msg, CkArrayID mgr, CmiUInt8 id, const 
       }
 #endif
       msg->array_hops()++;
+      // If we are hopping more than twice, we've discovered a stale chain
+      // of cache entries. Just route through home instead.
+      if (msg->array_hops() > 2 && CkMyPe() != homePe(id)) {
+        destPE = homePe(id);
+      }
       CkArrayManagerDeliver(destPE,msg,opts);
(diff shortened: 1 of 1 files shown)


epmikida:
#2060 I don't think should be closed when this is merged, but I don't know if there's a way for github to link to it and not close it?

src/ck-core/cklocation.C
```diff
@@ -2536,6 +2536,11 @@ int CkLocMgr::deliverMsg(CkArrayMessage *msg, CkArrayID mgr, CmiUInt8 id, const
       }
 #endif
       msg->array_hops()++;
+      // If we are hopping more than twice, we've discovered a stale chain
+      // of cache entries. Just route through home instead.
+      if (msg->array_hops() > 2 && CkMyPe() != homePe(id)) {
+        destPE = homePe(id);
```
evan-charmworks, line 2542: Is there a lookup cost of calling `homePe()`? Is it worth making a local variable to only call it once?

epmikida: The cost is (or should be) very small but it depends on the mapping used. If it's made a local variable then it will always be called once, whereas in this case it is never called (due to short circuiting) unless we are detecting a long chain.

PR #3034, opened 2020-08-23
cmake: enable cuda build

matthiasdiener:
Note that this doesn't need special.sh.


matthiasdiener:
The mempool bug shown above occurs with the old build as well. I think this is ready for review.

PR #3038, opened 2020-08-27
Set CMK_SMP to 1 in conv-mach-opt.mak for smp mode and multicore builds

nitbhat:


stwhite91 (dismissed):

CMakeLists.txt
```diff
@@ -229,8 +229,14 @@ if(${OMP})
     find_package(OpenMP REQUIRED)
 endif()
 
+if(${SMP})
```
evan-charmworks, line 232: I think the preferred style of using CMake variables in if statements is to leave out the `${}`, in this case, `if(SMP)`. We've been doing it this way in new PRs.

nitbhat: Okay, fixed in the latest build.

nitbhat:
> Does the old build system need this change too?

No, the old build system had these changes. I think they were just not ported to the cmake build system.

PR #3051, opened 2020-09-09
LB: Forward declare JSON library to fix Kokkos compilation

### LINKED ISSUE #3042: JSON header file preventing charmc to use Kokkos' nvcc_wrapper as underlying compiler
To allow both Charm++ and Kokkos code to be in the same compilation unit, `charmc` should be able to use Kokkos' `nvcc_wrapper` as the underlying compiler. This is because the Kokkos headers contain code with `__device__` flags that need to be passed through `nvcc_wrapper` (or `nvcc -x cu`). While current Charm++ code can be compiled with just `nvcc` (which treats input files according to their file extension), it fails to compile with `nvcc -x cu` (which treats the input files as `.cu`) because the JSON header file in Charm++ (`src/util/json.hpp`) does not compile (see https://github.com/nlohmann/json/issues/1773 for more details). This issue has not been fixed as of JSON v.3.9.1. When the JSON header file is switched with an older version, compilation either fails due to some weird memory allocation issue or it does not have the APIs that Charm++ uses (e..g `contains()`).

rbuch:
Fixes #3042.

### DIFF
--- a/src/ck-core/CMakeLists.txt
+++ b/src/ck-core/CMakeLists.txt
@@ -66,7 +66,7 @@ set(ldb-h-sources ../ck-ldb/BaseLB.h ../ck-ldb/CentralLBMsg.h
     ../ck-ldb/ckheap.h ../ck-ldb/ckset.h
     ../ck-ldb/elements.h
     ../ck-ldb/lbdb.h ../ck-ldb/manager.h
-    ../ck-ldb/idm.h ../util/json.hpp)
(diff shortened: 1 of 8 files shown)


evan-charmworks:
LGTM but I'll let someone test before approving.

minitu (approved):
Confirmed that it fixes the compilation error.

evan-charmworks (approved):

PR #3053, opened 2020-09-10
Doc: Check for program existence using POSIX standard "command" instead of "which"

rbuch:
`which -s` is broken, my local version of `which` has no support for the `-s` flag (and I can't find what it's supposed to mean, either). Using the POSIX standard `command` command fixes this for me and will be more compatible in general.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -29,7 +29,7 @@ clean:
 	rm -rf pdf html
 
 check-sphinx:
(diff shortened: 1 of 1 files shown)


evan-charmworks:
Most uses of `command -v` in the repo also pass `>/dev/null 2>&1`. Would that be useful here? Otherwise, LGTM.

matthiasdiener (approved):

evan-charmworks (approved):

PR #3060, opened 2020-09-16
Check for overflow when creating new futures 

ZwFink:


src/ck-core/ckfutures.C
```diff
@@ -136,6 +137,8 @@ int createFuture(void)
   /* if the freelist is empty, allocate more futures. */
   if (fs->freelist == -1) {
     origsize = fs->max;
+    // origsize may overflow CMK_REFNUM_TYPE, creating problems when waiting on this future
+    CkAssert( ( origsize <= std::numeric_limits<CMK_REFNUM_TYPE>::max() - origsize ) );
```
evan-charmworks, line 141: Could you explain the math on this line?

ZwFink: If origsize > ( MAX_REFNUM - origsize ) then (origsize + origsize) > MAX_REFNUM. 
freelist is reallocated with 2 * origsize capacity, and new futures can be created that have an id greater than MAX_REFNUM.

ZwFink: Though it's probably better to move the overflow check to when the ID is assigned to the new future. This version limits the number of futures to MAX_REFNUM / 2 when it can be simply MAX_REFNUM.

rbuch: Yeah, this check should probably be moved to right above line 147 and just be `CkAssert(fs->freelist <= std::numeric_limits<CMK_REFNUM_TYPE>::max())`.

rbuch (approved):

evan-charmworks (approved):

Issue #3061, opened 2020-09-17
NAMD's HybridLB crashes during load balancing


rbuch:
When running NAMD with `ldBalancer hybrid` with recent Charm (8d5480d3b99c886d2a151f3c194c2fb1443355a8), there is a crash during load balancing. Works fine in v6.10.2, so it's a regression in Charm++.

```
------------- Processor 3 Exiting: Called CmiAbort ------------
Reason: [3] Assertion "n<len" failed in file /home/ronak/Code/charm/netlrts-linux-x86_64-smp-clang-10/bin/../include/cklists.h line 225.

[3] Stack Traceback:
  [3:0] namd2 0xf3c0e6 
  [3:1] namd2 0xf4b82a 
  [3:2] namd2 0xe1ebf6 HybridBaseLB::PropagateInfo(Location*, int, int)
  [3:3] namd2 0xe294c7 CkIndex_HybridBaseLB::_call_PropagateInfo_marshall15(void*, void*)
  [3:4] namd2 0xe8e353 CkDeliverMessageFree
  [3:5] namd2 0xe93f24 
  [3:6] namd2 0xe915db _processHandler(void*, CkCoreState*)
  [3:7] namd2 0xf48a7c CsdScheduleForever
  [3:8] namd2 0xf487c2 CsdScheduler
  [3:9] namd2 0xf40794 
  [3:10] namd2 0xf43f57 
  [3:11] libpthread.so.0 0x7ffff75f56db 
  [3:12] libc.so.6 0x7ffff660aa3f clone
```

rbuch:
This is happening because of 921a637b97cc89df7c8a5c6dfeb00866290ef85a, presumably it's not copying a message that should be copied now.

rbuch:
#3063 should fix this bug as far as I can tell, but testing on Frontera with that patch still crashes in `PropagateInfo` for some reason. I'm digging to see what's going on.

PR #3069, opened 2020-10-01
Bug #3061: Fix converse header rank race condition in list send

rbuch:


stwhite91 (approved):

src/conv-core/convcore.C
```diff
@@ -3031,12 +3031,32 @@ void CmiGroupInit(void)
...
+  // for local PEs and there is a race between setting the rank in the Converse header and
+  // the actual comm thread send for remote PEs
+#if CMK_SMP
+  for (int i = 0; i < npes - 1; i++)
+  {
+    const char* msgCopy = CmiCopyMsg(msg, len);
+    CmiSyncSendAndFree(pes[i], len, msgCopy);
   }
+
+  if (npes > 0) CmiSyncSend(pes[npes - 1], len, msg);
```
matthiasdiener, line 3046:
```suggestion
  if (npes > 0) 
    CmiSyncSend(pes[npes - 1], len, msg);
```

tests/charm++/megatest/listsend.C
```diff
@@ -0,0 +1,48 @@
+#include "listsend.h"
+#include <numeric>
+
+#define VALUE 0x01ABCDEF
+
+void listsend_moduleinit(void) {}
```
matthiasdiener, line 6: no need for `(void)` as the argument for these functions.

src/conv-core/convcore.C
matthiasdiener: Also, can you add a comment on why `npes-1` is treated separately?

matthiasdiener (approved):

src/conv-core/convcore.C
```diff
@@ -3031,12 +3031,33 @@ void CmiGroupInit(void)
...
+    CmiSyncSendAndFree(pes[i], len, msgCopy);
   }
+  // No need to copy and free for the last send, so use the original message
+  if (npes > 0)
+    CmiSyncSend(pes[npes - 1], len, msg);
+#else
+  for (int i = 0; i < npes; i++)
+  {
+    // Copy if this is a self send to avoid unpack/send race
+    if (pes[i] == CmiMyPe() && npes > 1)
```
epmikida, line 3052: Why would this cause a race? Wouldn't the unpack happen after I've finished doing all of the sends?

epmikida (approved):
LGTM pending the answer to one question

src/conv-core/convcore.C
rbuch: Not necessarily, at least on some layers, even for non-SMP mode, the messaging layer still works by adding messages to a queue which it then processes between running EPs, but it checks its local queue before sending messages for external PEs (`src/arch/util/machine-common-core:1812-1816`):
```
    msg = CMIQueuePop(cs->recv); // checks local queue for msg
#endif
#if (!CMK_SMP || CMK_SMP_NO_COMMTHD) && !CMK_MULTICORE
    if (!msg) {
       AdvanceCommunication(0); // sends queued messages to remote PEs
```

When this happens, it is unpacked locally and then the same unpacked message is used for sends to remote PEs.

epmikida: Gotcha

PR #3076, opened 2020-10-07
Tracing: Add timestamp and runline to Projections logs

rbuch:


matthiasdiener:
Why not call it commandline (or cmdline)?

rbuch:
> Why not call it commandline (or cmdline)?
> 
> Also, does this need to increase the version number of the trace?

I thought that this name might be clearer, but I think my fears were probably overblown. I'll change it to `COMMANDLINE`.

I'd say no, these files will work with older versions of Projections with no issue (they just won't display the new information), and none of the existing schema changes, it just adds these two new fields.

matthiasdiener (approved):

matthiasdiener:
In a future PR, we could also add the Charm++ version.

stwhite91 (approved):

PR #3085, opened 2020-10-10
AMPI: Use ROMIO's MPI_File_{get,set}_errhandler instead of stubs

evan-charmworks:


evan-charmworks:
Ready for review

stwhite91 (approved):

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -10002,6 +10002,8 @@ AMPI_API_IMPL(int, MPI_File_create_errhandler, MPI_File_errhandler_function *fun
   return MPI_SUCCESS;
 }
 
+#if !CMK_AMPI_WITH_ROMIO
+// Disable ROMIO's get_errh.c and set_errh.c when implementing these.
 AMPI_API_IMPL(int, MPI_File_set_errhandler, MPI_File file, MPI_Errhandler errhandler)
```
matthiasdiener, line 10007: Why are we implementing these functions at all?

evan-charmworks: It seems like `errhandler` code is intended to be shared between MPI_File, MPI_Comm, and MPI_Win. ROMIO has this:

```cpp
/* Section 4.13.3 */
#ifndef MPICH
/* MPICH provides these definitions */
AMPI_CUSTOM_FUNC(int, MPI_File_set_errhandler, MPI_File file, MPI_Errhandler errhandler)
AMPI_CUSTOM_FUNC(int, MPI_File_get_errhandler, MPI_File file, MPI_Errhandler *errhandler)
#endif
```

No other `*_errhandler` functions are present in ROMIO.

matthiasdiener: Right, but is having these no-op functions in the AMPI/no ROMIO case useful at all?

matthiasdiener: Ok, let's leave them there for now.

matthiasdiener (approved):

PR #3083, opened 2020-10-09
AMPI: Remove NOIMPL declarations of MPI_File symbols provided by ROMIO

evan-charmworks:
Provided by:
- mpi-io/iread_all.c
- mpi-io/iread_atall.c
- mpi-io/iwrite_all.c
- mpi-io/iwrite_atall.c
- mpi-io/register_datarep.c

PR #3086, opened 2020-10-10
AMPI: Ensure function pointer shim objects are built after ROMIO

evan-charmworks:
Depends on #3083
Depends on #3085


stwhite91 (approved):

matthiasdiener (approved):
Out of curiosity, why do these need to be built after ROMIO?

evan-charmworks:
> Out of curiosity, why do these need to be built after ROMIO?

They need to see ROMIO's true headers and not the empty stub versions, or else they will lack ROMIO's symbols.

PR #3101, opened 2020-10-14
Persistent Intra-node GPU Messaging

minitu:
- Adds intra-node persistent GPU messaging with examples, OSU latency benchmark and documentation
- Changes in HAPI to support stack-allocated `CkCallback` and fixes to enable compilation with AMPI

**TODOs**
- ~~Intra-process put~~
- ~~Inter-process get & put~~
- ~~Test load balancing~~
- ~~Evaluate performance~~
- ~~Update documentation~~


minitu:
@stwhite91 @evan-charmworks Do you have an idea as how to fix the AMPI compilation error? I wanted to change `hapiAddCallback`'s API to take `CkCallback` objects instead of pointers, and it looks like the inclusion of `ckcallback.h` in `hapi.h` is breaking AMPI.

```
In file included from /home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/cklists.h:6:0,
                 from /home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/register.h:9,
                 from /home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/ckcallback.h:19,
                 from /home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/hapi.h:4,
                 from /home/runner/work/charm/charm/src/libs/ck-libs/ampi/ampi.h:12,
                 from /home/runner/work/charm/charm/src/libs/ck-libs/ampi/ampif.C:1:
/home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/pup_stl.h: In function ‘void PUP::operator|(PUP::er&, T&)’:
/home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/pup_stl.h:427:5: error: there are no arguments to ‘pup_bytes’ that depend on a template parameter, so a declaration of ‘pup_bytes’ must be available [-fpermissive]
     pup_bytes(&p, static_cast<void*>(&s), sizeof(T));
     ^~~~~~~~~
/home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include/pup_stl.h:427:5: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
[ 67%] Building CXX object src/util/CMakeFiles/conv-util.dir/pup_util.C.o
[ 67%] Building CXX object src/libs/ck-libs/CMakeFiles/modulecollide.dir/collide/collidecharm.C.o
Fatal Error by charmc in directory /home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/src/libs/ck-libs/ampi
   Command g++ -DCMK_GFORTRAN -DCMK_GFORTRAN -I/home/runner/work/charm/charm/netlrts-linux-x86_64-cuda/bin/../include -I/usr/include -D__CHARMC__=1 -DAMPI_ERRHANDLER_RETURN=0 -g -g -std=gnu++11 -fno-stack-protector -fno-lifetime-dse -std=gnu++11 -c /home/runner/work/charm/charm/src/libs/ck-libs/ampi/ampif.C -o CMakeFiles/ampif.dir/ampif.C.o returned error code 1
charmc exiting...
```

evan-charmworks:
> @stwhite91 @evan-charmworks Do you have an idea as how to fix the AMPI compilation error?

This got me most of the way:

```diff
$ git diff
diff --git a/src/libs/ck-libs/ampi/ampi.h b/src/libs/ck-libs/ampi/ampi.h
index 168149eaf..73c412aff 100644
--- a/src/libs/ck-libs/ampi/ampi.h
+++ b/src/libs/ck-libs/ampi/ampi.h
@@ -9,7 +9,9 @@
 #define AMPI_INTERNAL_SKIP_FUNCTIONS

 #if CMK_CUDA
-#include "hapi.h"
+#include <cuda_runtime.h>
+/* forward declare this type because it contains Charm features such as CkCallback */
+typedef struct hapiWorkRequest hapiWorkRequest;
 #endif

 /* NON-standard define: this lets people #ifdef on
```

However, it still fails because hapiAddCallback now takes a CkCallback. Any AMPI_CUSTOM_FUNC API should not use Charm types. You would need to move the declaration of hapiAddCallback out of hapi_functions.h and into hapi.h as a normal declaration instead of using the AMPI macro. For AMPI, it might be nice to add a new hapi symbol in its place that continues to take a function pointer and generates a CkCallback as part of its implementation.

stwhite91:
> For AMPI, it might be nice to add a new hapi symbol in its place that continues to take a function pointer and generates a CkCallback as part of its implementation.

Yeah I think this is the best workaround.

evan-charmworks:
After looking into the code a bit I am now in doubt of whether the hapi<-->AMPI interface was ever really tested. I didn't know this was how these `void *` callback parameters are interpreted:

```c++
#if CMK_CUDA
void CUDACallbackManager(void *fn, void *msg) {
  if (fn) {
    ((CkCallback*)fn)->send(msg);
  }
}
#endif
```

They're not function pointers but CkCallback pointers, and they were simply made `void *` to avoid the compilation issues we're seeing now despite it not being semantically valid to use CkCallback from AMPI.

In the future we should look into some kind of CI to test building pure C programs that include hapi's headers. Remember that C does not have function overloading, references, or default parameters. The class definition of CkCallback is C++-only as well. Anything in headers that uses these should be inside `#ifdef __cplusplus` guards.

Since I don't know if the AMPI interface ever worked properly we don't need to block this PR on it.

minitu:
@evan-charmworks

Is it true that `CkCallback` shouldn't be used inside AMPI? I see some of them being used in `ampi.C`.

Although the CI checks all passed, I'm seeing a build error when I try `./buildold all-test pamilrts-linux-ppc64le smp cuda -j -g --with-production` on Summit:

```
checking whether we are cross compiling... configure: error: in `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs/ampi/romio':
configure: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
gmake[3]: *** [/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/lib/libampiromio.a] Error 1
gmake[3]: Leaving directory `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs/ampi'
gmake[2]: *** [ampi] Error 2
gmake[2]: Leaving directory `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs'
gmake[1]: *** [charmlibs] Error 2
gmake[1]: Leaving directory `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs'
gmake: *** [CHARMLIBS] Error 2
```

Maybe `hapi.h` still doesn't properly separate C++ and C codes.

evan-charmworks:
> Is it true that `CkCallback` shouldn't be used inside AMPI? I see some of them being used in `ampi.C`.

It's fine for them to be used inside the implementation of AMPI. What I mean is that AMPI programs wouldn't have access to CkCallback unless they also link with Charm.

> Although the CI checks all passed, I'm seeing a build error when I try `./buildold all-test pamilrts-linux-ppc64le smp cuda -j -g --with-production` on Summit:
> 
> ```
> checking whether we are cross compiling... configure: error: in `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs/ampi/romio':
> configure: error: cannot run C compiled programs.
> If you meant to cross compile, use `--host'.
> See `config.log' for more details
> ```

Look at `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs/ampi/romio/config.log` to see what the issue is.

minitu:
Right. The build error seems to have occurred because configure tried to execute a binary that would only work with `jsrun` on a compute node of Summit:

```
configure:3950: ./conftest
./conftest: error while loading shared libraries: libcollectives.so.3: cannot open shared object file: No such file or directory
configure:3954: $? = 127
configure:3961: error: in `/gpfs/alpine/csc357/scratch/jchoi/charm/pamilrts-linux-ppc64le-cuda-smp/tmp/libs/ck-libs/ampi/romio':
configure:3963: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
```

evan-charmworks:
> Right. The build error seems to have occurred because configure tried to execute a binary that would only work with `jsrun` on a compute node of Summit:

This might be expected, it seems familiar to me. Try rebuilding AMPI with `--without-romio` passed to the build script.

minitu:
That fixed it, thanks!

minitu:
Ready for review. It may be easier to start with the documentation and look at the examples to see how the API works.

src/libs/ck-libs/ampi/ampi.C
```diff
@@ -11761,8 +11761,8 @@ int AMPI_GPU_Iinvoke(cudaStream_t stream, MPI_Request *request)
 
   // A callback that completes the corresponding request
   CkCallback *cb = new CkCallback(&AMPI_GPU_complete, newreq);
```
stwhite91, line 11763: Can this just be created on the stack instead of heap allocated?

minitu: I think the problem is `hapiWorkRequestSetCallback` and `hapiAddCallback` used in AMPI are declared as `AMPI_CUSTOM_FUNC` in `hapi_functions.h`, and Evan mentioned that these functions should not take `CkCallback` as a parameter. That's why these are taking a `void *` instead and require `CkCallback` to be allocated on the heap.

evan-charmworks: You don't have to allocate structures on the heap to pass them as `void *`. You can create `cb` as a normal local variable and pass `&cb` into the function call.

minitu: @evan-charmworks Good point, I didn't realize I could do that because there is a delete call afterwards anyways.

minitu: @evan-charmworks @stwhite91 Just pushed a commit to fix this issue

doc/charm++/manual.rst
```diff
@@ -9151,9 +9151,72 @@ Inter-node messages will be transferred using the naive host-staged mechanism
 where the data is first transferred to the host from the source GPU, sent over
 the network, then transferred to the destination GPU.
 
+An optimized mechanism for inter-process communication using CUDA IPC, POSIX shared memory
+and pre-allocated GPU communication buffers are availble through runtime flags.
+This significantly reduces the overheads from creation and opening of CUDA IPC handles
+especially for relatively smaller messages. ``+gpushm`` will turn on this optimization
```
evan-charmworks, line 9157:
```suggestion
An optimized mechanism for inter-process communication using CUDA IPC, POSIX shared memory,
and pre-allocated GPU communication buffers are available through runtime flags.
This significantly reduces the overhead from creation and opening of CUDA IPC handles,
especially for relatively small messages. ``+gpushm`` will turn on this optimization
```

evan-charmworks (approved):

stwhite91 (approved):

PR #3110, opened 2020-10-17
cmake: passthrough $CC $CXX $FC for compiler selection

matthiasdiener:
With this patch, the user can pass $CC $CXX $FC to build (similar to how compilers are selected in cmake proper):

`$ CC=clang CXX=clang++ ./build charm++ ...`


minitu (dismissed):

rbuch (approved):

evan-charmworks:
This seems maybe risky to merge right before a release. Has this been tested in production environments?

matthiasdiener:
> This seems maybe risky to merge right before a release. Has this been tested in production environments?

Imho, it's not very risky. Few places will have this set I believe. Would printing `Note: using compiler specified in $CC` help?

evan-charmworks (approved):

minitu (approved):

stwhite91 (approved):

PR #3112, opened 2020-10-19
build: clean up the VERSION file

### LINKED ISSUE #2844: Conflict between LLVM's version file and our VERSION file

matthiasdiener:
create two separate files, `charm-version.h` and `charm-version-git.h` in
`include/`, with the release/API version and git version respectively.

The files look like this:

```
$ cat ./netlrts-darwin-x86_64/include/charm-version.h
#define CHARM_VERSION 61002
#define CHARM_VERSION_MAJOR 6
#define CHARM_VERSION_MINOR 10
#define CHARM_VERSION_PATCH 2
$ cat ./netlrts-darwin-x86_64/include/charm-version-git.h
#define CHARM_VERSION_GIT "v6.11.0-devel-430-g50bd748ad*"
```

Fixes #2844.

For the old build system, `commitid.sh` creates the `charm-version-git.h` file containing the git version of Charm++ (this version was previously contained in `tmp/VERSION`.


CMakeLists.txt
```diff
@@ -1095,10 +1095,11 @@ add_custom_target(create_symlinks ALL
 #   add_dependencies(create_symlinks createlink)
 # endif()
 
-# Create the .vdir/.gdir/VERSION files
+# Create the .vdir/.gdir/charm-version.h/charm-api-version.h files
 file(WRITE ${CMAKE_BINARY_DIR}/include/.gdir   "${CMK_GDIR}\n")
 file(WRITE ${CMAKE_BINARY_DIR}/include/.vdir   "${CMK_VDIR}\n")
-file(WRITE ${CMAKE_BINARY_DIR}/include/VERSION "${CHARM_VERSION}\n")
+file(WRITE ${CMAKE_BINARY_DIR}/include/charm-version.h "#define CHARM_VERSION "${CHARM_VERSION}"\n")
+file(WRITE ${CMAKE_BINARY_DIR}/include/charm-api-version.h "#define CHARM_API_VERSION ${PROJECT_VERSION}\n")
```
nilsleiffischer, line 1102: I suggest to `#define CHARM_VERSION_{MAJOR,MINOR,PATCH}` as well

matthiasdiener:
This is somewhat risky and the changes are not very well tested, but it fixes an important issue users have been seeing. @rbuch: Could you please check if the changes related to the tarball packaging work as expected?

CMakeLists.txt
```diff
@@ -1095,10 +1095,11 @@ add_custom_target(create_symlinks ALL
 #   add_dependencies(create_symlinks createlink)
 # endif()
 
-# Create the .vdir/.gdir/VERSION files
+# Create the .vdir/.gdir/charm-version.h/charm-api-version.h files
 file(WRITE ${CMAKE_BINARY_DIR}/include/.gdir   "${CMK_GDIR}\n")
 file(WRITE ${CMAKE_BINARY_DIR}/include/.vdir   "${CMK_VDIR}\n")
-file(WRITE ${CMAKE_BINARY_DIR}/include/VERSION "${CHARM_VERSION}\n")
+file(WRITE ${CMAKE_BINARY_DIR}/include/charm-version.h "#define CHARM_VERSION "${CHARM_VERSION}"\n")
```
nilsleiffischer, line 1101: The `"` need to be escaped, and the line below also needs the quotation marks

matthiasdiener: Fixed, thanks!

src/scripts/charmc
```diff
@@ -389,7 +389,7 @@ filterLegacyLB()
 printVersion()
 {
 	findCharmBin $0
-	version=`cat $CHARMBIN/../include/VERSION`
+	version=`cat $CHARMBIN/../include/charm-version.h`
```
rbuch, line 392: This prints out the entire define line.

package-tarball.sh
```diff
@@ -65,7 +65,7 @@ git add -u
 object_id=$(git write-tree)
 
 # Construct the target file/folder name
-version="charm-$(cat src/scripts/VERSION)"
+version="charm-$(cat src/scripts/charm-version.h)"
```
rbuch, line 68: This prints out the entire define line.

rbuch (commented):
Using `$(cat charm-version.h | grep CHARM_VERSION | awk '{print $3}')` in `package-tarball.sh` doesn't strip the version of the quotation marks, is there some other command that should be used for that? I'm confused why its use in `src/scripts/Makefile` doesn't result in extra quotation marks.

src/scripts/charmc
matthiasdiener: Fixed, thanks. With the most recent commit in this PR, this prints the API version, not the "git" version (as previously). f you want, we can change that.

package-tarball.sh
matthiasdiener: fixed, thanks.

matthiasdiener:
> Using `$(cat charm-version.h | grep CHARM_VERSION | awk '{print $3}')` in `package-tarball.sh` doesn't strip the version of the quotation marks, is there some other command that should be used for that? I'm confused why its use in `src/scripts/Makefile` doesn't result in extra quotation marks.

Good catch. Both `package-tarball.sh` and `src/scripts/Makefile` were incorrect and should be fixed now.

evan-charmworks:
Should this PR add mention of `VERSION`'s removal to `CHANGES`? It might be good to offer documentation on the forward transition path for any Charm users who parse it.

matthiasdiener:
> Should this PR add mention of `VERSION`'s removal to `CHANGES`? It might be good to offer documentation on the forward transition path for any Charm users who parse it.

We could add this and also add documentation to the manual regarding `CHARM_API_VERSION` etc.
Another idea is to `#include` both files into `charm-api.h` or so.

src/scripts/charmc
rbuch: That sounds fine to me, so long as the preamble when running a program still prints out the "git" version.

matthiasdiener:
Turns out that CHARM_VERSION is already used in conv-autoconfig.h (value is the same as CHARM_API_VERSION).
So I'll rename CHARM_API_VERSION_MAJOR etc. to CHARM_VERSION_MAJOR, and CHARM_VERSION to CHARM_VERSION_GIT.

rbuch:
One thing to note is some applications (at least Enzo-E) read the root VERSION file as part of their build process. I don't think that should stop us from making this change, just know that it will be a breaking one.

src/scripts/configure.ac
```diff
@@ -42,15 +42,26 @@ CHARMINC="."
...
+echo "#define CHARM_VERSION_PATCH $CHARM_VERSION_PATCH" >> ../include/charm-version.h
 
 charmout="charmconfig.out"
 /bin/rm -rf $charmout
 MAKEFILE_EXT=Make.extlib
 
 Echo() {
-	echo $* 
+	echo $*
 	echo $* >> $charmout
```
matthiasdiener, line 65:
```suggestion
	echo $* 
	echo $* >> $charmout
```

src/ck-core/charm-api.h
```diff
@@ -39,10 +42,10 @@ libraries written in Charm for other languages.
 #endif
 
 
-/** 
+/**
```
matthiasdiener, line 45:
```suggestion
/** 
```

src/ck-core/charm-api.h
```diff
@@ -39,10 +42,10 @@ libraries written in Charm for other languages.
...
  * Define a new Fortran-callable API routine, returning void,
- * that does nothing but map its arguments onto some 
- * (presumably analogous) C routine. 
+ * that does nothing but map its arguments onto some
+ * (presumably analogous) C routine.
```
matthiasdiener, line 48:
```suggestion
 * that does nothing but map its arguments onto some 
 * (presumably analogous) C routine. 
```

src/ck-core/charm-api.h
```diff
@@ -56,7 +59,7 @@ FTN_NAME(CAPITALNAME,lowername) routine_args { \
 	Cname c_args;\
 }
 
-/** 
+/**
```
matthiasdiener, line 62:
```suggestion
/** 
```

matthiasdiener:
Ready for review.

evan-charmworks:
Does this work correctly from a packaged tarball too? I get this from CMake with current master from a tarball:

```
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
==============================
-- Configuring done
Charm++ 6.10.2 configuration: 
  OS version:      Ubuntu 18.04.5 LTS (Linux 5.4.0-1031-azure, x86_64)
  Charm++ version:  (dev)
```

I don't know if passing `--release` to package-tarball.sh changes anything.

matthiasdiener:
> Does this work correctly from a packaged tarball too? I get this from CMake with current master from a tarball:
> 
> ```
> fatal: not a git repository (or any of the parent directories): .git
> fatal: not a git repository (or any of the parent directories): .git
> ==============================
> -- Configuring done
> Charm++ 6.10.2 configuration: 
>   OS version:      Ubuntu 18.04.5 LTS (Linux 5.4.0-1031-azure, x86_64)
>   Charm++ version:  (dev)
> ```
> 
> I don't know if passing `--release` to package-tarball.sh changes anything.

Good point, should be fixed now.

evan-charmworks:
buildold says this:

```
fatal: not a git repository (or any of the parent directories): .git
Copying charm-version-git.h.new = "" over charm-version-git.h = 
./system_ln $PWD/charm-version-git.h ../include/charm-version-git.h
```

matthiasdiener:
> buildold says this:
> 
> ```
> fatal: not a git repository (or any of the parent directories): .git
> Copying charm-version-git.h.new = "" over charm-version-git.h = 
> ./system_ln $PWD/charm-version-git.h ../include/charm-version-git.h
> ```

Ok, but this is not a regression with this PR I think.

evan-charmworks (approved):

buildold
```diff
@@ -761,8 +761,8 @@ EOF
     fi
     $DESTINATION/tmp/system_ln $DESTINATION/$newdir $newdir
   done
-  rm -f VERSION
-  $DESTINATION/tmp/system_ln $DESTINATION/include/VERSION VERSION
+  rm -f charm-version.h
+  $DESTINATION/tmp/system_ln $DESTINATION/include/charm-version.h charm-version.h
```
minitu, line 765: Do we also want to remove `charm-version-git.h` here?

matthiasdiener: Not sure, but I think the `rm` refers to the symlink only, and there is no symlink created for `charm-version-git.h`.

minitu (approved):

PR #3114, opened 2020-10-20
Updates Xlat-I AST Building Interface to use the new Entry Attribute Format

Wingpad:
This fixes a build issue present when using Charj on newer versions of Charm++.


evan-charmworks (approved):

matthiasdiener (approved):

src/xlat-i/interfaceBuilder.h
```diff
@@ -347,12 +347,12 @@ namespace Builder {
       : ret(ret_)
       , name(name_)
       , stackSize(stackSize_)
-      , eattrib(0)
+      , attribs(NULL)
```
rbuch, line 350: Minor, but should prefer `nullptr` here and below.

Wingpad: I chose `NULL` vs. `nullptr` to maintain internal consistency with the rest of the file's contents, since it uses `NULL` in other places. I could change this one use to `nullptr`, or refactor them all in one go in another PR. Any thoughts on which would be the best way to go?

evan-charmworks: IMHO we should use nullptr in new code, particularly where it is important in constructors and function calls. Various files in the codebase are already internally inconsistent w.r.t. NULL/nullptr.

minitu (dismissed):

rbuch (approved):

src/xlat-i/interfaceBuilder.h
```diff
@@ -347,12 +347,12 @@ namespace Builder {
...
       , name(name_)
       , stackSize(stackSize_)
-      , eattrib(0)
+      , attribs(nullptr)
       , sdag(sdag_)
     { }
 
     void addAttribute(BUILDER_ENTRY_ATTRIBUTES attribute) {
-      eattrib |= attribute;
+      attribs = new xi::Attribute(attribute, nullptr, attribs);
```
rbuch, line 355: I assume this gets deleted somewhere?

Wingpad: Now that you bring my attention to it, this is definitely getting leaked. That being said, deletion would only be necessary at the end of the program's execution and, also because this API is applied in a rather limited capacity, this would probably not be a big deal. I am honestly unsure how memory clean-up is works in xlat-i and, by extension, its API builder. I will have to look into this more and see if `Entry` (the attribute's "owner") is even getting deleted.

evan-charmworks (approved):

Issue #1915, opened 2018-05-16
Make the build script clean/remove old builds


stwhite91:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1915*

---
If a user has a netlrts-linux-x86_64/ and a netlrts-linux-x86_64-smp/, then they do a "./build charm++ netlrts-linux-x86_64", our build script should remove the existing netlrts-linux-x86_64 directory before doing anything else.

We could potentially also add a "./build clean" command to remove all builds, and a "./build rebuild charm++" to do "cd <build_dir>/tmp && make charm++"

epmikida:
*Original date: 2018-05-16 20:31:12*

---
I'm not sure if removing old builds should necessarily be the default, but either way something sensible should be done. Right now it appears to be broken completely. I did two ./builds in a row and the second one broke (after recompiling a bunch of unchanged files):

make: *** No rule to make target `charmrun', needed by `charmrun-target'.  Stop.
-------------------------------------------------
Charm++ NOT BUILT. Either cd into netlrts-darwin-x86_64/tmp and try

I would think making it work like a typical make file would be the most natural. As in, doing a ./build just defaults to running make in the appropriate directory if it exists, which will obey all dependency rules, only compile what needs to be compiled, and not rerun all the configuration checks. And then ./build clean <arch> maybe to clean that dir, and ./build clean all or ./build clean would remove all old builds?

stwhite91:
*Original date: 2018-05-16 21:08:54*

---
I agree with what you've said: our './build' should behave like 'make'.

evan-charmworks:
*Original date: 2018-05-22 19:42:20*

---
~~https://charm.cs.illinois.edu/gerrit/4215~~ https://github.com/UIUC-PPL/charm/commit/8d1ddb135c56a814bb6ea79c275e6de124b9444d
~~https://charm.cs.illinois.edu/gerrit/4216~~ https://github.com/UIUC-PPL/charm/commit/3466cb4d22a615d6116e5ec92652cd33b94ad252
~~https://charm.cs.illinois.edu/gerrit/4217~~ https://github.com/UIUC-PPL/charm/commit/93cfbf74ff3772cc5fd0d9046026400a62856cce

stwhite91:
*Original date: 2018-05-25 19:00:36*

---
All 3 of those patches have been merged, so rebuilding an already built target should work smoothly now.. We could still add ./build clean <arch> and ./build clean all

PR #3125, opened 2020-10-23
build: Detect repeated invocations and warn users

evan-charmworks:
Fixes a bug where partial rebuilds in tmp/ lacked -build-shared.
Fixes a bug where re-invoking ./buildold with AMPI-only failed.

### DIFF
--- a/buildcmake
+++ b/buildcmake
@@ -421,7 +421,17 @@ then
   if [[ $FORCE -eq 0 ]]
   then
     echo "To attempt a partial rebuild, run 'make' in \"$builddir\"."
-    echo "To remove the existing data and run configure again, run this script with --force."
+    echo "To remove the existing data and reconfigure, run this script again with --force."
+    exit 1
+  else
+    echo "Removing existing data and starting over. (--force)"
+  fi
+elif [[ -e "$builddir" ]]
+then
+  echo "\"$builddir\" already exists but is in an inconsistent state."
+  if [[ $FORCE -eq 0 ]]
+  then
+    echo "To remove the existing data and reconfigure, run this script again with --force."
     exit 1
   else
     echo "Removing existing data and starting over. (--force)"
--- a/buildcmake
+++ b/buildcmake
@@ -129,6 +129,7 @@ opt_compiler=""
 # Misc options
 VERBOSE=0
 ONLY_CONFIGURE=0
+FORCE=0
 
 processArgs() {
 
@@ -340,6 +341,10 @@ while [[ $# -gt 0 ]]; do
       # Run only cmake, not make
       ONLY_CONFIGURE=1
       ;;
+    --force)
+      # Overwrite existing configuration
+      FORCE=1
+      ;;
     -j)
       opt_parallel="$arg"
       if [[ $# -gt 0 && $1 == [0-9]* ]]; then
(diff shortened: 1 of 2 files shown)


evan-charmworks:
See #1915.

minitu (approved):

matthiasdiener:
I'm think this feature needs more discussion. I had left it out intentionally from the cmake build for several reasons:
- This usually does not work as expected in normal cmake build systems (you can't re-run cmake in an already configured folder with different options than in the original cmake invocation and expect it to work without removing at least all of cmake's internal state files)
- Instead of re-running the build script or cmake, you can just run `make` in the already configured folder to continue your build (this works even if there were changes to the source code or the build system itself)
- If your `build` command is even slightly different on the second invocation (even adding something innocuous like `--libdir=`), you'll often get an inconsistent build configuration leading to hard-to-debug bugs. We have seen this often in the old build system.

Imho, unless there are strong reasons to have this functionality, it's safer to always remove the build directory.

matthiasdiener (changes requested):
Blocking this until further discussion.

buildold
```diff
@@ -699,14 +704,29 @@ then
...
 # ===================================================
 
 if [ -f "$DESTINATION/tmp/basics" ]
 then
-  Echo "File structure already created; skipping directly to make"
-else
+  Echo "The configure step has already taken place for the specified build."
+  if [ -z "$FORCE" ]
+  then
+    Echo "To attempt a partial rebuild, run make in \"$DESTINATION/tmp\"."
```
matthiasdiener, line 723:
```suggestion
    Echo "To attempt a partial rebuild, run 'make' in \"$DESTINATION/tmp\"."
```

buildcmake
```diff
@@ -340,6 +341,10 @@ while [[ $# -gt 0 ]]; do
       # Run only cmake, not make
       ONLY_CONFIGURE=1
       ;;
+    --force)
+      # Overwrite existing configuration
```
matthiasdiener, line 345:
```suggestion
      # Overwrite existing build directory
```

buildcmake
```diff
@@ -378,12 +383,55 @@ eval processArgs "$@"
...
+
+
+# Check whether configure step has already taken place
 
+if [[ -f "$builddir/Makefile" ]]
+then
+  echo "The configure step has already taken place for the specified build."
+  if [[ $FORCE -eq 0 ]]
+  then
+    echo "To attempt a partial rebuild, run make in \"$builddir\"."
```
matthiasdiener, line 423:
```suggestion
    echo "To attempt a partial rebuild, run 'make' in \"$builddir\"."
```

matthiasdiener (approved):

matthiasdiener (changes requested):
Wait - have you tested that this works when pressing Ctrl-C in the middle of the cmake invocation? At that point, the Makefile might not exist, but parts of the configuration have already run and won't be re-executed.

Edit: It might be better to check for presence of `$builddir/CMakeCache.txt`.

evan-charmworks:
> Wait - have you tested that this works when pressing Ctrl-C in the middle of the cmake invocation? At that point, the Makefile might not exist, but parts of the configuration have already run and won't be re-executed.

I've run buildcmake in one terminal with `while true ; do [[ -f 'multicore-linux-x86_64/Makefile' ]] && echo 'Exists' && break ; done` in another and it only came back true once CMake was done executing. Is there any value in a partially-completed configure step? In such a scenario buildcmake will always `rm -rf "$builddir"` and start over.

matthiasdiener:
> > Wait - have you tested that this works when pressing Ctrl-C in the middle of the cmake invocation? At that point, the Makefile might not exist, but parts of the configuration have already run and won't be re-executed.
> 
> I've run buildcmake in one terminal with `while true ; do [[ -f 'multicore-linux-x86_64/Makefile' ]] && echo 'Exists' && break ; done` in another and it only came back true once CMake was done executing. Is there any value in a partially-completed configure step? In such a scenario buildcmake will always `rm -rf "$builddir"` and start over.

Ok, thats a fair point. In case of a partial configure, the config is just silently discarded. My confusion was due to the 'silent' aspect.

matthiasdiener (approved):

minitu (approved):

evan-charmworks:
Added an additional check because silently running `rm -rf` sounds like a bad idea, especially with arbitrary `--destination` parameters.

buildcmake
```diff
@@ -378,12 +383,65 @@ eval processArgs "$@"
...
+  echo "The configure step has already taken place for the specified build."
+  if [[ $FORCE -eq 0 ]]
+  then
+    echo "To attempt a partial rebuild, run 'make' in \"$builddir\"."
+    echo "To remove the existing data and reconfigure, run this script again with --force."
+    exit 1
+  else
+    echo "Removing existing data and starting over. (--force)"
+  fi
+elif [[ -e "$builddir" ]]
```
matthiasdiener, line 429: why `-e` and not `-d`? (same below)

evan-charmworks: If someone mistakenly points --destination at an existing file, for example.

matthiasdiener (approved):

stwhite91 (approved):

matthiasdiener:
This PR breaks the spack build:

```bash
$ spack install -j1 -v charmpp@develop
==> Installing charmpp
==> No binary for charmpp found: installing from source
==> charmpp: Executing phase: 'install'
==> [2020-11-14-17:51:05.098606] './build' 'LIBS' 'netlrts-darwin-x86_64' 'clang' 'gfortran' '-j1' '--destination=/Users/mdiener/Work/spack/opt/spack/darwin-macos-skylake/apple-clang-12.0.0/charmpp-develop-5remb5pdx4lkggpjmyd2yxcpnke5khkr' 'smp' '--build-shared' '--with-production'
"/Users/mdiener/Work/spack/opt/spack/darwin-macos-skylake/apple-clang-12.0.0/charmpp-develop-5remb5pdx4lkggpjmyd2yxcpnke5khkr" already exists but is in an inconsistent state.
To remove the existing data and reconfigure, run this script again with --force.
```

PR #3134, opened 2020-10-26
Begin Modernization of our Futures Interface

Wingpad:
This represents a proposal to modernize our futures interface, moving them towards a more object-oriented interface. `ck::future` is templated, and can be used with any pup'able types; pup'ing is automatically handled behind-the-scenes.

`CkFuture CkCreateFuture(void)` -> `ck::future<T>()`
`void CkReleaseFuture(CkFuture fut)` -> `void ck::future<T>::release()`
`int CkProbeFuture(CkFuture fut)` -> `bool ck::future<T>::probe()`
`void *CkWaitFuture(CkFuture fut)` -> `T ck::future<T>::get()`
`void CkSendToFuture(CkFuture fut, void *msg)` -> `void ck::future<T>::set(T)`


examples/charm++/future/fib.C
```diff
@@ -0,0 +1,44 @@
+#include <ck/future.h>
+#include "fib.decl.h"
+
+#define THRESHOLD 3
+
+struct Main : public CBase_Main {
+  Main(CkArgMsg* m) { thisProxy.run(atoi(m->argv[1])); }
```
stwhite91, line 7: delete m

stwhite91 (commented):
This looks much nicer. If there existing documentation for futures, it should be updated, otherwise documentation should be added.

examples/charm++/future/fib.C
Wingpad: I changed the constructor to a `nokeep` entry method, which should resolve this issue. This was copied verbatim from the existing `fib` example, so we may want to fix that at some point as well.

epmikida: At one point, (and maybe still), we planned on making the Main ctor nokeep by default: https://github.com/UIUC-PPL/charm/pull/2423

epmikida: Not sure if we want to revisit that change

Wingpad:
It seems like the Windows build issues stem from using `CkMarshallMsg` in `ckfutures.h`, does anyone have any thoughts on where to put the object-oriented interface so this isn't an issue? Perhaps another header file, if so what would be a good name for it? Something like `<ck/future.hpp>` would be the most modern solution, but that wouldn't fit into typical Charm++ conventions well (and I'm unsure whether the build scripts would handle it correctly).

src/ck-core/charm++.h
```diff
@@ -1315,45 +1308,6 @@ int CkRegisterEp(const std::string& name, CkCallFnPtr call, int msgIdx, int char
...
-    //Set the message's queueing type 
-    env->setQueueing((unsigned char)opts->getQueueing()); 
-  } 
-  return m; 
-} 
-
-
-
-
-
```
minitu, line 1356: A little concerned about this change, did you check if all the codes with `CkAllocateMarshallMsg*` calls now include `ckmarshall.h` instead of (or in addition to) `charm++.h`?

examples/charm++/future/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	$(CHARMC) $(CHARMCLINKFLAGS) -o $@ $+ 
+
+.SECONDARY: $(patsubst %.C,%.decl.h,$(wildcard *.C))
+.SECONDARY: $(patsubst %.C,%.def.h,$(wildcard *.C))
+
+%.o: %.C %.decl.h %.def.h
+	$(CHARMC) $(CHARMCFLAGS) $<
+
+%.decl.h %.def.h: %.ci
+	$(CHARMC) $(CHARMCFLAGS) $<
```
evan-charmworks, line 21: This Makefile construction for building decl/def is unsafe because it will run charmc on the .ci file _twice_, once for each of the two files, and there will sometimes be a race condition in each charmxi process' file writing which results in malformed header output. I recommend defining the build rule only for the decl, and adding a new build rule for the def that is chained to the decl but does nothing of its own.

```suggestion
%.decl.h: %.ci
	$(CHARMC) $(CHARMCFLAGS) $<

%.def.h: %.decl.h
```

doc/charm++/manual.rst
```diff
@@ -5796,19 +5796,21 @@ is created, a *reference* is returned immediately. However, if the
...
-    CkFuture CkCreateFuture(void)
-    void CkReleaseFuture(CkFuture fut)
-    int CkProbeFuture(CkFuture fut)
-    void *CkWaitFuture(CkFuture fut)
-    void CkSendToFuture(CkFuture fut, void *msg)
+   CkFuture CkCreateFuture(void); /* or */ ck::future<T>()
+   void CkReleaseFuture(CkFuture fut); /* or */ void ck::future<T>::release()
+   int CkProbeFuture(CkFuture fut); /* or */ bool ck::future<T>::probe()
+   void *CkWaitFuture(CkFuture fut); /* or */ T ck::future<T>::get()
+   void CkSendToFuture(CkFuture fut, void *msg); /* or */ void ck::future<T>::set(T)
```
evan-charmworks, line 5808: It might be more readable to either make these two separate lists, or to make a table with a C column and a C++ column.

src/ck-core/charm++.h
Wingpad: Since `ckmarshall.h` is included in `charm++.h`, it shouldn't be necessary to include it separately. This separation _should_ only be relevant for files included in `charm++.h`, such as `ckfutures.h`; everything related to marshal messages may not be accessible yet otherwise, and I encountered three different issues related to this prior to this separation.

examples/charm++/future/Makefile
Wingpad: Thanks! This was copied verbatim from `examples/charm++/fib` so we may want to backport our preferred fix.

src/ck-core/charm++.h
minitu: Ah I missed that, thanks

evan-charmworks (approved):

stwhite91 (approved):

PR #3138, opened 2020-10-26
cmake,doc: fix LB build doc

### LINKED ISSUE #3136: Make error when building individual LB on v6.11.0-beta1
Following the documentation for [building individual LBs](https://charm.readthedocs.io/en/latest/charm++/manual.html?highlight=CkStartLB#compiler-and-runtime-options-to-use-load-balancing-module), subsection 2, leads to the following error:
```
$ cd multicore-linux-x86_64/tmp
$ make MetisLB
make: *** No rule to make target 'MetisLB'.  Stop.
```

In the previous v6.10.2, this command would create a module.

Perhaps this change is intended with the new build system, in which case all that is needed is a documentation update.

matthiasdiener:
Fixes #3136.


doc/charm++/manual.rst
```diff
@@ -3152,13 +3141,12 @@ Below are the descriptions about the compiler and runtime options:
    path to the library and link the load balancer into an application
    using *-module FooLB*.
 
-   You can create a library by modifying the Makefile in the following
-   way. This will create *libmoduleFooLB.a*.
+   You can create a library in the following way. This will create
+   *libmoduleFooLB.a*.
 
-   .. code-block:: makefile
+   .. code-block::
```
evan-charmworks, line 3147: Should this be `:: bash` or `:: none`?

rbuch: Is there a good way of specifying this when building with CMake (as in specifying an specific LB to be built as part of the Charm++ build command)?

Also, perhaps this should still be included under the "Building individual load balancers" heading since this also applies to non-user written LBs.

rbuch: I'm on board with building all LBs by default, I agree that the cost is low and it makes things simpler for users. The only issue is for strategies like ZoltanLB or ScotchLB that have external dependencies, which users should still have some way of easily building.

I think the code sample under "Write and use your own load balancer" was intended to be put into application Makefiles, but I'm not sure.

evan-charmworks (approved):

minitu (approved):

Issue #3131, opened 2020-10-25
+balancer help doesn't show TreeLB sub-balancers


evan-charmworks:
From @fmahebert via https://github.com/UIUC-PPL/charm/issues/3123#issuecomment-716086771

> However... after compiling with `-module EveryLB`, the output of `./hello +balancer help` suggests that the majority of the LBs that were available on v6.10.2 are not available on v6.11.0-beta1. Only these are displayed:
> 
> ```
> Available load balancers:
> * TreeLB:	TreeLB
> * DistributedLB:	The distributed load balancer
> * CentralLB:	CentralLB base class
> * MetisLB:	Use Metis(tm) to partition object graph
> ```

If this message can't list all the TreeLB subtypes, it should at least indicate they exist and maybe where a list can be found.

Maybe CentralLB should be hidden from this list as well.

matthiasdiener:
Could also replace the CkAbort with CkExit.

PR #3140, opened 2020-10-28
LB #3131: Show pluggable TreeLB strategies in help message

rbuch:
I'm not 100% sold on using using macros for registration and name printing as done here, but I think it's the only real way to map class names to strings "automatically." I think ideally the strategy classes would register themselves using a `map` in the factory and some static initialization, but that gets a little bit complicated with the templating here.

### DIFF
--- a/src/ck-ldb/LBManager.C
+++ b/src/ck-ldb/LBManager.C
@@ -122,7 +122,7 @@ static void createLoadBalancer(const char* lbname, const char* legacybalancer = 
   {  // invalid lb name
     CmiPrintf("Abort: Unknown load balancer: '%s'!\n", lbname);
     lbRegistry.displayLBs();  // display help page
-    CkAbort("Abort");
+    CkExit(1);
   }
   // invoke function to create load balancer
   int seqno = LBManagerObj()->getLoadbalancerTicket();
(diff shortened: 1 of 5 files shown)


src/ck-ldb/LBManager.C
```diff
@@ -122,7 +122,7 @@ static void createLoadBalancer(const char* lbname, const char* legacybalancer =
   {  // invalid lb name
     CmiPrintf("Abort: Unknown load balancer: '%s'!\n", lbname);
     lbRegistry.displayLBs();  // display help page
-    CkAbort("Abort");
+    CkExit(1);
```
matthiasdiener, line 125: maybe:
```suggestion
    if(lbname == "help)
      CkExit(0);
    else
      CkExit(1);
```

src/ck-ldb/LBManager.C
```diff
@@ -39,14 +39,14 @@ class LBDBRegistry
   // table for all available LBs linked in
   struct LBDBEntry
   {
-    const char* name;
+    std::string name;
     LBCreateFn cfn;
     LBAllocFn afn;
-    const char* help;
+    std::string help;
     int shown;  // if 0, donot show in help page
```
matthiasdiener, line 46:
```suggestion
    bool shown;  // if false, do not show in help page
```

evan-charmworks (approved):

matthiasdiener (approved):

Issue #2763, opened 2020-03-12
entry_method_post_api/*/simpleZeroCopy/ example failing


stwhite91:
```
2020-03-11T15:23:27.7089464Z ../../../../../../bin/testrun  +p4 ./simpleZeroCopy 32 +balancer GreedyLB 
2020-03-11T15:23:27.7236422Z 
2020-03-11T15:23:27.7237103Z Running on 4 processors:  ./simpleZeroCopy 32 +balancer GreedyLB 
2020-03-11T15:23:27.7373418Z charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 4  ./simpleZeroCopy 32 +balancer GreedyLB 
2020-03-11T15:23:27.8076073Z Charm++> Running in non-SMP mode: 4 processes (PEs)
2020-03-11T15:23:27.8076547Z Converse/Charm++ Commit ID: v6.10.0-rc2-169-gd3ee46c1e
2020-03-11T15:23:27.8222326Z Isomalloc> Synchronized global address space.
2020-03-11T15:23:27.8226712Z CharmLB> Load balancer assumes all CPUs are same.
2020-03-11T15:23:27.8421436Z Charm++> Running on 1 hosts (1 sockets x 2 cores x 1 PUs = 2-way SMP)
2020-03-11T15:23:27.8541477Z Charm++> cpu topology info is gathered in 0.030 seconds.
2020-03-11T15:23:27.8549940Z CharmLB> GreedyLB created.
2020-03-11T15:23:27.8582081Z send: completed
2020-03-11T15:23:27.8593218Z zerocopySend: completed
2020-03-11T15:23:27.9062422Z mixedSend: completed
2020-03-11T15:23:27.9263670Z sdagRun: Iteration 2 completed
2020-03-11T15:23:27.9278145Z sdagRun: Iteration 3 completed
2020-03-11T15:23:27.9467928Z sdagRun: Iteration 4 completed
2020-03-11T15:23:28.5103841Z sdagRun: Iteration 5 completed
2020-03-11T15:23:28.5582914Z sdagRun: Iteration 6 completed
2020-03-11T15:23:28.6221010Z sdagRun: Iteration 7 completed
2020-03-11T15:23:28.6861134Z sdagRun: Iteration 8 completed
2020-03-11T15:23:28.7513958Z sdagRun: Iteration 9 completed
2020-03-11T15:23:28.9982832Z sdagRun: Iteration 10 completed
[...]

PR #3151, opened 2020-11-12
Fix rare bug seen in ZC EM Post API examples

nitbhat:
The bug was causing a segfault because of the execution of the POST
entry method before the allocation of the receive buffers. This was
happening because the allocation was not happening in the object's
constructor. The fix involved moving the allocation of receive buffers to the
constructor of the object. This ensures that the buffers are
allocated before the POST entry method executes.

### DIFF
--- a/examples/charm++/zerocopy/entry_method_post_api/prereg/simpleZeroCopy/simpleZeroCopy.C
+++ b/examples/charm++/zerocopy/entry_method_post_api/prereg/simpleZeroCopy/simpleZeroCopy.C
@@ -8,6 +8,7 @@
 #define TOTAL_ITER    40
 
 int numElements;
+CProxy_Main mProxy;
 
 //Main chare
 class Main : public CBase_Main{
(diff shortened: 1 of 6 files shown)


minitu:
Is there a reason why `mainProxy` cannot be a global readonly variable?

evan-charmworks:
This looks like a candidate for fixing #2763

nitbhat:
> Is there a reason why `mainProxy` cannot be a global readonly variable?

No, not really. Just fixed this in the last commit.

nitbhat:
> This looks like a candidate for fixing #2763

Yes, I thought so too. (Esp the failures seen right at the beginning definitely could be due to this). Additionally, quite a few of the stack traces refer to `zerocopySend` too. I'm not sure about the failures that are occurring at the end of the `sdagRun` execution, there might be another bug there.

evan-charmworks (approved):

minitu (approved):

PR #3160, opened 2020-11-17
Link Converse as one library

evan-charmworks:
This combines conv-core, conv-util, conv-partition, conv-ldb, conv-machine, and tmgr.


CMakeLists.txt
```diff
@@ -741,9 +741,172 @@ add_dependencies(charmxi create_symlinks)
...
+    src/util/pup_toNetwork.C
+    src/util/pup_toNetwork4.C
+    src/util/pup_util.C
+    src/util/pup_xlater.C
+    src/util/spanningTree.C
+)
+
+configure_file(${conv-util-h-sources} ${CMAKE_BINARY_DIR}/include/ COPYONLY)
+
+if(${CMK_CAN_LINK_FORTRAN})
```
matthiasdiener, line 899:
```suggestion
if(CMK_CAN_LINK_FORTRAN)
```

CMakeLists.txt
```diff
@@ -741,9 +741,172 @@ add_dependencies(charmxi create_symlinks)
 # Converse
 add_subdirectory(src/QuickThreads)
 add_subdirectory(src/util/boost-context)
+
```
matthiasdiener, line 744: I'd consider putting these into a separate file, maybe `cmake/converse.cmake`

matthiasdiener (commented):
Could you mention here why these libraries are merged, and not e.g. conv-ldb or ckqt as well?

CMakeLists.txt
evan-charmworks: Would that require that all paths have `../` at the beginning?

evan-charmworks (commented):
It looks like this PR is exposing previously rare issues with decl/def not being generated when expected, affecting the CkMarshall and MetaBalancer objects.

> Could you mention here why these libraries are merged, and not e.g. conv-ldb or ckqt as well?

conv-ldb is an oversight since it is added to the link line from a different place in charmc. Ideally I'd think ckqt should only be linked with the libthreads library that makes use of it.

CMakeLists.txt
matthiasdiener: Probably, yes

matthiasdiener: Wait, if you `include()` the cmake file, paths should still be relative to the main CmakeLists.txt, i.e., without `../`.

matthiasdiener (dismissed):
For the future, is there any reason not to do the same with the charm libraries? Only reason I can think of is when multiple libraries contain the same symbol, which is hopefully rare.

evan-charmworks:
> For the future, is there any reason not to do the same with the charm libraries? Only reason I can think of is when multiple libraries contain the same symbol, which is hopefully rare.

Other than charmc's configurable link-time options like -memory and -thread, Charm++ could also benefit from this. I think that would just be combining libcompletion and libNDMeshStreamer into libck, but it would need to be investigated specifically. As discussed above, libckqt could also be combined into the correct libthreads-*.

stwhite91 (approved):
It would be good to open an issue to simplify / consolidate the rest of the libraries as you mentioned

matthiasdiener (approved):

PR #3162, opened 2020-11-23
CMake: Output simplearrayhello objects to its own folder

evan-charmworks:


matthiasdiener:
What's the reason for the change? I was thinking not to compile in the simplearrayhello/ folder in order not to confuse the user .

evan-charmworks:
@matthiasdiener My only concern is that the files are generated in CMAKE_SOURCE_DIR/* instead of CMAKE_BINARY_DIR/* (like Metis was previously). I'm not attached to this way of fixing that.

matthiasdiener:
> @matthiasdiener My only concern is that the files are generated in CMAKE_SOURCE_DIR/* instead of CMAKE_BINARY_DIR/* (like Metis was previously). I'm not attached to this way of fixing that.

Which files are you talking about? Nothing related to simplearrayhello should be generated in `CMAKE_SOURCE_DIR`. it should generate: 
- netlrts-darwin-x86_64/CMakeFiles/hello.dir/tests/charm++/simplearrayhello/hello.C.o
- netlrts-darwin-x86_64/bin/hello
- netlrts-darwin-x86_64/include/hello.{decl, def}.h

evan-charmworks:
My mistake.

PR #3163, opened 2020-11-23
Embed hwloc into libconverse

### LINKED ISSUE #2599: automake: warning: am_libhwloc_embedded_la_rpath was already defined in condition TRUE, which includes condition !HWLOC_BUILD_STANDALONE

evan-charmworks:
Fixes #2599

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -779,7 +779,41 @@ ExternalProject_Add(hwloc
     COMMAND cp -LRf ${CMAKE_SOURCE_DIR}/contrib/hwloc/include/hwloc ${CMAKE_BINARY_DIR}/include/
     COMMAND cp -f ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/include/hwloc/autogen/config.h ${CMAKE_BINARY_DIR}/include/hwloc/autogen/
 )
-file(GLOB hwloc-objects ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/*.o)
+
+set(hwloc-objects
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/base64.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/bind.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/bitmap.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/components.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/diff.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/distances.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/memattrs.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/misc.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/pci-common.o
+  ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/shmem.o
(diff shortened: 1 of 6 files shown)


evan-charmworks:
This is WIP. It works for buildold but not for CMake. The current problem is that `file(GLOB hwloc-objects ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/*.o)` runs during configure time instead of after the hwloc target builds. I'm not sure how much CMake hackery it will take to get these objects added to the converse target.

matthiasdiener:
> This is WIP. It works for buildold but not for CMake. The current problem is that `file(GLOB hwloc-objects ${CMAKE_BINARY_DIR}/hwloc-prefix/src/hwloc-build/hwloc/*.o)` runs during configure time instead of after the hwloc target builds. I'm not sure how much CMake hackery it will take to get these objects added to the converse target.

Instead of globbing, why not just list them manually? You might also have to add SET_SOURCE_FILES_PROPERTIES like here: https://stackoverflow.com/questions/38609303/how-to-add-prebuilt-object-files-to-executable-in-cmake

evan-charmworks:
> Instead of globbing, why not just list them manually?

I was hoping to avoid that, because such a list would be fragile with regard to different platforms, configuration options, hwloc version updates. For example, topology-{linux,darwin,windows}.o and topology-cuda.o.

matthiasdiener:
> > Instead of globbing, why not just list them manually?
> 
> I was hoping to avoid that, because such a list would be fragile with regard to different platforms, configuration options, and hwloc version updates. For example, topology-{linux,darwin,windows}.o and topology-cuda.o.

Ok, fair. Can you check - does it work to add the .a instead of the .o?

evan-charmworks:
> Ok, fair. Can you check - does it work to add the .a instead of the .o?

If I add the .a to the converse target and also add SET_SOURCE_FILES_PROPERTIES for it, it shows up in the build line for libconverse.a (shown by `VERBOSE=1 make`), but charmc does not do anything with it and linking hello still fails.

matthiasdiener:
> > Instead of globbing, why not just list them manually?
> 
> I was hoping to avoid that, because such a list would be fragile with regard to different platforms, configuration options, and hwloc version updates. For example, topology-{linux,darwin,windows}.o and topology-cuda.o.

I've implemented something without globbing. Are we relay that platform-dependent? All I could see was `topology-{linux,darwin,windows}.o` CUDA support is always disabled.

evan-charmworks:
Windows:
```
make[2]: *** No rule to make target 'hwloc-prefix/src/hwloc-build/hwloc/base64.o', needed by 'lib/libconverse.a'.  Stop.
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:624: CMakeFiles/converse.dir/all] Error 2
```

Maybe we need to define the objects as outputs of the hwloc target somehow.

matthiasdiener:
> Windows:
> 
> ```
> make[2]: *** No rule to make target 'hwloc-prefix/src/hwloc-build/hwloc/base64.o', needed by 'lib/libconverse.a'.  Stop.
> make[2]: *** Waiting for unfinished jobs....
> make[1]: *** [CMakeFiles/Makefile2:624: CMakeFiles/converse.dir/all] Error 2
> make[1]: *** Waiting for unfinished jobs....
> ```

Should be fixed now.

> Maybe we need to define the objects as outputs of the hwloc target somehow.

Unfortunately, I think there is no other way then listing them in the way I did.

evan-charmworks:
> Unfortunately, I think there is no other way then listing them in the way I did.

I didn't think about the file extension difference. I was wondering why this error happened even though we specify that the converse target depends on the hwloc target. I think this is sufficient.

evan-charmworks:
Ready for review

matthiasdiener (approved):

minitu (approved):

PR #3170, opened 2020-11-27
Create AtSync EndReceiver to simplify CkLocMgr

rbuch:


epmikida (dismissed):

evan-charmworks (approved):

epmikida (dismissed):

rbuch:

evan-charmworks (dismissed):

src/ck-core/cksyncbarrier.C
```diff
@@ -131,39 +145,33 @@ void CkSyncBarrier::propagate_atsync()
...
-  if (client_count == 1 && !flood_atsync)
-  {
-    if (clients.front()->chare->isLocMgr()) return;
-  }
-
-  // If there are no clients, resume as soon as we're turned on
-  if (client_count == 0)
-  {
-    cur_epoch++;
-    CallReceivers();
```
kavithachandrasekar, line 165: This CallReceivers() calls ResumeFromSync on chares for Periodic LB I think. It probably needs to be added back to the code. Also, with LB refactor I think the part which corresponds to this, that added receivers for Periodic LB (!usesAtSync) was incorrectly removed from ckFinishConstruction() in cklocation.C.

rbuch: Should `ResumeFromSync` be called in periodic LB? The documentation only states that it is called when using `AtSync` mode, which I think is the only sensible mode to call it in, since it's the only one that offers a causality chain for it.

kavithachandrasekar: Sounds good, I guess we can remove CallReceivers() if there's no use case for it. I am not sure what the previous use case was, but the older code before LB refactor used to call ResumeFromSync (if implemented) for periodic LB.

rbuch: Okay, I think removing it makes use cleaner. I think the previous behavior was incidental and not an intentional decision.

rbuch:
Abandoned in favor of #3248.

PR #3171, opened 2020-11-30
LB: Remove typedef for structs in lbdb.h

rbuch:
Pursuant to P1766R1 <http://wg21.link/p1766r1>, remove typedef name
for linkage on non-C compatible structs.

Clang 11 warns about this, this patch fixes those warnings.

### DIFF
--- a/src/ck-ldb/lbdb.h
+++ b/src/ck-ldb/lbdb.h
@@ -36,44 +36,44 @@ extern int _lb_version;
   /*  as OBJ_ID_SZ */
 
 #if CMK_LBDB_ON
-typedef struct {
+struct LDHandle {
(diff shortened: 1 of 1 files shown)


matthiasdiener (approved):
By the way, why aren't these classes?

rbuch:
> By the way, why aren't these classes?

I don't think there's really any reason why other than the inertia of the past. I think most of them "behave" like structs (as in the fields can vary independently, see: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-struct), so I think it's sensible to leave them that way, but that's a very slight preference.

evan-charmworks (approved):

PR #3175, opened 2020-12-01
change name of probe to is_ready within ck::future

Wingpad:
ensures consistency with C++ futures. this feature is likely not used yet, so this is the optimal time to change it


evan-charmworks (approved):

evan-charmworks:
@matthiasdiener It looks like the Spack CI cannot handle branches from external repositories.

matthiasdiener:
> @matthiasdiener It looks like the Spack CI cannot handle branches from external repositories.

Yeah, that's true. Feel free to merge this with the test failing for now, passing the test is not required for a merge.

matthiasdiener:
For pull requests from forks, we will need to merge #3177 first.

stwhite91 (approved):

src/ck-core/ckfutures.h
```diff
@@ -84,7 +84,7 @@ namespace ck {
       CkSendToFuture(handle_, msg);
     }
 
-    bool probe() { return CkProbeFuture(handle_); }
+    bool is_ready() { return CkProbeFuture(handle_); }
```
NK-Nikunj, line 87: In reference to Prof. Kale's comment on the proposed change, can we keep both `is_ready` and `probe` as synonyms to keep it similar to both C++ future's member function `is_ready`, and MPI's `probe` function?

NK-Nikunj (changes requested):

src/ck-core/ckfutures.h
Wingpad: Probe is retained within the C-like interface, which can serve as a frame-of-reference for MPI users. For this reason, and the fact that they are documented side-by-side, I do not think an alias is necessary within the C++ interface.

NK-Nikunj: Thanks for the clarification. LGTM thanks!

NK-Nikunj (approved):

PR #3183, opened 2020-12-04
Tracing: Add username to Projections logs

rbuch:


matthiasdiener (approved):

src/ck-perf/trace-common.C
```diff
@@ -19,6 +19,14 @@
 #include "allEvents.h"          //projector
 #include "register.h" // for _entryTable
 
+// To get username
+#if defined(_WIN32) || defined(_WIN64)
```
evan-charmworks, line 23: Not terribly important, but `#ifdef _WIN32` would be sufficient here. It is present for both 32-bit and 64-bit.

evan-charmworks (approved):

PR #3186, opened 2020-12-08
Tracing: Add hostname to Projections logs

rbuch:
In making this patch, I also discovered that we currently have a `gethostname` check in `configure.ac`. I don't think this is really needed (plus, the result `CMK_GET_HOSTNAME` is already not universally used) since both POSIX and Windows guarantee this function exists. Am I missing some reason why we need this check or can it be removed?

### DIFF
--- a/src/ck-perf/trace-common.C
+++ b/src/ck-perf/trace-common.C
@@ -21,10 +21,30 @@
 
 // To get username
 #if defined(_WIN32) || defined(_WIN64)
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <Lmcons.h>
 #else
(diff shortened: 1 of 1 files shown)


matthiasdiener (approved):

evan-charmworks (approved):
> In making this patch, I also discovered that we currently have a `gethostname` check in `configure.ac`. I don't think this is really needed (plus, the result `CMK_GET_HOSTNAME` is already not universally used) since both POSIX and Windows guarantee this function exists. Am I missing some reason why we need this check or can it be removed?

I don't see a need to keep it. I checked old autobuilds and the test passes on BG/Q too.

PR #3188, opened 2020-12-08
Use Winsock2.h instead of deprecated winsock.h in ntohl check

rbuch:


evan-charmworks:
Should the use in src/conv-core/debug-conv.C also be fixed?

rbuch:
> Should the use in src/conv-core/debug-conv.C also be fixed?

Indeed it should, I pushed a fixup commit to do so.

evan-charmworks (approved):

minitu (approved):

PR #3189, opened 2020-12-08
Add new Isomalloc APIs useful for AMPI privatization

evan-charmworks:


src/conv-core/isomalloc.C
```diff
@@ -1025,6 +1029,37 @@ struct isommap
...
+
+    CmiUnlock(lock);
+    return mapped;
+  }
+
+  void * permanent_alloc(size_t size, size_t align)
+  {
+    const size_t realsize = CMIALIGN(size, pagesize);
+
+    CmiLock(lock);
```
stwhite91, line 1050: Here and above, can we just lock around the call to map_global_memory? We always want to keep critical sections as short as possible when using locks. And if so, should the locks be moved inside map_global_memory?

evan-charmworks: The locks are here to provide exclusive access to the contents of the isommap struct, particularly the allocated_extent member. In the general case the locks should be uncontended because each AMPI rank has its own context. They are only here for the case of something like AMPI+OpenMP.

matthiasdiener: Can you add this as a comment?

src/conv-core/isomalloc.C
```diff
@@ -1017,6 +1017,10 @@ struct isommap
     }
   }
```
matthiasdiener, line 1019: Can you document the purpose of this function?

src/conv-core/isomalloc.C
evan-charmworks: > Can you add this as a comment?

Added to the definition of the lock in `struct isommap`.

src/conv-core/isomalloc.C
evan-charmworks: Added a comment to its definition.

matthiasdiener (approved):

evan-charmworks:
Needs one more review

stwhite91 (approved):

PR #3195, opened 2020-12-11
Cleanup CkSyncBarrier code

rbuch:
Fixes stylistic issues, particularly with variable and method names. Additionally, changes the semantics of the kick epoch, previously it would trigger the barrier when it was greater than or equal to the current epoch, now it must be strictly greater than the current epoch to trigger the barrier (which mirrors the behavior of regular AtSync elements).

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -1511,7 +1511,7 @@ CkMigratable::~CkMigratable() {
 	if (barrierRegistered) {
 	  DEBL((AA "Removing barrier for element %s\n" AB,idx2str(thisIndexMax)));
 	  if (usesAtSync)
-	    myRec->getSyncBarrier()->RemoveClient(ldBarrierHandle);
+	    myRec->getSyncBarrier()->removeClient(ldBarrierHandle);
 	}
 
   if (_lb_args.metaLbOn()) {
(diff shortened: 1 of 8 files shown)


evan-charmworks:
Is "Kick" a standard term? What exactly does it mean in this context?

rbuch:
> Is "Kick" a standard term? What exactly does it mean in this context?

It is a standard term (at least in related contexts, like watchdog timers, although the purpose is reversed there). The point here is to kick PEs with no AtSync elements on them so that they participate in the AtSync barrier (so that they can do things like participate in a group reduction to start load balancing). When there's no AtSync element on the PE, nothing local is going to trigger it, so this kick needs to be propagated in order to do so.

If there are any naming or other suggestions, I'd love to hear them, I want to make this as clear as possible.

evan-charmworks:
Some ideas: prompt, trigger, induce, invoke, initiate

rbuch:
> Some ideas: prompt, trigger, induce, invoke, initiate

I think I prefer "kick" over these since it seems more clearly distinct from "triggering the barrier," which seems like the most natural term to call reaching the barrier and executing the receivers, and it has the existing similar usage in the watchdog context.

stwhite91 (dismissed):

epmikida (approved):

evan-charmworks (approved):

epmikida (approved):

PR #3196, opened 2020-12-12
Documentation: Refactor AMPI manual section on privatization methods

evan-charmworks:


evan-charmworks:
Best viewed with `git diff master... --color-moved`

doc/ampi/manual.rst
```diff
@@ -519,53 +519,227 @@ provided that it takes place *after* ``MPI_Init``.
...
-are not careful a wrong result may be produced by the program. Figure
+there is a single instance of those variables, so they will be shared
+among different threads in the single address space, and this could
+lead to the program producing an incorrect result or crashing. Figure
 :numref:`fig_global` shows an example of a multi-threaded
 application with two threads in a single process. :math:`var` is a
 global or static variable in this example. Thread 1 assigns a value to
 it, then it gets blocked for communication and another thread can
-continue. Thereby, thread 2 is scheduled next and accesses :math:`var`
+continue. Then, thread 2 is scheduled next and accesses :math:`var`,
```
matthiasdiener, line 536: Is this example (and the figure below) really helpful or necessary?

doc/ampi/manual.rst
```diff
@@ -519,53 +519,227 @@ provided that it takes place *after* ``MPI_Init``.
...
 -----------------------------
 
-For the before-mentioned benefits to be effective, one needs to map
-multiple user-level threads onto each processor. Traditional MPI
-programs assume that the entire processor is allocated to themselves,
-and that only one thread of control exists within the process’s address
-space. So, they may safely use global and static variables in the
-program. However, global and static variables are problematic for
+For the aforementioned benefits to be effective, AMPI needs to map
+multiple user-level threads onto each OS process. Traditional MPI
```
matthiasdiener, line 523: This makes it sound like only user-level threads are the issue, but it might also be kernel-level threads right?

Maybe:
```suggestion
AMPI maps multiple threads onto each OS process. Traditional MPI
```

evan-charmworks: User-level threads alone are enough to cause privatization issues. The concurrency introduced by kernel threads could expose additional issues, but it might also obscure them nondeterministically. Maybe this section could say something like "AMPI maps multiple ranks onto each OS process in the form of user-level threads, which can execute cooperatively, concurrently, or some combination of both depending on how the Charm++ runtime system was built." Is that too much detail?

doc/ampi/manual.rst
evan-charmworks: I think _some_ example is important to have here, for illustration. I don't know if the figure is the best approach. Instead we could always use this simpler example from presentations I have given:

```c
int rank_global;

void print_ranks(void)
{
  MPI_Comm_rank(MPI_COMM_WORLD, &rank_global);

  MPI_Barrier(MPI_COMM_WORLD);

  printf("rank: %d\n", rank_global);
}
```

matthiasdiener: > I think _some_ example is important to have here, for illustration. I don't know if the figure is the best approach. Instead we could always use this simpler example from presentations I have given:
> 
> ```c
> int rank_global;
> 
> void print_ranks(void)
> {
>   MPI_Comm_rank(MPI_COMM_WORLD, &rank_global);
> 
>   MPI_Barrier(MPI_COMM_WORLD);
> 
>   printf("rank: %d\n", rank_global);
> }
> ```

I think this example (+ 1 phrase of explanation or so) works much better.

doc/ampi/manual.rst
matthiasdiener: > User-level threads alone are enough to cause privatization issues. The concurrency introduced by kernel threads could expose additional issues, but it might also obscure them nondeterministically. Maybe this section could say something like "AMPI maps multiple ranks onto each OS process in the form of user-level threads, which can execute cooperatively, concurrently, or some combination of both depending on how the Charm++ runtime system was built." Is that too much detail?

I think that might be a bit too verbose. Maybe:

"In AMPI, ranks are implemented as user-level threads that run on top of OS processes or OS threads, depending on the how the  Charm++ runtime was built."

matthiasdiener (dismissed):

evan-charmworks:
@matthiasdiener It seems like the Travis ARM build is stalling forever when exiting early as part of the check for only `doc/` being changed. Do you know why that would happen?

matthiasdiener (approved):

matthiasdiener:
Needs another review

rbuch (approved):

PR #3199, opened 2020-12-13
Enable PUP'ing of C++ Shared Pointers

Wingpad:
This patch enables PUP'ing of C++ shared pointers, where previously we could only PUP unique pointers. This limitation seems rather arbitrary to me, given we're on C++11 and should have support for both.


NK-Nikunj (approved):
LGTM! Thanks for adding it.

stwhite91 (approved):

PR #3200, opened 2020-12-13
charmc: allow extra options to override standard options

matthiasdiener:
Allows specifying (e.g.) -std=c++14 meaningfully

### DIFF
--- a/src/scripts/charmc
+++ b/src/scripts/charmc
@@ -73,6 +73,8 @@ OPTS_CC=''
 OPTS_CXX=''
 OPTS_F90=''
 OPTS_LD=''
+# Additional options specified by the user
+ARG_OPTS=()
 ARG_OPTS_CPP=()
 ARG_OPTS_LDRO=()
 ARG_OPTS_CC=()
@@ -400,7 +402,7 @@ do
 		# Make sure these compiler options are passed through even if no
 		# file to compile is specified
 		PASSTHROUGH_COMPILER_OPTIONS+=("$arg")
-		OPTS+=("$arg")
+		ARG_OPTS+=("$arg")
 		;;
 
 	"-machine")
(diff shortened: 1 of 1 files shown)


src/scripts/charmc
```diff
@@ -946,7 +948,7 @@ do
 
 	-*|+*)
 #		Warning "passing unrecognized option $arg to all compilers and linkers"
-		OPTS="$OPTS  $arg"
+        OPTS_EXTRA+=" $arg"
```
evan-charmworks, line 951:
```suggestion
		OPTS_EXTRA+=" $arg"
```

src/scripts/charmc
```diff
@@ -957,7 +959,7 @@ do
 	*)
 #		printUsage  "Error: Unrecognized argument $arg"
 #		Warning "passing unrecognized option $arg to all compilers and linkers"
-		OPTS="$OPTS  $arg"
+        OPTS_EXTRA+=" $arg"
```
evan-charmworks, line 962:
```suggestion
		OPTS_EXTRA+=" $arg"
```

evan-charmworks:
This PR changes charmc behavior so that arbitrary options passed to charmc take precedence over arguments specified with `-cc-option` et al, instead of the other way around as it is currently. Would it be possible to append ARG_OPTS_* to OPTS_* after OPTS_EXTRA instead of before it?

rbuch (approved):

matthiasdiener:
> This PR changes charmc behavior so that arbitrary options passed to charmc take precedence over arguments specified with `-cc-option` et al, instead of the other way around as it is currently. Would it be possible to append ARG_OPTS_* to OPTS_* after OPTS_EXTRA instead of before it?

I genuinely don't know which way is better here. Are there any known users of `-cc-option` etc.?

evan-charmworks (approved):

PR #3201, opened 2020-12-13
charmc: replace '[' with '[['

matthiasdiener:


src/scripts/charmc
```diff
@@ -2090,21 +2081,21 @@ fi
...
+if [[ "$MPI_INTEROPERATE" = "" ]]
 then
 case "$OBJECT" in
 *.o)
 	Debugf "Converting $FILES to $OBJECTFILES"
 	nFiles=`echo $FILES | wc -w`
 	nObj=`echo $OBJECTFILES | wc -w`
 # One source file, one .o: just move the .o
-	if [ $OBJECT != $OBJECTFILES ]
+	if [[ " $OBJECT" != "$OBJECTFILES" ]]
```
evan-charmworks, line 2098:
```suggestion
	if [[ "$OBJECT" != "$OBJECTFILES" ]]
```

matthiasdiener: That won't work, since $OBJECTFILES has a leading space in the single file case.

evan-charmworks: Something is not right here, because the `mv` command in the `then` clause of this `if` statement is what is failing on Darwin builds. In any case, could this be noted in a comment?

matthiasdiener:
Done in #3709.

PR #3203, opened 2020-12-14
Fix Future Creation and Memory Management

Wingpad:
- ~~Switch to Factory Method for Future Creation~~
- Add no-op `PUP::reconstruct` constructor
- Postpone Message Deletion until Release
- Update Docs and Example

Creating futures in the default constructor led to spurious future creation when pup'ing, necessitating an empty `PUP::reconstruct` constructor.

Deleting the message after the first `get` led to failures when calling `get` multiple times. Instead, message deletion is now delayed until release which checks to see if there is a msg and deletes it.

Also added a sensible error message for currently unsupported use-cases (i.e. non-local value retrieval).


evan-charmworks:
Instead of interfering with the default constructor, would it work to leave it unchanged and instead define an empty `future(PUP::reconstruct) { }`?

Wingpad:
> Instead of interfering with the default constructor, would it work to leave it unchanged and instead define an empty `future(PUP::reconstruct) { }`?

Oho! I wasn't aware of this method for enabling PUP recreation. This is much better ^_^

minitu (dismissed):

src/ck-core/ckfutures.h
```diff
@@ -84,8 +88,14 @@ namespace ck {
...
 
-    bool is_ready() { return CkProbeFuture(handle_); }
-    void release() { CkReleaseFuture(handle_); }
+    CkFuture handle() const { return handle_; }
+    bool is_ready() const { return CkProbeFuture(handle_); }
+    void release() {
+      if ((handle_.pe == CkMyPe()) && is_ready()) {
+        delete (CkMarshallMsg *)CkWaitFuture(handle_);
+      }
+      CkReleaseFuture(handle_);
```
rbuch, line 97: This also needs to be called on the `handle_.pe`, I think.

rbuch (approved):

minitu (approved):

PR #3214, opened 2020-12-22
LB: Don't show CentralLB in help message

rbuch:


src/ck-ldb/CentralLB.C
```diff
@@ -45,20 +45,26 @@ CkGroupID loadbalancer;
 int * lb_ptr;
 bool load_balancer_created;
 
-CreateLBFunc_Def(CentralLB, "CentralLB base class")
```
matthiasdiener, line 48: Would it make sense to add this as an (optional?) parameter to `CreateLBFunc_Def`?

rbuch: Done via #3224.

evan-charmworks (approved):

epmikida (approved):

PR #3215, opened 2020-12-23
Add macOS ARM64 build targets

evan-charmworks:
buildcmake and buildold both tested.

If you follow Homebrew's instructions to install in x86_64 mode while they implement ARM64 support, particularly to install autoconf and automake, be sure to **not** install Make or CMake with it. Let the build process pick up the Make that comes with Xcode, and use an official CMake package with its binary added to PATH. Otherwise the CMake and the compiler will assume x86_64 is the target.


matthiasdiener (approved):

matthiasdiener:
> buildcmake and buildold both tested.
> 
> If you follow Homebrew's instructions to install in x86_64 mode while they implement ARM64 support, particularly to install autoconf and automake, be sure to **not** install Make or CMake with it. Let the build process pick up the Make that comes with Xcode, and use an official CMake package with its binary added to PATH. Otherwise the CMake and the compiler will assume x86_64 is the target.

Out of curiosity, does the x86_64 build work on macOS/ARM?

evan-charmworks:
> Out of curiosity, does the x86_64 build work on macOS/ARM?

buildcmake: No, because the OS and architecture passed in are ignored and it tries building arm8 anyway.
buildold: No, it fails when building boost-context because nothing is telling it to produce x86_64 objects and building arm64 objects out of x86_64 assembly obviously won't work. We would need to add `-arch x86_64` and `-arch arm64` to the right places in the conv-mach*.sh files for this to work.

There is also the possibility to run under Rosetta 2 by prepending `arch -x86_64` to commands, so the build process thinks it is running on x86_64.

buildcmake: Still fails:
```
ld: warning: ignoring file CMakeFiles/conv-cpm.dir/__/xlatcpm/conv-cpm.C.o, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
Undefined symbols for architecture x86_64:
  "_main", referenced from:
      start in crt1.10.6.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Fatal Error by charmc in directory /Users/charmworks/Charmworks/charm/netlrts-darwin-x86_64-bad/src/conv-core
   Command clang++ -multiply_defined suppress -stdlib=libc++ -mmacosx-version-min=10.7 -D_DARWIN_C_SOURCE -ftls-model=initial-exec -rdynamic -o ../../bin/conv-cpm CMakeFiles/conv-cpm.dir/__/xlatcpm/conv-cpm.C.o -L/Users/charmworks/Charmworks/charm/netlrts-darwin-x86_64-bad/lib -Wl,-search_paths_first -Wl,-headerpad_max_install_names returned error code 1
charmc exiting...
make[2]: *** [bin/conv-cpm] Error 1
make[1]: *** [src/conv-core/CMakeFiles/conv-cpm.dir/all] Error 2
```
buildold: Succeeds.

In general, Charm does not have much support for cross-compilation. The only such configuration I know to be supported is building i386 binaries from x86_64, and only with buildold.

minitu (approved):

PR #3216, opened 2020-12-23
LB: Modernize and cleanup MetisLB

rbuch:


src/ck-ldb/ckgraph.h
```diff
@@ -63,10 +63,9 @@ class Edge {
...
       bytes(_bytes) {
     }
-    ~Edge() { }
-    inline int getNeighborId() { return id; }
-    inline int getNumMsgs() { return msgs; }
-    inline int getNumBytes() { return bytes; }
+    inline int getNeighborId() const { return id; }
+    inline int getNumMsgs() const { return msgs; }
+    inline int getNumBytes() const { return bytes; }
     inline void setNumBytes(int _bytes) { bytes = _bytes; }
```
matthiasdiener, line 69: No need for `inline` in these functions (same below).

src/ck-ldb/MetisLB.C
```diff
@@ -5,128 +5,126 @@
...
 CreateLBFunc_Def(MetisLB, "Use Metis(tm) to partition object graph")
 
-MetisLB::MetisLB(const CkLBOptions &opt): CBase_MetisLB(opt)
+    MetisLB::MetisLB(const CkLBOptions& opt)
+    : CBase_MetisLB(opt)
 {
   lbname = "MetisLB";
-  if (CkMyPe() == 0 && !quietModeRequested)
-    CkPrintf("CharmLB> MetisLB created.\n");
+  if (CkMyPe() == 0 && !quietModeRequested) CkPrintf("CharmLB> MetisLB created.\n");
```
matthiasdiener, line 24: fwiw, I prefer the old way (with a line break before the `CkPrintf`).

src/ck-ldb/ckgraph.h
rbuch: Fixed.

src/ck-ldb/MetisLB.C
rbuch: Fixed.

rbuch: If this a thing we want to enforce, we should set `AllowShortIfStatementsOnASingleLine` in the Clang-Format configuration correspondingly.

matthiasdiener (dismissed):
LGTM, but maybe clarify in the commit message that this is removing the WEIGHTED option.

rbuch:
> LGTM, but maybe clarify in the commit message that this is removing the WEIGHTED option.

Done.

src/ck-ldb/MetisLB.C
rbuch: Done in #3229.

src/ck-ldb/ckgraph.h
```diff
@@ -14,159 +14,164 @@
...
-    bool available;	// if the processor is available
+public:
+  ProcInfo() = default;
+  ProcInfo(int i, double ov, double tl, double sp, bool avail)
+      : id(i), _overhead(ov), _totalLoad(tl), _pe_speed(sp), available(avail)
+  {
+  }
+  int getProcId() const { return id; }
+  void setProcId(int _id) { id = _id; }
+  double getTotalLoad() const { return _totalLoad; }
```
epmikida, line 32: Why two accessors for total load?

epmikida (dismissed):

src/ck-ldb/ckgraph.h
epmikida: Also as a followup...do we even want the accessors by reference? are they necessary compared to just get/set methods like are used everywhere else here?

evan-charmworks: Agreed that the accessors need a little bit more cleanup. At the very least, if we want to keep the ones returning by reference, there should be const-qualified versions that return const references. But I'm not sure if we need accessors returning by reference at all.

rbuch: Good call, I've removed the ones that return by reference, I don't think they're needed at all.

evan-charmworks (approved):

epmikida (approved):

stwhite91 (approved):

PR #3218, opened 2020-12-26
Fix Potential Invalid Free Due to Dynamic Cast

Wingpad:
This fixes potential invalid frees that can arise when a dynamic cast "changes" the address of the pointer. This _rarely_ occurs, but complex inheritance relationships can lead to these classes of failures (particularly when `PUP::able` is not the last class that a class publicly inherits). To make this possible, the virtual destructor of `PUP::able` needed to be made `public` from `protected`, but this is consistent with `PUP::er` and, since the destructor is a no-op, does not break encapsulation.

### DIFF
--- a/src/util/pup_stl.h
+++ b/src/util/pup_stl.h
@@ -516,7 +516,12 @@ using Requires = typename requires_impl<
   inline void pup(PUP::er &p, std::shared_ptr<T> &t) {
     PUP::able* _ = (p.isUnpacking()) ? nullptr : t.get();
     p(&_);
-    if (p.isUnpacking()) { t.reset(dynamic_cast<T*>(_)); }
+    if (p.isUnpacking()) {
+      // the shared ptr must be created with the original PUP::able ptr
+      // otherwise the dynamic cast can lead to invalid free's
+      // (it changes the pointer's address)
+      t = std::dynamic_pointer_cast<T>(std::shared_ptr<PUP::able>(_));
+    }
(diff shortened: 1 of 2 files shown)


minitu (approved):

src/util/pup_stl.h
```diff
@@ -516,7 +516,12 @@ using Requires = typename requires_impl<
   inline void pup(PUP::er &p, std::shared_ptr<T> &t) {
     PUP::able* _ = (p.isUnpacking()) ? nullptr : t.get();
     p(&_);
-    if (p.isUnpacking()) { t.reset(dynamic_cast<T*>(_)); }
+    if (p.isUnpacking()) {
+      // the shared ptr must be created with the original PUP::able ptr
+      // otherwise the dynamic cast can lead to invalid free's
+      // (it changes the pointer's address)
```
rbuch, line 522: How would that change the pointer's address?

Wingpad: "With multiple inheritance, different base-class sub-objects might be at different locations within the complete object. In that case, a down-cast might have to adjust the pointer value to point to the complete object." (source: https://stackoverflow.com/a/27320572/1426075)

Wingpad: I suppose better phrasing would be "changes the pointer's **value**" but since it holds an address I slipped a bit.

rbuch (approved):

PR #3219, opened 2020-12-26
Examples: Fix size of array allocation in PUP for jacobi2d-2d-decomposition

rbuch:


stwhite91 (approved):
It would be nicer to have these be std::vector's

matthiasdiener:
Are the other Jacobi examples not affected?

rbuch:
> Are the other Jacobi examples not affected?

No, the others are correct.

rbuch:
> It would be nicer to have these be std::vector's

Agreed, they probably need even more cleanup than that, but I think that should be another PR.

matthiasdiener (approved):

PR #3223, opened 2020-12-28
Fix compiler warning in CkCompactVec

minitu:
`CkCompactVec::lastnull` is defined as a `size_t` but is assigned -1 in various places.


evan-charmworks:
Looking briefly at the rest of class CkCompactVec, I believe this change will introduce off-by-one errors into its implementation. It might be better to replace the `-1`s with `std::numeric_limits<size_t>::max()` which will keep the same bit pattern as -1 but be more semantically appropriate. If that is too long to replace all six uses of -1, maybe add `static constexpr size_t sentinel = std::numeric_limits<size_t>::max();` to the class definition and use that variable instead.

matthiasdiener (approved):

evan-charmworks (approved):

PR #3224, opened 2020-12-29
LB: Modernize registration

rbuch:


evan-charmworks (approved):

src/ck-ldb/BaseLB.h
```diff
@@ -269,37 +269,29 @@ class LBVectorMigrateMsg : public CMessage_LBVectorMigrateMsg {
...
-
-#else		/* CMK_LBDB_ON */
-
-#define CreateLBFunc_Def(x, str)	\
-void Create##x(void) {} 	\
-BaseLB *Allocate##x(void) { return NULL; }	\
-static void lbinit(void) {}
-
-#endif		/* CMK_LBDB_ON */
+#  define CreateLBFunc_Def(x, str)                                                       \
```
Wingpad, line 282: Should we consider adding explicit deprecation warnings here? Among potential solutions, we could generate a deprecated function as part of this macro, and call it within one of the other generated functions.

epmikida (dismissed):
LGTM but I agree with @Wingpad that we should add explicit deprecation if we can. Anything that helps us move towards actual removal of that old code.

I also feel like the usage of lambdas and the templated function is a bit funky, just in that it feels like we are adding yet another piece on top of an already kinda long function call stack just to register an LB. But I think this is definitely a step in the right direction.

src/ck-ldb/BaseLB.h
rbuch: Done.

rbuch:
> LGTM but I agree with @Wingpad that we should add explicit deprecation if we can. Anything that helps us move towards actual removal of that old code.
> 
> I also feel like the usage of lambdas and the templated function is a bit funky, just in that it feels like we are adding yet another piece on top of an already kinda long function call stack just to register an LB. But I think this is definitely a step in the right direction.

I agree it's a bit complex, but I think it's worth doing this for the added flexibility and to remove the reliance on macros.

Wingpad (approved):

evan-charmworks (approved):

PR #3225, opened 2020-12-30
Tracing: Write SMPMODE even when comm thread tracing is disabled

rbuch:


src/ck-perf/trace-common.C
```diff
@@ -264,11 +264,13 @@ void traceWriteSTS(FILE *stsfp,int nUserEvents) {
 #if CMK_SMP_TRACE_COMMTHREAD
```
evan-charmworks, line 264: Could this code be responsible for the empty row that is present if you pass `--enable-tracing-commthread` to a multicore build, and if so, would this fix it?

```suggestion
#if CMK_SMP_TRACE_COMMTHREAD && !CMK_SMP_NO_COMMTHD
```

rbuch: Yeah, that should, although I'd say we should probably prevent building like that in the first place, giving a build time error when someone tries to enable comm thread tracing when there is no comm thread.

rbuch:
The Darwin tracing failure is expected until UIUC-PPL/projections#115 is merged.

src/ck-perf/trace-common.C
```diff
@@ -261,14 +261,16 @@ extern const char* const CmiCommitID;
 /** Write out the common parts of the .sts file. */
 void traceWriteSTS(FILE *stsfp,int nUserEvents) {
   fprintf(stsfp, "MACHINE \"%s\"\n",CMK_MACHINE_NAME);
-#if CMK_SMP_TRACE_COMMTHREAD
+#if CMK_SMP_TRACE_COMMTHREAD && !CMK_SMP_NO_COMMTHD
```
evan-charmworks, line 264: Not to keep causing churn because of this one line, but after thinking more I think this would be more pedantically correct.
```suggestion
#if CMK_SMP_TRACE_COMMTHREAD && CMK_SMP && !CMK_SMP_NO_COMMTHD
```

rbuch: Looks good to me, but what case is this meant to capture that the other defines would not?

evan-charmworks: Yes, exactly that.

evan-charmworks (approved):

src/ck-perf/trace-common.C
rbuch: Gotcha, makes sense.

stwhite91 (approved):

PR #3226, opened 2020-12-31
Enable CkPointer to PUP Any PUP'able Pointer

Wingpad:
`CkPointer` was previously restricted to pointers of types derived from `PUP::able`, this PR enables it to PUP any PUP'able type, and ensures correct restoration when that type has a `PUP::reconstruct` constructor. This is accomplished using the newly introduced `PUP:ptr_helper` struct, which is now used for `std::unique_ptr` and `std::shared_ptr` values of Non-`PUP::able` types as well. Overall, this is a step towards making our handling of pointer types more consistent.

NOTE: Everything here applies to `CkReference` as well, since it is derived from `CkPointer`.


src/util/pup_stl.h
```diff
@@ -520,7 +502,7 @@ using Requires = typename requires_impl<
       // the shared ptr must be created with the original PUP::able ptr
       // otherwise the dynamic cast can lead to invalid free's
       // (it changes the pointer's address)
-      t = std::dynamic_pointer_cast<T>(std::shared_ptr<PUP::able>(_));
+      t = std::static_pointer_cast<T>(std::shared_ptr<PUP::able>(_));
```
Wingpad, line 505: (We know this is a valid cast so we can use `static` instead of `dynamic` cast.)

rbuch (commented):
I would  update the constructors of `CkPointer` to be a bit cleaner here, too.

Wingpad:
> I would update the constructors of `CkPointer` to be a bit cleaner here, too.

They're rather functionally streamlined as it currently stands, so what would "cleaner" ideally entail?

rbuch:
> > I would update the constructors of `CkPointer` to be a bit cleaner here, too.
> 
> They're rather functionally streamlined as it currently stands, so what would "cleaner" ideally entail?

Make the default constructor use `nullptr`, perhaps use delegated constructors so that there's a central place for the logic to go (although there basically isn't any logic here, so perhaps not necessary).

stwhite91 (approved):

stwhite91 (approved):

evan-charmworks (approved):

PR #3230, opened 2021-01-05
Overhaul FutureState Using STL Containers

Wingpad:
This is a quality of life improvement for `CkFuture` which replaces the non-STL, internally managed previous future pool with one based on STL objects. This more robust interface will be used to implement `ck::wait_any` and `ck::wait_all`. Additionally, it can enable continuations in a fairly seamless manner.


src/ck-core/ckfutures.C
```diff
@@ -20,20 +20,100 @@ in remote process control.
...
+  }
+};
+
+struct FutureState {
+  using request_t = std::shared_ptr<FutureRequest>;
+  using request_queue_t = std::vector<request_t>;
+ private:
+  std::unordered_map<CkFutureID, void*> values;
+  std::unordered_map<CkFutureID, request_queue_t> waiting;
+  std::set<CkFutureID> freeList;
```
rbuch, line 58: I might be more efficient to use a `std::vector`, `std::queue` or `std::stack` here. Do you need the semantics of a set for something?

src/ck-core/ckfutures.C
```diff
@@ -115,45 +195,12 @@ class CkSemaPool {
...
-
-  for (i=lo; i<hi; i++)
-    array[i].next = i+1;
-  array[hi-1].next = fs->freelist;
-  fs->freelist = lo;
-}
-
 static 
 inline
 int createFuture(void)
```
rbuch, line 200: This should be returning a `CkFutureID`.

src/ck-core/ckfutures.C
Wingpad: A set was chosen to preclude issues with repeated frees of the same future. If this is something we would rather specify as undefined behavior, a queue (or any of the others could be chosen).

src/ck-core/ckfutures.C
```diff
@@ -20,20 +20,100 @@ in remote process control.
...
+
+struct FutureState {
+  using request_t = std::shared_ptr<FutureRequest>;
+  using request_queue_t = std::vector<request_t>;
+ private:
+  std::unordered_map<CkFutureID, void*> values;
+  std::unordered_map<CkFutureID, request_queue_t> waiting;
+  std::set<CkFutureID> freeList;
+
+  CkFutureID last;
```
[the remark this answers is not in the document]
Wingpad: No, unique within a PE. A future's globally unique ID is the pair (homePe, id)

src/ck-core/ckfutures.C
rbuch: I see. I think it's fine to disallow repeated frees (the old semantics didn't allow it), but I'll leave the choice up to you.

Wingpad: I am leaning towards changing it to a vector and adding `CkAssert(!(freeList contains id) && "repeated frees of the same future");` so if error-checking is enabled it'll notify the user about repeated frees and, otherwise, will not have have any performance impact.

stwhite91 (approved):

rbuch (approved):

PR #3231, opened 2021-01-05
Extend Future Capabilities with Wait All/Some/Any

Wingpad:
These are mechanisms by wait to wait on multiple futures simultaneously without introducing a busy-loop in the user's code. Most critically, perhaps, this represents a steps towards bringing our futures on par with those of other runtimes. The ability to retrieve the first value available among multiple futures can be used as a mechanism for implementing many higher-level constructs (e.g. something analogous to SDAG's overlap).


src/ck-core/ckfutures.h
```diff
@@ -99,7 +107,38 @@ namespace ck {
...
+    inline bool operator==(const future<T>& other) const {
+      return (this->handle_.id == other.handle_.id) &&
+             (this->handle_.pe == other.handle_.pe);
+    }
   };
+
+  // returns a pair with the value and fulfilled future
+  template<typename InputIter, typename T = typename InputIter::value_type::value_type>
+  std::pair<T, InputIter> wait_any(InputIter first, InputIter last) {
+    std::vector<CkFutureID> handles(last - first);
```
rbuch, line 120: The construction and filling of handles should be uniform here and in `wait_all`, both should reserve their space and use the same arguments to `transform`.

src/ck-core/ckfutures.h
```diff
@@ -99,7 +107,38 @@ namespace ck {
...
+    std::transform(first, last, handles.begin(),
+      [](future<T>& f) { return f.handle().id; });
+    auto pair = CkWaitAnyID(handles);
+    auto value = future<T>::unmarshall_value((CkMarshallMsg*)pair.first);
```
rbuch, line 124:
```suggestion
    auto value = future<T>::unmarshall_value(static_cast<CkMarshallMsg*>(pair.first));
```

rbuch: Replace C-style casts in other places, too.

src/ck-core/ckfutures.h
```diff
@@ -58,25 +58,33 @@ void _futuresModuleInit(void);
 #ifdef __cplusplus
 }
 
+std::vector<void*> CkWaitAllIDs(const std::vector<CkFutureID>& handles);
+std::pair<void*,CkFutureID> CkWaitAnyID(const std::vector<CkFutureID>& handles);
```
rbuch, line 62:
```suggestion
std::pair<void*, CkFutureID> CkWaitAnyID(const std::vector<CkFutureID>& handles);
```

src/ck-core/ckfutures.C
```diff
@@ -231,9 +275,10 @@ static void setFuture(CkFutureID handle, void *pointer)
   if (pointer==NULL) CkAbort("setFuture called with NULL!");
 #endif
   fut->value = pointer;
-  for (t=fut->waiters; t; t=CthGetNext(t))
+  for (auto t: fut->waiters) {
```
rbuch, line 278:
```suggestion
  for (auto t : fut->waiters) {
```

Wingpad:
This is dependent on PR#3230, which should be merged first; this PR will be rebased onto master once that has occurred. (Is there a better way to express this relationship?)

src/ck-core/ckfutures.C
```diff
@@ -201,6 +200,51 @@ void *CkWaitFutureID(CkFutureID handle)
...
+    if (!future->ready) {
+      future->waiters.push_back(self);
+    }
+  }
+  do {
+    for (int i = 0; i < n; i++) {
+      // when a future is ready
+      if (futures[i]->ready) {
+        // drop ourself from all the futures
+        for (auto future: futures) Future_dropWaiter(future, self);
```
rbuch, line 228:
```suggestion
        for (auto future : futures) Future_dropWaiter(future, self);
```
Perhaps also make it `const auto& future`.

src/ck-core/ckfutures.C
```diff
@@ -231,9 +275,10 @@ static void setFuture(CkFutureID handle, void *pointer)
   if (pointer==NULL) CkAbort("setFuture called with NULL!");
 #endif
   fut->value = pointer;
-  for (t=fut->waiters; t; t=CthGetNext(t))
+  for (auto t: fut->waiters) {
     CthAwaken(t);
-  fut->waiters = 0;
+  }
+  fut->waiters = {};
```
rbuch, line 281:
```suggestion
  fut->waiters.clear();
```

rbuch:
> This is dependent on [PR#3230](https://github.com/UIUC-PPL/charm/pull/3230), which should be merged first; this PR will be rebased onto master once that has occurred. (Is there a better way to express this relationship?)

What you're doing is right, GitHub will automatically change the target branch to `master` once the branch this depends on is merged.

src/ck-core/ckfutures.C
Wingpad: It would be physically const, but not logically. Also, since it is a pointer and passed-by-value anyway, is there any benefit to `auto&` over `auto`?

src/ck-core/ckfutures.C
```diff
@@ -201,6 +200,51 @@ void *CkWaitFutureID(CkFutureID handle)
...
+      // when a future is ready
+      if (futures[i]->ready) {
+        // drop ourself from all the futures
+        for (auto future: futures) Future_dropWaiter(future, self);
+        // then return the value/handle pair
+        return std::make_pair(futures[i]->value, handles[i]);
+      }
+    }
+    // suspend until a future becomes ready...
+    CthSuspend();
```
Wingpad, line 234: For some reason, in multi-process configurations, this fails unless `CthSuspend` comes before `future...->value`, even when all the futures are ready (which should be invalid, so things are flipped). Any thoughts on why this may be occurring?

src/ck-core/ckfutures.C
```diff
@@ -242,6 +242,31 @@ void *CkWaitFutureID(CkFutureID handle)
...
+      }
+    }
+  } while (true);
+}
+
+std::vector<void*> CkWaitAllIDs(const std::vector<CkFutureID>& ids) {
+  // a more sophisticated solution should be implemented that interleaves
+  // waiting on multiple futures... but this works in the meantime
+  std::vector<void*> results;
+  for (auto id: ids) {
```
rbuch, line 264:
```suggestion
  for (auto id : ids) {
```

src/ck-core/ckfutures.C
```diff
@@ -242,6 +242,31 @@ void *CkWaitFutureID(CkFutureID handle)
   return value;
 }
 
+std::pair<void*, CkFutureID> CkWaitAnyID(const std::vector<CkFutureID>& handles) {
+  auto fs = &(CpvAccess(futurestate));
+  auto ready = handles.end() != std::find_if(handles.begin(), handles.end(),
```
Wingpad, line 247:
```suggestion
  const auto ready = std::any_of(handles.begin(), handles.end(),
```

stwhite91:
MPI has a "Waitsome" which returns however many requests have already completed. That might be useful for futures users as well.

Wingpad:
> MPI has a "Waitsome" which returns however many requests have already completed. That might be useful for futures users as well.
That would be fairly straightforward to add. I can have a preliminary implementation by this evening, and include it in this PR or separately.

NK-Nikunj (approved):
LGTM. Thanks!

examples/charm++/future/fib.C
```diff
@@ -1,40 +1,79 @@
...
+        for (const auto& value : pair.first) {
+          sum += value;
+        }
+        // in a real program, we could go off and do something else...
+        // but we'll just wait on whatever's still outstanding
+        for (auto it = pair.second; it < pending.end(); it++) {
+          sum += it->get();
+        }
+      } else {
+        // odds use wait all
```
rbuch, line 59: Nitpicking, but I'd change this comment to make it clear that this branch applies to odds that are not divisible by three.

stwhite91 (commented):
Can you add a section to the futures part fo the manual on these? Otherwise looks good

Wingpad:
> Can you add a section to the futures part fo the manual on these? Otherwise looks good

Added a little table and pointer to the examples.

examples/charm++/future/fib.C
```diff
@@ -1,40 +1,79 @@
...
+        }
+      } else {
+        // odd n's (that are not divisible by three) use wait all
+        auto values = ck::wait_all(pending.begin(), pending.end());
+        for (const auto& value : values) {
+          sum += value;
+        }
+      }
+      // release any remaining futures
+      for (auto& f : pending) {
```
rbuch, line 66: I assume this is where the release happens for the odd cases? If so, I think you should make it clear that they all need to be released, right now it's odd that the even case is released in that block but the other two cases are not.

src/ck-core/ckfutures.C
```diff
@@ -262,6 +262,31 @@ void *CkWaitFutureID(CkFutureID handle)
   return value;
 }
 
+std::pair<void*, CkFutureID> CkWaitAnyID(const std::vector<CkFutureID>& handles) {
+  auto fs = &(CpvAccess(futurestate));
+  const auto ready = std::any_of(handles.begin(), handles.end(),
+    [&fs](const CkFutureID& id) { return fs->is_ready(id); });
+  if (!ready) fs->request(handles, std::make_shared<FutureToThread>(CthSelf()));
+  do {
+    if (!ready) CthSuspend();
```
rbuch, line 271: What is responsible for resuming this thread?

examples/charm++/future/fib.C
Wingpad: Yes, the remaining futures from the odd cases are released here. I attempted to allude to that with "remaining" futures, but I will make that clearer.

src/ck-core/ckfutures.C
Wingpad: The FutureBOC object wakes it up when it fulfills the `FutureToThread` request.

doc/charm++/manual.rst
```diff
@@ -5863,9 +5863,25 @@ is ``unsigned short``, limiting each PE to 65,535 outstanding futures.
 To increase this limit, build Charm++ with a larger *CMK_REFNUM_TYPE*, e.g. specifying
 ``--with-refnum-type=uint`` to use ``unsigned int`` when building Charm++.
 
-
-The Converse version of future functions can be found in the :ref:`conv-futures`
-section.
+There are additional facilities for operating on collections of futures, which include:
+
++-------------------+-----------+-------------------------------------------------+
```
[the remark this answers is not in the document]
Wingpad: Two quick questions. This will increase the character-width of the table, is it OK if I exceed the 80 character (de jure) limit? Also, they include template type arguments. Do you think it would be OK to list them as such (without context),  `std::pair<T, InputIter>`, `std::vector<T>`, and `std::pair<std::vector<T>, InputIter>`,  or should I also include descriptions of `T` and `InputIter` here?

examples/charm++/future/fib.C
```diff
@@ -62,7 +62,12 @@ struct Fib : public CBase_Fib {
           sum += value;
         }
       }
-      // release any remaining futures
+      // release any futures still in pending
```
rbuch, line 65:
```suggestion
      // release any still pending futures
```

examples/charm++/future/fib.C
```diff
@@ -62,7 +62,12 @@ struct Fib : public CBase_Fib {
...
+      // ps#1, if futures aren't removed, e.g. with ``wait_all``,
+      //       this is all of them.)
```
rbuch, line 67:
```suggestion
      // ps#1, if futures aren't removed, e.g. with ``wait_all``,
      //       this is all of them.
```

examples/charm++/future/fib.C
```diff
@@ -62,7 +62,12 @@ struct Fib : public CBase_Fib {
...
+      // ps#2, _all futures must be released_ or memory will be
+      //       leaked and the future id will be unavailable for
+      //       reuse (implementation details)
```
rbuch, line 70: I'd remove `(implementation details)`, I don't think it really adds valuable information here.

doc/charm++/manual.rst
```diff
@@ -5865,20 +5865,20 @@ To increase this limit, build Charm++ with a larger *CMK_REFNUM_TYPE*, e.g. spec
...
-+-------------------+-----------+-------------------------------------------------+
++-------------------+-----------+---------------------------------------------------+
+| Function          | Blocking? | Description                                       |
++===================+===========+===================================================+
+| ``ck::wait_any``  | Yes       | Take a value, as soon as one is available, and    |
+|                   |           | return a pair with the value and position of the  |
+|                   |           | fulfilled future. (``std::pair<T, InputIter>``)   |
++-------------------+-----------+---------------------------------------------------+
+| ``ck::wait_all``  | Yes       | Wait for all the futures to become available, and |
+|                   |           | return a vector of values. (``std::vector<T>``)   |
```
rbuch, line 5876: Does `wait_all` also release the corresponding futures (as alluded to in your ps#1 comment)? If so, that should be clearly documented here.

Wingpad: No, wait all does not release any futures.

src/ck-core/ckfutures.C
Wingpad: Note, a `FutureToThread` request is also used on ckfutures.h:251 as well -- so it may be better to put some design notes about how future requests near their definition. I will put a back-reference to those sections here.

Wingpad: Comment added (see updates)

examples/charm++/future/fib.C
```diff
@@ -1,40 +1,87 @@
...
+          // wait for any of the futures in pending to become ready
+          auto pair = ck::wait_any(pending.begin(), pending.end());
+          // add the received value to the sum
+          sum += pair.first;
+          // then release and erase the fulfilled future
+          pair.second->release();
+          pending.erase(pair.second);
+        }
+      } else if (n % 3 == 0) {
+        // n's that are divsible by three (but not two) use wait_some
```
rbuch, line 48:
```suggestion
        // n's that are divisible by three (but not two) use wait_some
```

src/ck-core/ckfutures.C
```diff
@@ -262,6 +267,36 @@ void *CkWaitFutureID(CkFutureID handle)
...
+    for (const auto& handle : handles) {
+      if (fs->is_ready(handle)) {
+        return std::make_pair((*fs)[handle], handle);
+      }
+    }
+  } while (true);
+}
+
+std::vector<void*> CkWaitAllIDs(const std::vector<CkFutureID>& ids) {
+  // a more sophisticated solution should be implemented that interleaves
```
stwhite91, line 291: This seems worth doing now, and doesn't seem like it would be difficult to implement if I understand correctly. Separately, it would be better to initialize the vector to size and then assign rather than repeatedly calling push_back.

Wingpad: Perhaps this comment should be deleted, or hedged somewhat, since the more sophisticated approach is unlikely to make a significant difference at this time. As `CkWaitFuture` need not be called for a value to arrive, both approaches are constrained by the wait time for the last future to arrive. The more sophisticated approach could, potentially, have fewer fewer invocations of CthSuspend/CthAwaken... but it's unclear to me whether that would be worthwhile or not, as the longest-wait-time is invariant and the dominant performance factor.

epmikida (approved):
LGTM. Moving forward, is CkFutureID something we'd want to convert to a 64bit ID?

Wingpad:
> LGTM. Moving forward, is CkFutureID something we'd want to convert to a 64bit ID?

Eventually, yes, it's not difficult to exceed its numerical limits at this current time. However, it currently maps to a refnum so enabling a 64-bit ID requires some refactoring of the FutureBOC chare-group.

stwhite91 (approved):

PR #3237, opened 2021-01-07
Tracing: Mark phase boundary when PE reaches AtSync barrier

rbuch:


stwhite91 (approved):
Does this patch require corresponding changes in Projections?

rbuch:
> Does this patch require corresponding changes in Projections?

It's not currently used in Projections. It was added to support https://github.com/UIUC-PPL/charm/pull/3238, to measure communication between `AtSync`s. It should also be added to `trace-projections` and Projections itself, but that's a separate change.

evan-charmworks (approved):

PR #3238, opened 2021-01-07
Tracing: Add commSummary tracemode

rbuch:
Still need to add documentation, but code can still be reviewed.

### DIFF
--- a/src/ck-perf/trace-commSummary.C
+++ b/src/ck-perf/trace-commSummary.C
@@ -74,38 +74,16 @@ void TraceCommSummary::traceClose(void)
   // remove myself from traceArray so that no tracing will be called.
   CkpvAccess(_traces)->removeTrace(this);
 
-  std::ofstream f(std::string(CkpvAccess(traceRoot)) + "." + std::to_string(myPe) + ".csumm");
-  f << "selfCount\n";
-  for (const auto& count : selfCount)
+  std::ofstream f(std::string(CkpvAccess(traceRoot)) + "." + std::to_string(myPe) +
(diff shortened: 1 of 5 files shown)


src/ck-perf/trace-commSummary.C
```diff
@@ -0,0 +1,101 @@
...
+  CkpvAccess(_trace) = new TraceCommSummary(argv);
+  CkpvAccess(_traces)->addTrace(CkpvAccess(_trace));
+}
+
+TraceCommSummary::TraceCommSummary(char** argv) : myPe(CkMyPe()), myNode(CkMyNode())
+{
+  if (CkpvAccess(traceOnPe) == 0) return;
+}
+
+void TraceCommSummary::beginExecute(CmiObjId* tid) { beginExecute(-1, -1, -1, myPe, 0); }
```
matthiasdiener, line 24: Would it be possible to mark these `inline`?

rbuch: I assumed any competent compiler would take care of it. Do you think it would be beneficial?

matthiasdiener: I think it is only guaranteed when member functions are defined in a class header, not in a cpp file: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inline

rbuch: I'm not sure how much good it'll do here since they're invoked virtually via a `Trace*`, but I moved the definitions of the functions that just wrap the detailed `beginExecute` into the header.

PR #3213, opened 2020-12-22
LB: Build all available LBs by default

rbuch:
Not doing this by default probably causes more confusion among users than is worthwhile and simplifies the process of building and using load balancers.

Issue #3139, opened 2020-10-27
Suggestions for documentation clarifications


fmahebert:
Over the past few months, @moxcodes and I have been working to add load balancing and checkpoint/restart in SpECTRE.

Here's a list of some things we've encountered in the Charm++ documentation that we felt could be clarified, as well as some questions that we thought would be useful to have addressed by the documentation. This list is a bit lengthy, so we've done our best to organize it by section and break it up into smaller items that can be checked off.

We're happy to discuss any items where we haven't made ourselves/our questions clear!

Based on Charm++ v6.11.0-beta1, and documentation from https://charm.readthedocs.io/en/latest/index.html as of late Oct 2020.

---

2.2.6 - Load Balancing

2.2.6.1

- [x] First sentence: our understanding is that singletons cannot migrate either, in which case the statement should be restricted to array chares only.

- [x] Are LB measurements done for every charm application, or only in cases where the user "turns on" LB by linking in a LB module or calling charm functions or using command-line args? If one of the latter, which?

- [x] Can the documentation state (here or in a different section, if more appropriate) when LB measurements start? Do measurements start as soon as the application starts up, or later? In other words, will startup costs enter into the LB decision on the first LB invocation? Does the answer depend on whether periodic, at sync, or manual LB is used?

- [x] Does the load balancing database get reset after a LB invocation? Does this depend on whether any chares are actually migrated by the LB?

2.2.6.2

- [x] Since v6.11 will be gathering several LBs under TreeLB and eliminating other LBs, it would help discoverability and usability if the documentation were updated to provide an accurate listing of available LBs. Specific concerns:

  1) [x] [Already in #3131] Given that many LBs are implemented within TreeLB, it would be nice to improve their discoverability. Ideally these sub-balancers would once again appear in the output from running `exec +balancer help`, but providing a detailed listing of the sub-balancers implemented by TreeLB in the main documentation would also be very helpful. This information would be useful not just for running in "legacy" mode, but also for setting up the TreeLB config file.

  2) [x] Relatedly, since TreeLB exposes the ability to configure the individual LB strategies, it would be helpful to document the parameters for each of the strategies. Ideally, these would be listed in the documentation but also in the command-line help text so that if the documentation isn't updated, at least the command line utility can report details of the current branch of the code.

  3) [x] The documentation lists some LBs that appear to have been removed, e.g., TopoCentLB. We suggest removing from this listing any LBs that are not actually available.

  4) [x] The documentation uses phrases like
[...]

PR #3243, opened 2021-01-13
Doc #3139: Clarify LB, checkpoint docs

rbuch:
Clarifies almost all of #3139:

- The Metabalancer question is too far out of my wheelhouse, I'll punt that to @kavithachandrasekar (#3244).
- The manual mode questions were skipped because we're removing manual mode; AtSync mode should be able to replace it without any loss of functionality.

Note that this depends on several of the outstanding LB PRs (#3241, #3216, #3213).


rbuch:
Feedback from @nilsdeppe @moxcodes @fmahebert would be appreciated here.

nilsdeppe:
I'm happy to look, but might not be able to make time until later next week, if that's okay

doc/charm++/manual.rst
```diff
@@ -2894,96 +2922,85 @@ elements migratable (see migration section above) and choose a load
...
-   to the automatic load balancing mode.
+   until load balancing finishes. The user calls ``AtSync()`` at the
+   end of some iteration, then, when all elements reach that call,
+   load balancing is triggered. The objects can start executing again
+   when ``ResumeFromSync()`` is called. In this case, the user
+   redefines ``ResumeFromSync()`` to trigger the next iteration of the
+   application. This patter effectively results in a barrier at load
+   balancing time (see example here :numref:`lbexample`).
+
+#. **Periodic load balancing mode**: This mode uses a timer to perform
```
moxcodes, line 2989: [nitpick] the ordering of the overview and API details are reversed (above you introduce periodic before at-sync, and here you explain at-sync before periodic), so the guide might be nicer if the same ordering appeared for both.

doc/charm++/manual.rst
```diff
@@ -2894,96 +2922,85 @@ elements migratable (see migration section above) and choose a load
...
-   background.
-
-   *TurnManualLBOff()* turns off manual load balancing and switches back
-   to the automatic load balancing mode.
+   until load balancing finishes. The user calls ``AtSync()`` at the
+   end of some iteration, then, when all elements reach that call,
+   load balancing is triggered. The objects can start executing again
+   when ``ResumeFromSync()`` is called. In this case, the user
+   redefines ``ResumeFromSync()`` to trigger the next iteration of the
+   application. This patter effectively results in a barrier at load
```
moxcodes, line 2986: `patter`->`pattern`

doc/charm++/manual.rst
```diff
@@ -8280,15 +8281,31 @@ timestep. To save resources, the PUP routine can well omit the 8
...
+explicitly handle them.
+
+Checkpointing variables of type ``CProxy_*`` is completely valid.
+After the restart callback is invoked, the original chare structures
+will have been reconstructed and all proxy variables will be valid
+references to the restarted versions of whatever they originally
+referred to. The only caveat to this is when the application is
+restarted on a different number of processors than it was checkpointed
+on, in which case non-location invariant chares have special
+behavior. Singleton chares are not created or restored at all and
```
moxcodes, line 8305: Woah! This is surprising -- thank you for adding this note!

moxcodes (commented):
Thank you @rbuch for adding this documentation, it is much clearer! I have a just few small comments.
Do I understand correctly from the removal of any mention of `manual` load-balancing that it will simply not be supported?

doc/charm++/manual.rst
```diff
@@ -8342,7 +8370,9 @@ Section :numref:`loadbalancing`) after restart is suggested.
 If restart is not done on the same number of processors, the
 processor-specific data in a group/nodegroup branch cannot (and usually
 should not) be restored individually. A copy from processor 0 will be
-propagated to all the processors.
+propagated to all the processors. Additionally, singleton chares will
+not be restored at all in this case, so they must be specially handled
+when restarting on a different number of processors.
```
fmahebert, line 8375: Thanks for noting this behavior! Sanity check on my understanding: I interpret this to mean that some unpacking pup (probably some mainchare) needs to call something like `if (new number of procs) { my_singleton_proxy_ = CProxy_MySingleton::ckNew(args) }` .. is this about right? Adding a note to this effect (or fixing my understanding if needed) as an example might be helpful to close out this paragraph.

fmahebert (commented):
These are really useful clarifications @rbuch — thank you!

Below are a few further questions to check my understanding, and a few minor phrasing things that I noticed while reading through this.

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,196 @@ Examples are in ``examples/charm++/load_balancing`` and
...
+  above link in order to use this load balancer. Because of this
+  dependence, ScotchLB is not built by default; it can be built by
+  running ``make ScotchLB`` in the Charm++ build folder
+  (e.g. ``netlrts-linux-x86_64-smp/``). If SCOTCH is installed in a
+  non-standard location, use the *-incdir* and *-libdir* build time
+  options to point to the include and library directories used,
+  respectively.
+
+In distributed approaches, load data is only exchanged among
+neighboring processors. There is no global synchronization. However,
```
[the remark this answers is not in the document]
rbuch: Ah, yes, it is perhaps a bit misleading. Global synchronization _with the application_ isn't needed for the centralized load balancers; once any object calls `AtSync()`, it can immediately continue executing, the load balancer will run concurrently whenever it is ready.

What this is trying to say is that the internals of the distributed load balancer itself does not make any globally synchronous calls, while TreeLB does (although it only does so on LB iterations that go up to the root level of the tree, in general it only requires synchronization up to the root of the largest subtree used for LB on that iteration).

Relaxing your global synchronization may buy you performance, potentially a significant amount. The main issues to consider are:

- PUP costs may be higher since the PUP routine can now run anytime, so it may have to transfer transient state. Additionally, it may put more strain on the network since regular application messages may also be in flight. In particularly high traffic scenarios, object migration can have the knock-on effect of creating additional messaging due to location manager misdelivering messages to stale locations and having to forward them.  
- The load balancing statistics may be less accurate. Stats are taken from the LB database when the LB actually starts running, not when `AtSync()` is called. Thus, if that point is not colocated with a global synchronization, different objects may have completed different amounts of work (and perhaps even be on different iterations, etc.), so the load reported can be noisy.
- In return for these costs, you avoid the idle time of waiting for LB to complete, potentially a big savings for large jobs, particularly expensive LB algorithms, and very imbalanced workloads.

Also, a minor point, but for the sake of completeness, calling an LB before the previous call has finished is undefined behavior as far as I know, so some synchronization at some point is technically needed, although in practice one could almost certainly completely forego synchronizing these calls with no ill effect for realistic applications.

doc/charm++/manual.rst
rbuch: Good catch, I'll fix that.

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,196 @@ Examples are in ``examples/charm++/load_balancing`` and
...
+processor at runtime. Object migration can potentially improve the
+performance of the parallel program by migrating objects from
+overloaded processors to underloaded ones.
+
+Charm++ implements a generic, measurement-based load balancing
+framework which automatically instruments all Charm++ objects,
+collecting computational load and communication structure during
+execution and storing them into a load balancing database (this only
+happens when a load balancer is linked into the application; if no LB
+is used, this measurement does not happen). This instrumentation
```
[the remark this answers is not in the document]
rbuch: Good idea, I've changed the nomenclature to "is activated" and linked to the compiler/runtime options section for more information.

doc/charm++/manual.rst
```diff
@@ -2894,96 +2922,85 @@ elements migratable (see migration section above) and choose a load
...
+has been called by all relevant objects in order to prevent race
+conditions (the implementation is designed to be robust against these
+issues and will often be able to handle them, but we make no
+guarantees if these rules are not obeyed). If using dynamic insertion,
+please ensure that insertions and calls to ``AtSync()`` cannot be
+interleaved and that ``doneInserting()`` is called after insertions
+are complete and before any element calls ``AtSync()``. Insertions
+and/or deletions may begin again after load balancing is complete
+(i.e. ``ResumeFromSync()`` is called for an object on the given PE for
+insertions or for the object in question for deletions).
```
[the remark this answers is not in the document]
rbuch: Yes, exactly that. The criterion for deletions is actually stricter than the truth, really you should be in the clear so long as _any_ of the object on the PE of object to be deleted have had a `ResumeFromSync()` call. The trouble, of course, is that you don't know what PE an object is on when coming out of an LB invocation until that _exact_ object has been resumed, as it may have migrated during load balancing.

doc/charm++/manual.rst
rbuch: Indeed, something of that ilk would work. I've added a sentence to clarify this.

rbuch:
> Thank you @rbuch for adding this documentation, it is much clearer! I have a just few small comments.
> Do I understand correctly from the removal of any mention of `manual` load-balancing that it will simply not be supported?

Yes, the erstwhile manual load balancing mode will not be supported, mostly because it seemed to us that it wasn't offering any additional functionality compared to our bread and butter AtSync mode.

fmahebert:
Updates in new commit look good to me; thanks again @rbuch

moxcodes:
Agreed -- the new additions look good to me. Thanks for putting in all of these documentation improvements!

doc/charm++/manual.rst
```diff
@@ -2894,96 +2922,84 @@ elements migratable (see migration section above) and choose a load
...
+**AtSync mode** and **periodic load balancing mode**.  In *AtSync
+mode*, the application invokes the load balancer explicitly at an
+appropriate location (generally at a pre-existing synchronization
+boundary) to trigger load balancing by inserting a function call
+(``AtSync()``) in the application source code.  In *periodic load
+balancing mode*, a user specifies only how often load balancing is to
+occur, using the *+LBPeriod* runtime option to specify the time
+interval.
+
+.. note:: In AtSync mode, Applications that use dynamic insertion or
```
nilsdeppe, line 2937: Does this mean that when using periodic load balancing we can insert freely and not have to worry about anything? It sounds like it below. Maybe this note is better inside list element `1. AtsyncMode`?

doc/charm++/manual.rst
```diff
@@ -2894,96 +2922,84 @@ elements migratable (see migration section above) and choose a load
...
+   between consecutive LB invocations in seconds. For example,
+   *+LBPeriod 10.5* can be used to invoke load balancing roughly every
+   10.5 seconds. Additionally, no array element can have
+   ``usesAtSync`` set to true. In this mode, array elements may be
+   asked to migrate at any time, provided that they are not in the
+   middle of executing an entry method. Thus, the PUP routines for
+   array elements must migrate all data needed to reconstruct the
+   object at any point in its lifecycle (as opposed to AtSync mode,
+   where PUP routines for load balancing migration are only called
+   after AtSync() and before ResumeFromSync(), so they can make some
```
nilsdeppe, line 3001: It sounds like periodic load balancing works with dynamic insertion. Could you very verbosely state whether or not that's true in the documentation?

doc/charm++/manual.rst
```diff
@@ -3064,99 +3079,68 @@ to configure the load balancer, etc. These functions are:
...
+  - activate an LB: to actually ask the runtime to create an LB
+    strategy and use it for a given run. You can only activate load
+    balancers that have been already been linked in at compile
+    time. The special ``-balancer {balancer name}`` link time argument
+    both links a module and activates it at runtime by default.
 
 Below are the descriptions about the compiler and runtime options:
 
-#. **compile time options:**
+#. **Compile time options:**
```
nilsdeppe, line 3106: Maybe add `to charmc`?

doc/charm++/manual.rst
```diff
@@ -8280,15 +8259,31 @@ timestep. To save resources, the PUP routine can well omit the 8
...
+explicitly handle them.
+
+Checkpointing variables of type ``CProxy_*`` is completely valid.
+After the restart callback is invoked, the original chare structures
+will have been reconstructed and all proxy variables will be valid
+references to the restarted versions of whatever they originally
+referred to. The only caveat to this is when the application is
+restarted on a different number of processors than it was checkpointed
+on, in which case non-location invariant chares have special behavior:
+singleton chares are not created or restored at all and
```
nilsdeppe, line 8283: question: is `mainchare` a `singleton` in this context?

nilsdeppe (commented):
This is a big improvement. Thank you so much @rbuch! I have a few clarifying questions, that's all :)

doc/charm++/manual.rst
rbuch: Yes, the problem with dynamic insertion is one of of ascertaining whether all chares have called `AtSync`, so periodic completely sidesteps that. I agree that moving this into the AtSync section is better placement, I'll do that.

doc/charm++/manual.rst
rbuch: No, it's treated distinctly, I'll add a note to that effect.

doc/charm++/manual.rst
```diff
@@ -3001,6 +3001,11 @@ The detailed APIs of these two methods are described as follows:
    after AtSync() and before ResumeFromSync(), so they can make some
    assumptions about state).
 
+   .. note:: Dynamic insertion works with periodic load balancing with
+	     no issues. However, dynamic deletion does not, since
+	     deletion may occur while the load balancing strategy is
```
nilsdeppe, line 3006: Is there any way to support this from the user-end? As in, could I do `X` before/after deletion? I ask because with AMR we would want to delete the old object that we split into multiple.

rbuch: Currently, I think not. I've been trying some quick hacky things to get it to work, but to no avail, fixing this is going to require changes to the core location/migration code. The fundamental problem here is that deletions that occur while LB is running (which there is no way of preventing in periodic mode because LB and the application are completely decoupled) can cause the balancer to try to migrate an object that no longer exists, so it'll segfault if we try to send it anyway, or the send/receive counts won't match if we don't send it, causing a hang.

There are of course ways around this, namely empty messages for deleted objects, buffering object deletions while in LB, or something else of that ilk.

In general, we expect that "real applications" will use periodic load balancing mode only sparingly, so fixing this hasn't really been a priority for us. Would this be something that'd be particularly useful for you all?

nilsdeppe: :+1: I figured it would be rather difficult to deal with on the bookkeeping end. Thanks for explaining that! One user-facing improvement would be if there's a way to avoid a segfault and instead print an error message. I realize this might not be possible because of the amount of type erasure needed inside Charm++ and how the type erasure is done :)

At this point I wouldn't feel comfortable requesting periodic LB with deletion :) I think we first need to get all the LBs working and behaving as we expect for some interesting physics problems before we can make the call that the `AtSync` synchronization is a bottleneck. A possibly more useful/relevant feature would be to allow `AtSync` to occur with an explicit QD (or reduction) from the user. I could see us being able to take advantage of the sync by doing a reduction to compute a new optimal time step size or something else like that. However, even that I wouldn't say is worth putting any effort into right now

rbuch: Sorry, I wasn't clear, it currently does print an error like:
```
------------- Processor 3 Exiting: Called CmiAbort ------------
Reason: [3] LBDB::Migrate: Handle 13 no longer registered, range 0-25
```
I meant that it would cause a segfault if we added that feature without overhauling the migration code.

Okay, gotcha. Also, to be clear, do you meant that you'd like to trigger `AtSync` via QD or reduction? Doing so (on a single proxy) is pretty simple right now if you just create an entry method that just contains  `{ AtSync(); }`, and that can be passed as a broadcast callback to either QD or a reduction. Doing so across several proxies is a bit more complicated, but also possible via a similar mechanism. Is there something I'm missing about your request, or is it just a desire to get this functionality with more streamlined syntax?

nilsdeppe: Ah, okay, perfect!

Yep, you understood correctly. I didn't know that would work already. Thanks! I wouldn't worry about a more streamlined syntax or anything until we have figured out what actually works well in practice :)

rbuch: Great, makes sense. I think it'd be a nice little spoonful of syntactic sugar, so thanks for the suggestion.

Also, I've created #3270 to track the deletion of objects during LB.

minitu:
@rbuch Is there a way to see the updated doc on readthedocs?

matthiasdiener:
> @rbuch Is there a way to see the updated doc on readthedocs?

Rendered HTML is here: https://charm.readthedocs.io/en/ronak-lb-doc-clarifications/

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,196 @@ Examples are in ``examples/charm++/load_balancing`` and
...
+
+- **Greedy**: Uses a greedy algorithm that iterates over the objects
+  and assigns the heaviest remaining object to the least loaded
+  processor. (Old: ``+balancer GreedyLB``)
+
+- **GreedyRefine**: Uses a greedy algorithm that assigns the heaviest
+  remaining object to the least loaded processor when it is currently
+  assigned to a heavily loaded processor, otherwise leaves the object
+  on its current processor to limit migrations. It takes an optional
+  argument ``tolerance``, which specifies the tolerance it should allow
```
stwhite91, line 2633: Can you add a little more detail on this option? Is it like "+tolerance 1.1" to set it 10% higher?

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,196 @@ Examples are in ``examples/charm++/load_balancing`` and
...
 
-An example of a hierarchical strategy can be found in:
-
--  **HybridLB**: This calls GreedyLB at the lower level and RefineLB at
-   the root.
-
-Listed below are load balancers for debugging purposes:
+Custom strategies that are based on CentralLB or HybridBaseLB will
+continue to be supported for now, but support for these will likely be
+dropped in a future release.
```
stwhite91, line 2691: Should add that users should develop their own custom balancers on TreeLB from now on here.

doc/charm++/manual.rst
rbuch: Done.

doc/charm++/manual.rst
rbuch: Done.

minitu (approved):
Skimmed through it but didn't see any major issues.

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,199 @@ Examples are in ``examples/charm++/load_balancing`` and
...
-frequent load changes or due to size of data in the object being large,
-a strategy which favors migration minimization at the cost of balance
-(eg: RefineLB) is more suitable. DistributedLB and HybridLB are suitable
-for large number of nodes. Communication-aware load balancers like
-MetisLB and ScotchLB are suitable for communication intensive
-applications. RotateLB and RandCentLB are more useful for debugging
-object migrations. The compiler and runtime options are described in
+frequent load changes or due to size of data in the object being
+large, a strategy which favors migration minimization at the cost of
+balance (eg: RefineLB) is more suitable. DistributedLB is suitable for
```
epmikida, line 2709: Does Refine currently work? I don't ever remember hearing it was back to a good state. Last I checked it was horribly slow. Unusably so.

doc/charm++/manual.rst
```diff
@@ -2894,96 +2927,89 @@ elements migratable (see migration section above) and choose a load
...
+balancing mode*, a user specifies only how often load balancing is to
+occur, using the *+LBPeriod* runtime option to specify the time
+interval.
+
+The detailed APIs of these two methods are described as follows:
+
+#. **AtSync mode**: Using this method, elements can be migrated only
+   at certain points in the execution, when the application invokes
+   ``AtSync()``. In order to use AtSync mode, one should set
+   ``usesAtSync`` to true in the array element constructor. When an
```
epmikida, line 2947: The wording of this para is maybe a bit confused w.r.t. where usesAtSync should be set. Although I'm not exactly sure how to improve it. IMO it should state right away that usesAtSync should be set in the constructor of all elements which take part in the synchronization process. Those that do not set atsync will still be migrated though so their PUP should be able to handle migration at any time in their life cycle. Currently this is stated at the end of the para instead.

This also leaves out any description of AtSync and arrays which are bound together, which maybe is what it should do? Using both together has come up as a point of confusion though in both paratreet and Quinoa. Maybe it's best to discuss in the section about bound arrays though? Similarly there is no mention of how load is represented for bound objects (or any objects really, including those where setMigration is set to false). But this is probably more relevant for the section where we discuss how to write new strategies.

epmikida (commented):
Mostly looks good, just a few comments for discussion

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,199 @@ Examples are in ``examples/charm++/load_balancing`` and
...
+  (e.g. ``netlrts-linux-x86_64-smp/``). If SCOTCH is installed in a
+  non-standard location, use the *-incdir* and *-libdir* build time
+  options to point to the include and library directories used,
+  respectively.
+
+In distributed approaches, load data is only exchanged among
+neighboring processors and there is no messaging outside this local
+neighborhood. However, they will not, in general, provide an immediate
+restoration for load balance - the process is iterated until the
+desired load balance can be achieved.
```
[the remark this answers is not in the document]
rbuch: Agreed, rereading it, I think it's misleading and overspecfies what distributed strategies consist of (and it's wrong for the DistributedLB case). I've corrected, simplified, and generalized this description.

doc/charm++/manual.rst
rbuch: The new RefineLB implementations should be reasonable, the old CentralLB version was the one with a problem, I think. I'll test it again to make sure.

doc/charm++/manual.rst
rbuch: I discussed this with Eric offline a bit, and I've expanded the description of AtSync and participation a bit to hopefully make it clearer.

I think the discussions about bound arrays and about how load is represented are both outside the scope of this section.

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,199 @@ Examples are in ``examples/charm++/load_balancing`` and
...
-   into account.
+TreeLB and its pluggable strategies supersede the previous
+implementations of centralized and hierarchical load balancing. To use
+TreeLB, the user selects one of several trees. Each level corresponds
+to a different division of the overall execution (e.g. PE,
+process). Each level is configurable with a list of strategies,
+frequency, and other parameters. See :numref:`treeLb` below for more
+detail, along with configuration and execution instructions. The
+following strategies can be used with TreeLB (the old runtime
+selection syntax still works and is specified in parentheses):
```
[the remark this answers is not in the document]
rbuch: Fixed.

doc/charm++/manual.rst
```diff
@@ -2555,173 +2557,199 @@ Examples are in ``examples/charm++/load_balancing`` and
...
+
+- **Random**: Randomly assigns objects to processors. (Old: ``+balancer RandCentLB``)
+
+- **Rotate**: Moves objects to the next available PE every time it is
+  called. It is useful for debugging PUP routines and other migration
+  related bugs. (Old: ``+balancer RotateLB``)
+
+However, not all non-distributed load balancers are currently
+supported by TreeLB, namely the centralized communication-aware load
+balancers:
```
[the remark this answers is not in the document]
rbuch: Fixed.

epmikida (approved):

minitu (approved):

PR #3247, opened 2021-01-13
Remove AtSync's dependence on existence of LB

rbuch:
Previously, AtSync would only be called when at least one load balancer existed, this removes that restriction, allowing AtSync to be cleanly used for non-LB purposes.

A new class of receiver, a BeginReceiver, is added as part of this PR. BeginReceiver and EndReceiver are used for setup and cleanup tasks for regular Receivers, which is where things like LB or other useful work is done. Currently, BeginReceiver and EndReceiver are used in `CkLocMgr` to trigger `RegisteringObjects` and `DoneRegisteringObjects`.


rbuch:
Abandoned in favor of #3248.

PR #3248, opened 2021-01-13
Disentangle AtSync-LB-CkLocMgr dependencies

rbuch:
Previously, `CkLocMgr` was shoehorned into being an AtSync client and AtSync would only be called when at least one load balancer existed, this removes these restrictions, allowing AtSync to be cleanly used for non-LB purposes and creating less blurry boundaries between the various components.

New classes of receivers, BeginReceiver and EndReceiver, are added as part of this PR. BeginReceiver and EndReceiver are used for setup and cleanup tasks for regular Receivers, which is where things like LB or other useful work is done. Currently, BeginReceiver and EndReceiver are used in CkLocMgr to trigger RegisteringObjects and DoneRegisteringObjects.

Replaces #3247 and #3170, this is a combined version of both of those PRs.

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -1635,7 +1635,9 @@ void CkMigratable::AtSync(int waitForMigration)
 {
 	if (!usesAtSync)
 		CkAbort("You must set usesAtSync=true in your array element constructor to use AtSync!\n");
-	if(myRec->getLBMgr()->getNLoadBalancers() == 0) {
+	// Only actually call AtSync when a receiver exists, otherwise skip it and
+	// directly call ResumeFromSync
+	if (!myRec->getSyncBarrier()->hasReceivers()) {
 		ResumeFromSync();
 		return;
 	}
@@ -2087,8 +2089,8 @@ CkLocMgr::CkLocMgr(CkMigrateMessage* m)
 
 CkLocMgr::~CkLocMgr() {
 #if CMK_LBDB_ON
-  syncBarrier->RemoveClient(lbBarrierClient);
-  syncBarrier->RemoveReceiver(lbBarrierReceiver);
+  syncBarrier->RemoveBeginReceiver(lbBarrierBeginReceiver);
+  syncBarrier->RemoveEndReceiver(lbBarrierEndReceiver);
   lbmgr->UnregisterOM(myLBHandle);
 #endif
   map->unregisterArray(mapHandle);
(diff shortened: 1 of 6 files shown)


epmikida (commented):
Looks good to me other than my confusion about hasReceivers.

epmikida:
> Looks good to me other than my confusion about hasReceivers.

Resolved in an offline discussion with Ronak. In summary, the current hasReceivers is an optimization that works because right now the barrier is only used with LB. In the future to make the barrier more general purpose, that optimization wouldn't work. But this PR is still a step in the right direction, and doesn't claim to make the barrier fully general purpose yet anyway.s

epmikida (approved):

rbuch:
> > Looks good to me other than my confusion about hasReceivers.
> 
> Resolved in an offline discussion with Ronak. In summary, the current hasReceivers is an optimization that works because right now the barrier is only used with LB. In the future to make the barrier more general purpose, that optimization wouldn't work. But this PR is still a step in the right direction, and doesn't claim to make the barrier fully general purpose yet anyway.s

Yes, and just to add a bit, this is a generalization of the optimization it used to use, making it applicable for more than just load balancing. I wanted to err on the side of not changing the semantics of the call as much as possible, so I preserved the behavior of not really calling into the barrier when there are no receivers. This is predicated on the assumption that receivers are globally added in some sense (or, another way, that there are no PEs with no receivers when some PE has a receiver).

src/ck-core/cksyncbarrier.C
```diff
@@ -131,39 +164,33 @@ void CkSyncBarrier::propagate_atsync()
...
-    if (clients.front()->chare->isLocMgr()) return;
-  }
-
-  // If there are no clients, resume as soon as we're turned on
-  if (client_count == 0)
-  {
-    cur_epoch++;
-    CallReceivers();
-  }
+  if ((client_count == 0 && currentKick < cur_epoch) || startedAtSync) return;
```
kavithachandrasekar, line 193: It would be good to retain the previous comment 'If there are no clients, resume as soon as we're turned on'. Also does currentKick < cur_epoch ensure that code does not return for a future epoch in case client count !=0 in the future? It would be to add a comment on that as well.

src/ck-core/cksyncbarrier.C
```diff
@@ -43,8 +45,8 @@ LDBarrierClient CkSyncBarrier::AddClient(Chare* chare, std::function<void()> fn,
     // it can trigger. Do this asynchronously so that the caller functions for object
     // construction finish first.
     at_count += epoch - cur_epoch;
-    if (at_count >= clients.size())
-      thisProxy[thisIndex].CheckBarrier(false);
+    if (on && !startedAtSync && at_count >= clients.size() + 1)
```
kavithachandrasekar, line 48: Can the inserting the LBClient object be moved above this code block to avoid the '+1' in the condition, so that the check is similar to the code in RemoveClient() fn definition.

src/ck-core/cksyncbarrier.C
rbuch: That comment isn't strictly true anymore, now there's the additional stipulation that we must have also received a kick corresponding to the current epoch or beyond. In any case, I'll add more explanation.

`currentKick < cur_epoch` is used here to ensure that the barrier has received a kick corresponding to at least the current epoch. When there are no clients, this is the criterion that needs to be met in order to trigger the barrier. It has nothing to do with the client count in the future.

src/ck-core/cklocation.C
```diff
@@ -1633,16 +1633,20 @@ void CkMigratable::ckFinishConstruction(int epoch)
...
+        "AtSync!\n");
+  // Only actually call AtSync when a receiver exists, otherwise skip it and
+  // directly call ResumeFromSync
+  if (!myRec->getSyncBarrier()->hasReceivers()) {
+    ResumeFromSync();
+    return;
+  }
+  myRec->AsyncMigrate(!waitForMigration);
+  if (waitForMigration) ReadyMigrate(true);
+  ckFinishConstruction();
```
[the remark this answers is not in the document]
rbuch: Fixed.

src/ck-core/cksyncbarrier.C
rbuch: More explanation added in updated commits.

src/ck-core/cksyncbarrier.C
rbuch: Done in 04825392ec79b38cccd8f65d55b89fce0b4a2a9c.

kavithachandrasekar (approved):

epmikida (approved):

PR #3251, opened 2021-01-14
LB: Fix SetLBPeriod, GetLBPeriod

rbuch:


src/ck-ldb/LBManager.C
```diff
@@ -1116,10 +1116,7 @@ void LBChangePredictor(LBPredictorFunction* model)
 void LBSetPeriod(double second)
 {
 #if CMK_LBDB_ON
-  if (CkpvAccess(lbmanagerInited))
-    LBManager::Object()->SetLBPeriod(second);
-  else
-    _lb_args.lbperiod() = second;
+  _lb_args.lbperiod() = second;
```
kavithachandrasekar, line 1119: This still doesn't seem to call periodic LB. I believe the issue is that the code that checks lbperiod setTimer() in LBManager init() fn is executed before LBSetPeriod() is called by user code. Maybe making another setTimer() from inside LBSetPeriod() would fix the issue.

rbuch: Right, I think the behavior should be that periodic LB only happens when the user passes `+LBPeriod <period>` on the command line. In its current state, this PR has that behavior. It allows the user to change the period, but that only has any effect on LB frequency when they have passed the `+LBPeriod <period>` flag.

What do you think the behavior should be? Should the user be allowed to start periodic LB by just making a call to `SetLBPeriod` from within the program without having passed the requisite flag? If so, adding a call to `setTimer` here as you suggested should fix that. If not, then I suppose that we should maybe print out a warning when the user tries to use this when periodic LB mode isn't enabled.

kavithachandrasekar: It might be helpful to enable periodic LB when LBSetPeriod/SetLBPeriod, since user wouldn't need to pass the additional command line argument. Based on discussion with Ronak, this approach might be okay as long as usesAtsync is set to false.

rbuch: I've added support for this in the latest commits on this PR.

kavithachandrasekar (approved):

rbuch:
Retargeting this change on top of #3243 for cleaner staging and to ensure that the documentation changes are consistent with the behavior in each commit.

rbuch:
CI for ChaNGa will fail until https://github.com/N-BodyShop/changa/pull/89 is merged.

src/ck-ldb/LBManager.C
```diff
@@ -518,7 +527,7 @@ void LBManager::init(void)
   if (manualOn) TurnManualLBOn();
 #endif
 
-  if (_lb_args.lbperiod() != -1.0)  // check if user set LBPeriod - fix later
+  if (_lb_args.lbperiod() != -1.0)
```
epmikida, line 530: Would this be safer as `if (_lb_args.lbperiod() > 0.0)`?

epmikida (approved):

src/ck-ldb/LBManager.C
rbuch: The intention was just to see if they had changed it from the default, but I think your suggestion is better, we should make sure it's positive, too. I'll change it.

evan-charmworks (approved):

epmikida (approved):

PR #3256, opened 2021-01-18
AMPI: split doc into multiple files

matthiasdiener:
- makes searching more precise
- reduces 'wall of text'

This is how the docs look like with this patch: https://charm.readthedocs.io/en/ampi-split-doc/

If this looks alright, we could modify the other manuals too.


rbuch (approved):
I assume you've visually inspected the output and it looks the same as before?

matthiasdiener:
> I assume you've visually inspected the output and it looks the same as before?

Yeah, the docs themselves look the same, but are split into several files (ie, you need to click on a section in the table of contents to go to a new section, it's not scrolling through the whole manual on a single web page). You can try out the search at https://charm.readthedocs.io/en/ampi-split-doc/

rbuch:
> > I assume you've visually inspected the output and it looks the same as before?
> 
> Yeah, the docs themselves look the same, but are split into several files (ie, you need to click on a section in the table of contents to go to a new section, it's not scrolling through the whole manual on a single web page). You can try out the search at https://charm.readthedocs.io/en/ampi-split-doc/

Gotcha.

Is there a way to view them all in one page with the scheme? Sometimes I personally find it useful to have it all on one page so I can Ctrl+F for things on it rather than using RtD's search. It's not a deal breaker if not, I'm mostly just curious.

matthiasdiener:
> > > I assume you've visually inspected the output and it looks the same as before?
> > 
> > 
> > Yeah, the docs themselves look the same, but are split into several files (ie, you need to click on a section in the table of contents to go to a new section, it's not scrolling through the whole manual on a single web page). You can try out the search at https://charm.readthedocs.io/en/ampi-split-doc/
> 
> Gotcha.
> 
> Is there a way to view them all in one page with the scheme? Sometimes I personally find it useful to have it all on one page so I can Ctrl+F for things on it rather than using RtD's search. It's not a deal breaker if not, I'm mostly just curious.

No I think there is no way to do that :-( (apart from searching the pdf).

minitu:
RST apparently doesn't automatically do line breaks for inline literals? Some command lines in `https://charm.readthedocs.io/en/ampi-split-doc/ampi/05-examples.html` go out of the white block.

matthiasdiener:
> RST apparently doesn't automatically do line breaks for inline literals? Some command lines in `https://charm.readthedocs.io/en/ampi-split-doc/ampi/05-examples.html` go out of the white block.
> (I don't think this PR introduced it but would be good to fix anyways)

Yeah, we should fix this in a future patch.

minitu (approved):

Issue #3257, opened 2021-01-19
Commbench takes over 30 mins to run on Golub (with both ucx, verbs) 


nitbhat:
There is some weird performance issue on Golub which causes commbench to take over 30-35 mins to run. This is delaying ucx and verbs autobuild runs. (both smp and nonsmp). 

The same test takes about 5s or so to run on Golub and Frontera.

rbuch:
> There is some weird performance issue on Golub which causes commbench to take over 30-35 mins to run. This is delaying ucx and verbs autobuild runs. (both smp and nonsmp).
> 
> The same test takes about 5s or so to run on Golub and Frontera.

I assume you mean it takes about 5s or so to run on some other machine and Frontera, not Golub and Frontera, right?

nitbhat:
> > There is some weird performance issue on Golub which causes commbench to take over 30-35 mins to run. This is delaying ucx and verbs autobuild runs. (both smp and nonsmp).
> > The same test takes about 5s or so to run on Golub and Frontera.
> 
> I assume you mean it takes about 5s or so to run on some other machine and Frontera, not Golub and Frontera, right?

Sorry, that was typo. I meant to write that it takes 5s on Cori and Frontera. (and not Golub, where it takes about 30 mins).

nitbhat:
As suggested by @matthiasdiener, I dug into the autobuild logs and found that the commbench's run on Golub on Sept 16, 2020 took about 6s (http://charm.cs.illinois.edu/autobuild/old.2020_09_16__01_04/ucx-linux-x86_64.txt). 

Since then, I see that for a long period the tests haven't successfully run (and when they did they took about 30 odd mins). 

Currently, we're seeing some build issues on Golub because of the unavailability of verbs/ucx headers. I've reached out to the sysadmins to see what's causing them. Once they're fixed, I can check if there's a particular charm commit causing the slowdown. (or if it's being caused by some environment change on Golub that happened around that time frame)

matthiasdiener:
This is a test that took ~27 mins: http://charm.cs.illinois.edu/autobuild/old.2021_01_01__01_15/ucx-linux-x86_64.txt

A diff of the output for the commbench looks like this:

```patch
--- fast
+++ slow
@@ -1,82 +1,82 @@
-make[3]: Entering directory `/scratch/users/nbhat4/autobuild/ucx/charm/ucx-linux-x86_64/benchmarks/converse/commbench'
-../../../bin/testrun  ./pgm +p4  
[...]

PR #3258, opened 2021-01-19
Skip commbench on Golub because it times out autobuild run

nitbhat:
I've opened a separate issue to look into the performance issue on golub. (https://github.com/UIUC-PPL/charm/issues/3257).

### DIFF
--- a/benchmarks/converse/commbench/Makefile
+++ b/benchmarks/converse/commbench/Makefile
@@ -13,7 +13,7 @@ LOCALHOSTNAME:=$(shell hostname 2>/dev/null)
 
 test: commbench
 ifneq (,$(findstring golub,$(LOCALHOSTNAME)))
-	@echo "Skipping commbench on Golub because of performance bug"
+	@echo "Skipping commbench on Golub because of performance bug #3257"
 else
 	$(call run, ./commbench +p4 )
 endif
--- a/benchmarks/converse/commbench/Makefile
+++ b/benchmarks/converse/commbench/Makefile
@@ -19,7 +19,11 @@ else
 endif
 
 testp: commbench
+ifneq (,$(findstring golub,$(LOCALHOSTNAME)))
+	@echo "Skipping commbench on Golub because of performance bug"
+else
 	$(call run, ./commbench +p$(P))
+endif
 
 memoryAccess.o: memoryAccess.c
 	$(CHARMC) memoryAccess.c
--- a/benchmarks/converse/commbench/Makefile
+++ b/benchmarks/converse/commbench/Makefile
@@ -9,8 +9,14 @@ all: commbench
 commbench: $(OBJS)
 	$(CHARMC) -o commbench $(OBJS) -language converse++
 
+HOST=$(shell hostname)
+
 test: commbench
+ifneq (,$(findstring golub,$(HOST)))
+	@echo "Skipping commbench on Golub because of performance bug"
+else
 	$(call run, ./commbench +p4 )
+endif
 
 testp: commbench
 	$(call run, ./commbench +p$(P))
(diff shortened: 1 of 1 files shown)


benchmarks/converse/commbench/Makefile
```diff
@@ -9,8 +9,14 @@ all: commbench
 commbench: $(OBJS)
 	$(CHARMC) -o commbench $(OBJS) -language converse++
 
+HOST=$(shell hostname)
+
 test: commbench
+ifneq (,$(findstring golub,$(HOST)))
```
evan-charmworks, line 15: We should use the `:=` Makefile operator here so that the shell command is only called once at the time the variable is defined, otherwise it will be called each time the variable is used.

I also recall a prior experience where using a variable named HOST in a Makefile caused problems on certain systems where such a variable is defined in the environment, so I recommend naming it something more unique.

```suggestion
LOCALHOSTNAME:=$(shell hostname 2>/dev/null)

test: commbench
ifneq (,$(findstring golub,$(LOCALHOSTNAME)))
```

evan-charmworks:
Should the check apply to TESTP too?

rbuch (dismissed):
+1 Evan's suggestion.

nitbhat:
> Should the check apply to TESTP too?

Since we don't run the `testp` target (unlike the `test` target that is exercised during autobuild), I think we can skip that for now.

benchmarks/converse/commbench/Makefile
nitbhat: That makes sense, thanks @evan-charmworks.

evan-charmworks:
> > Should the check apply to TESTP too?
> 
> Since we don't run the `testp` target (unlike the `test` target that is exercised during autobuild), I think we can skip that for now.

Are you sure it wouldn't be good to add it as a proactive measure, in case we add testp to autobuild?

nitbhat:
> > > Should the check apply to TESTP too?
> > 
> > 
> > Since we don't run the `testp` target (unlike the `test` target that is exercised during autobuild), I think we can skip that for now.
> 
> Are you sure it wouldn't be good to add it as a proactive measure, in case we add testp to autobuild?

I see, it wouldn't hurt to add it. Let me do that right now.

evan-charmworks (approved):

benchmarks/converse/commbench/Makefile
```diff
@@ -9,11 +9,21 @@ all: commbench
 commbench: $(OBJS)
 	$(CHARMC) -o commbench $(OBJS) -language converse++
 
+LOCALHOSTNAME:=$(shell hostname 2>/dev/null)
+
 test: commbench
+ifneq (,$(findstring golub,$(LOCALHOSTNAME)))
+	@echo "Skipping commbench on Golub because of performance bug"
```
matthiasdiener, line 16:
```suggestion
	@echo "Skipping commbench on Golub because of performance bug #3257"
```

benchmarks/converse/commbench/Makefile
```diff
@@ -9,11 +9,21 @@ all: commbench
...
+else
 	$(call run, ./commbench +p4 )
+endif
 
 testp: commbench
+ifneq (,$(findstring golub,$(LOCALHOSTNAME)))
+	@echo "Skipping commbench on Golub because of performance bug"
```
matthiasdiener, line 23:
```suggestion
	@echo "Skipping commbench on Golub because of performance bug #3257"
```

matthiasdiener (approved):

evan-charmworks (approved):

nitbhat:
This PR is reverted by https://github.com/UIUC-PPL/charm/pull/3281 because the slowdown is fixed by changing the slurm job script to have `-n` argument passed to it.

PR #3264, opened 2021-01-23
CI: Update NAMD build to use new GitLab repo

rbuch:


minitu (approved):

matthiasdiener:
Out of curiosity, why does this need the token?

matthiasdiener (approved):

rbuch:
> Out of curiosity, why does this need the token?

While it's open source and all that, due to the license terms, the NAMD team requires a click through before being able to access the source code, so we need some authentication scheme to get access to the code.

PR #3266, opened 2021-01-26
CkLocMgr: Refactor PR1

epmikida:
This is the first PR of the overhaul of CkLocMgr/CkArray/CkMigratable. It takes the distributed hash table for ID -> PE and removes it from CkLocMgr and into a new, separate, CkLocCache (naming not final). A design doc for the overall redesign coming soon.


src/ck-core/cklocation.C
```diff
@@ -2908,51 +2928,72 @@ void CkLocMgr::sendMsg(CkArrayMessage* msg, CkArrayID mgr, const CkArrayIndex& i
   CmiUInt8 id;
   if (lookupID(idx, id))
   {
+    // We know the elements ID so we can go through normal delivery channels
```
rbuch, line 2931:
```suggestion
    // We know the element's ID so we can go through normal delivery channels
```

src/ck-core/cklocation.h
```diff
@@ -299,6 +299,47 @@ enum CkElementCreation_t : uint8_t
 #  define CMK_RANK_0(pe) pe
 #endif
 
+class CkLocCache : public CBase_CkLocCache
+{
+private:
+  // Map of ID to PE
+  typedef std::unordered_map<CmiUInt8, int> IdPeMap;
```
rbuch, line 306: Minor, but I think we should prefer `using` over `typedef`.

src/ck-core/cklocation.h
```diff
@@ -444,20 +494,22 @@ class CkLocMgr : public IrrGroup
   {
     return CMK_RANK_0(map->procNum(mapHandle, idx));
   }
-  inline bool isHome(const CkArrayIndex& idx) const
+  inline bool isHome(const CkArrayIndex& idx) const { return homePe(idx) == CkMyPe(); }
```
rbuch, line 497: These probably don't need `inline` unless something weird is going on with includes.

src/ck-core/cklocation.h
```diff
@@ -444,20 +494,22 @@ class CkLocMgr : public IrrGroup
...
-  int lastKnown(const CkArrayIndex& idx);
-  int lastKnown(CmiUInt8 id);
-
-  inline void insertID(const CkArrayIndex& idx, const CmiUInt8 id)
+  inline int lastKnown(const CmiUInt8 id) const { return cache->lastKnown(id); }
+  inline int lastKnown(const CkArrayIndex& idx) const
   {
-    if (compressor) return;
-    idx2id[idx] = id;
+    int pe = whichPE(idx);
```
rbuch, line 508: Why doesn't this use `cache->lastKnown(id)` a la `whichPE`?

src/ck-core/cklocation.C
```diff
@@ -2498,6 +2537,19 @@ void CkLocMgr::processAfterActiveRgetsCompleted(CmiUInt8 id)
   }
 }
 
+void CkLocMgr::deliverAllBufferedMsgs(CmiUInt8 id)
```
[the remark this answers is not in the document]
epmikida: Once messaging moves to CkArray this goes away/is significantly simplified. But the distinction comes from the reasons for buffering. Sometimes messages are buffered because we are still dealing with array indices, and some are buffered after we leave index-land and move to ID-only-town.

src/ck-core/cklocation.h
epmikida: For now I left it as is because homePe always has to convert from id to index, so since we already have the index, in the common case I wanted to avoid going from index to ID to index. Some of this will have to change anyways once we completely remove CkLocMgr from CkLocCache.

src/ck-core/ckmemcheckpoint.C
```diff
@@ -1178,7 +1178,7 @@ void CkMemCheckPT::recoverAll(CkArrayCheckPTMessage * msg,std::vector<CkGroupID>
 			p|idx;
                         p|id;
 			CkLocMgr * mgr = (CkLocMgr *)CkpvAccess(_groupTable)->find(gID).getObj();
-    			int homePe = mgr->homePe(idx);
+    	int homePe = mgr->homePe(idx);
```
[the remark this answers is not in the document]
epmikida: Whoops. Yup. I'll remove that.

epmikida:
Changed base branch to master and rebased, now that the precleanup patch has been merged.

evan-charmworks (approved):

src/ck-core/cklocation.C
```diff
@@ -2375,11 +2404,18 @@ CkLocMgr::CkLocMgr(CkArrayOptions opts)
...
-  if (map == NULL)
+  map = static_cast<CkArrayMap*>(CkLocalBranch(mapID));
+  if (map == nullptr)
     CkAbort("ERROR! Local branch of array map is NULL!");
+  // TODO: Should this be registered here, CkArray, CkLocMgr, or none?
   mapHandle = map->registerArray(opts.getEnd(), thisgroup);
 
+  cacheID = opts.getLocationCache();
+  cache = static_cast<CkLocCache*>(CkLocalBranch(cacheID));
+  if (cache == nullptr)
```
stwhite91, line 2415: Could make this and the other error checking in this function conditionally compiled on CMK_ERROR_CHECKING

stwhite91 (approved):

src/ck-core/cklocation.C
epmikida: Do you think that is worth doing for stuff like this that is only at creation of a new array? It's not a performance critical operation, and it's not like it's going to save overhead over a ton of very frequent calls.

stwhite91: Yeah, it's fine in that kind of non-performance-critical case

PR #3269, opened 2021-01-28
Replace PUPed raw arrays in src/ck-ldb/ with std::vector

evan-charmworks:
Salvaged from parts of #2108


evan-charmworks:
I don't understand what is causing this error.

```
Undefined symbols for architecture x86_64:
  "std::__1::vector<LDObjKey, std::__1::allocator<LDObjKey> >::~vector()", referenced from:
      LBDatabase::MulticastSend(LDOMHandle const&, unsigned long long*, int, unsigned int, int) in libck.a(LBDatabase.C.o)
ld: symbol(s) not found for architecture x86_64
```

evan-charmworks:
The issue is that when you put a class with non-trivial operators like std::vector inside a union (which wasn't valid at all before C++11), you have to call its destructor manually. Looking at how to do this, I found it confusing and difficult to maintain, and since this PR's goal is cleanup, I decided to revert the changes to LDCommDesc. (For reference, C++17 provides std::variant to handle situations like this.)

stwhite91 (approved):

minitu (approved):

PR #3272, opened 2021-02-02
Add Support for Using Futures as Callbacks

Wingpad:


src/ck-core/ckfutures.C
```diff
@@ -426,6 +426,23 @@ public:
   }
 };
 
+// TODO :
+// a poorly located helper function for getting a buffer from a message
+// should be moved elsewhere (when PR#3228 is merged), and probably named ck::get_message_buffer(void*)
+void* CkGetMsgBuffer(void* msg) {
+  auto env = UsrToEnv(msg);
```
rbuch, line 433: Should you so choose:
```suggestion
  auto* env = UsrToEnv(msg);
```

src/ck-core/ckfutures.C
```diff
@@ -426,6 +426,23 @@ public:
...
+  auto idx = env->getMsgIdx();
+  if (idx == CMessage_CkMarshallMsg::__idx) {
```
rbuch, line 435: Perhaps better as a `switch` statement?

Wingpad: Hmm, the `__idx` members are not compile-time constants so it does not seem like they can be used as the value for a `case` statement (`error: expression must have a constant value`).

rbuch: Ah, I see.

stwhite91:
Should this be documented?

src/ck-core/ckfutures.C
```diff
@@ -426,6 +426,23 @@ public:
   }
 };
 
+// TODO :
+// a poorly located helper function for getting a buffer from a message
+// should be moved elsewhere (when PR#3228 is merged), and probably named ck::get_message_buffer(void*)
+void* CkGetMsgBuffer(void* msg) {
```
evan-charmworks, line 432: Do you have any ideas for how this function could be more robust? Ideally it should work with any `message` defined in a .ci file.

evan-charmworks (commented):
Being able to use a future as a callback sounds very useful for enabling SDAG-like semantics for Charm APIs that use callbacks.

We have a use case for this PR in AMPI for creating new chare arrays asynchronously, as a replacement for the current use of `CkCallback::resumeThread` which we've recently found to be unsafe. The message type in question is `CkArrayCreatedMsg`, so I had to change `CkGetMsgBuffer` to `return msg;` instead of abort, but it seems to work otherwise. I don't know if that's something safe enough to merge though.

src/ck-core/ckfutures.C
Wingpad: Ah, excellent. Glad there's already a potential use-case in mind, I hope for it to be useful (especially in conjunction with wait all/any).

Hmm, good question. Perhaps we can discuss in core later. Handling user-defined message types is tricky. In my mind, the simplest path would be to make a specializable template class (a la `std::hash`) that users can specialize for their own message types. Since a user's message's internal structure can vary widely, I think it would be best for them to statically provide the offset somehow -- as we cannot virtualize messages (AFAIK).

src/ck-core/charm.h
```diff
@@ -426,7 +426,19 @@ extern void CkArrayExtSend_multi(int aid, int *idx, int ndims, int epIdx, int nu
 
 /*@}*/
 
+typedef CMK_REFNUM_TYPE CkFutureID;
+typedef struct _ckFuture {
+  CkFutureID id;
+  int        pe;
```
epmikida, line 432: This does seem pretty trivially convertible to 64bit ID if that's something we want to pursue, which is good. Current 64bit ID formate is [TAG | PE | Element ID], where TAG is a 3bit tag specifying the type of the ID (future in this case).

epmikida (dismissed):
Should there be a simpler/more dedicated test case for this as well? Something added to the future test (if there is one) or the callback test (if there is one)?

src/ck-core/charm.h
Wingpad: Do other examples come to mind where this type of abstraction would work well? This seemed like a natural place to experiment with it because it's doing multiple reductions -- but putting tests elsewhere would be fine as well. The futures example isn't really doing anything where this sort of callback mechanism makes sense (it's a Fibonacci program) but something that just hammers the futures API could be written as well, I suppose.

Indeed, it maps quite nicely to that formulation; however, do note my response on the other PR regarding its use as a refnum.

src/ck-core/ckfutures.C
Wingpad: Having thought about this for a bit and drawing a blank wrt a better solution, we may consider generating a list of "essential" internal message types and restrict `ck::future` to receiving messages of such types. Perhaps adding a disclaimer along the lines of:

"_In most cases, ``ck::future<T>`` cannot receive messages/values via ``get`` that have been not been sent/set by a corresponding call to ``set``. However, it can receive messages of some internal message types when sent via a call to `CkSendFuture`. Note, others may be wrapped as a PUP-able value and explicitly received as the expected type(s); for example, one might wrap a message as `CkMarshalledMsg`, sent via a `CkMarshallMsg`, then type-casted to the received message on the receiver-side_."

Wingpad: I updated the code to support array-creation messages and docs, can you check to see if these updates are satisfactory?

doc/charm++/manual.rst
```diff
@@ -5793,10 +5798,16 @@ futures, which include the following functions:
...
 
-You will note that the object-oriented versions are methods of `ck::future`,
-which can be templated with any pup'able type. An example of the
-object-oriented interface is available under `examples/charm++/future`,
-with an equivalent example for the C-compatible interface presented below:
+The object-oriented versions are methods of ``ck::future<T>``, which can be templated with any
+PUP-able type. Note, in most cases, messages/values cannot be retrieved via ``get`` when they were
+not been sent/set by a corresponding call to ``set``; however, it can receive messages of supported,
+internal message types sent via ``CkSendFuture``. Other message types must be wrapped as a PUP-able
+value and explicitly received as the expected message type(s); for example, one might wrap a message
```
stwhite91, line 5805: It would be nice to have an example or test program showing this

Wingpad: Which part of this, more specifically? This is getting into "expert-user" territory and we document how to prepare internal messages elsewhere so, to some extent, it is probably OK to omit some documentation wrt using `CkSendFuture` as such. As for using `CkMarshalledMsg`, that's likely best documented elsewhere/in a separate PR, if it isn't documented already.

evan-charmworks (dismissed):

epmikida (dismissed):

evan-charmworks:
This PR has a conflict that needs to be resolved before it can be merged.

evan-charmworks (dismissed):

src/ck-core/ckcallback.h
```diff
@@ -77,12 +78,16 @@ class CkCallback {
       case bcastArray: return "CkCallback::bcastArray";
       case bcastSection: return "CkCallback::bcastSection";
       case replyCCS: return "CkCallback::replyCCS";
+	  case sendFuture: return "CkCallback::sendFuture";
```
rbuch, line 81: Indentation doesn't match surrounding.

rbuch (approved):
LGTM except minor formatting stuff.

src/ck-core/ckcallback.h
evan-charmworks: I agree that we should fix the formatting.

evan-charmworks (approved):

stwhite91 (approved):

PR #3279, opened 2021-02-05
Revert "buildcmake: Use -cpmod to copy Fortran .mod files (#3159)"

evan-charmworks:
This reverts commit 3cd2c7b17bb4be260d9606ad65b6c9de2a6c3429, which did not actually work, and CI did not catch it.

### DIFF
--- a/cmake/converse.cmake
+++ b/cmake/converse.cmake
@@ -152,7 +152,11 @@ set(conv-util-cxx-sources
 
 if(CMK_CAN_LINK_FORTRAN)
     add_library(conv-utilf pup_f.f90)
-    target_compile_options(conv-utilf PRIVATE -cpmod)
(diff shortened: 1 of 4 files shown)


matthiasdiener:
should it be `target_link_options`?

evan-charmworks:
The usage of the parameter is entirely wrong, you have to specify the path to copy to and the .M file as its own command.

I tried `target_link_libraries` just to be sure and it gave:

```
CMake Error at src/libs/ck-libs/ampi/CMakeLists.txt:479 (target_link_libraries):
  Object library target "ampimod" may not link to anything.
```

`target_link_options` was added in CMake 3.13.

ericjbohm (approved):

matthiasdiener (approved):

stwhite91 (approved):

PR #3283, opened 2021-02-12
Reimplement CmiNode{All}Barrier as spinbarrier

rbuch:
Provides a large performance benefit over the previous implementation. Can be slightly slower than a pure spinbarrier, but much more cooperative with the system, so drastically outperforms a spinbarrier in oversubscribed scenarios or when other threads have useful work to do (not to mention power savings).

I removed the usage of C++20's `std::barrier` because performance there is a question mark, so I think it's better to use something we understand the performance of until we can benchmark the stdlib version. I am checking for and using C++20's atomic wait/notify because it promises to be "often more efficient than simple polling or pure spinlocks" (in addition to being much, much cleaner than the alternative).

Below is some example benchmarking data on Bridges-2 (2 sockets with 64 cores each), the test consisted of 100,000 consecutive barriers with no other work across 16, 32, 64, and 128 threads. Times reported are average time per barrier in microseconds. "pthread barrier" is the pthreads barrier, "cv barrier" is the previous implementation, "hybrid barrier" is the new implementation, and "spin barrier" is a simple pure spinbarrier. Going over 128 cores makes the spin barrier take so long that I didn't even let the tests finish.

Edit: The barrier has changed slightly since these runs, performance should be even better now.
```
[ronak@r474 barrier_test]$ ./a.out 16 100000
pthread barrier
24.6729
cv barrier
34.0666
hybrid barrier
2.18664
spin barrier
2.18501
[ronak@r474 barrier_test]$ ./a.out 32 100000
pthread barrier
51.8358
cv barrier
74.7482
hybrid barrier
5.00963
spin barrier
4.61406
[ronak@r474 barrier_test]$ ./a.out 64 100000
pthread barrier
71.419
cv barrier
158.54
hybrid barrier
8.63663
spin barrier
7.034
[ronak@r474 barrier_test]$ ./a.out 128 100000
pthread barrier
163.136
cv barrier
375.582
hybrid barrier
37.1808
spin barrier
26.4953

```


src/arch/util/machine-smp.C
```diff
@@ -514,48 +512,103 @@ static void CmiDestroyLocks(void)
...
 private:
-  std::mutex barrierMutex;
-  std::condition_variable barrierCond;
-  bool curSense;
-  unsigned int curCount;
+  alignas(CMI_CACHE_LINE_SIZE) std::atomic<bool> curSense;
+  alignas(CMI_CACHE_LINE_SIZE) std::atomic<unsigned int> curCount;
   const unsigned int barrierCount;
+
+  void pause() const
```
evan-charmworks, line 578: Could be marked `static` since it does not use `this`. May also be a good case for CMI_FORCE_INLINE since the function body is either one or zero instructions.

src/arch/util/machine-smp.C
```diff
@@ -514,48 +512,103 @@ static void CmiDestroyLocks(void)
...
+    // when spinning by ARM.
+    // https://developer.arm.com/documentation/dui0489/i/arm-and-thumb-instructions/yield?lang=en
+    asm volatile("yield");
+#  else
+    // For other platforms, do nothing.
+#  endif
+  }
+
+  template <size_t N>
+  void repeat_pause() const
```
evan-charmworks, line 601: Could be marked `static`.

src/arch/util/machine-smp.C
rbuch: Done.

src/arch/util/machine-smp.C
rbuch: Done.

src/arch/util/machine-smp.C
```diff
@@ -514,48 +512,117 @@ static void CmiDestroyLocks(void)
...
+
+    // If we're last, reset the count and flip the sense.
+    // Using acq_rel here effectively makes this read-modify-write a fence.
+    // Note: we compare against 1 because fetch_sub returns the previous value.
+    if (curCount.fetch_sub(1, std::memory_order_acq_rel) == 1)
+    {
+      curCount.store(barrierCount, std::memory_order_relaxed);
+      curSense.store(!sense, std::memory_order_release);
+#  if __cpp_lib_atomic_wait
+      curSense.notify_all();
```
matthiasdiener, line 538: Doesn't that need `<condition_variable>` ?

matthiasdiener (approved):

src/arch/util/machine-smp.C
rbuch: Nope, just `<atomic>` (https://en.cppreference.com/w/cpp/atomic/atomic/wait and https://en.cppreference.com/w/cpp/atomic/atomic/notify_all). It's a C++20 addition.

evan-charmworks (approved):

stwhite91 (approved):

Issue #3145, opened 2020-11-10
LB-related hangs using SpECTRE & charm v6.11.0-beta1


fmahebert:
Testing SpECTRE with charm tag v6.11.0-beta1, I find reproducible hangs when I turn on the LB.

I am running some evolution, and varying the number of chare array elements being balanced as well as the number of HPC system nodes I run the executable on. SpECTRE is using AtSync LB.

With `+balancer GreedyLB` and `+balancer RefineLB`, I consistently find the run hangs after a small number of LB calls (I've observed ~3-30). For the same setup, the hang will occur after the same number of LB calls, though this number depends on the number of chare elements and the number of processors. The hang occurs after returning from the LB, while attempting the next series of timesteps.

With `+balancer DummyLB` or without invoking any command-line LB, there is no hang and the evolution completes as expected. I am using the same SpECTRE version in which AtSync LB seemed to work on 6.10.2.

I haven't yet checked other LBs for this behavior.

It's not clear to me if I'm seeing a problem in the LBs themselves, or if the LB AtSync interface has changed and SpECTRE needs updating. If the latter, we haven't yet found an indication of such an interface change within the documentation.

I am happy to try any tests that may be helpful in diagnosing this issue.

rbuch:
Thanks for reporting this, it looks like it might be the same as #3108, which also hangs upon returning from load balancing. It seems likely that this is caused by something inside LB since it's now observable in two different applications at roughly the same point and when an actual load balancer is being used.

fmahebert:
I just tried with RotateLB and got a hang there too...

rbuch:
Do you still get crashes if you run with the `+LBSyncResume` flag? I'm not 100% sure, but I think the issue with #3108 and likely the issue here, too, is that `ResumeFromSync()` is called before all migrations are globally finished, so there's a race and not every element gets resumed (I still haven't proven this to myself definitively, but it looks like it's the case, and adding that flag empirically makes Enzo-E stop hanging). The old LB system doesn't guarantee that migrations are globally finished before `ResumeFromSync()` is called either, but based on the design, I think it's more likely that they would be done before control moves on.

I would test this on SpECTRE myself, but grepping through the source, I couldn't find anywhere `usesAtSync` is used or where `AtSync()` is called. Is there a particular branch or something that I should be looking at?

fmahebert:
[...]

Issue #3108, opened 2020-10-16
RotateLB hangs with v6.11.0-beta1


jobordner:
The Enzo-E load-balancing regression test hangs when RotateLB is used.  Some observations:

- works fine for +p2 and +p3: no hangs within the 10 balancing steps performed
- +p4 runs hang when calling the balancer for the 4th time
- +p4 hanging on the 4th attempt is repeatable
- hanging behavior does not occur in charm-6.10.2
- no other load balancers tested hung

rbuch:
Thanks for the report. Can you provide build information and the command line you used to execute so I can reproduce this?

jobordner:
This is using the `solver-dd` branch of https://github.com/jobordner/enzo-e.git run using "`charmrun +p4 bin/enzo-p input/load-balance-4.in +balancer RotateLB`"  I just tried the master branch of the main repository https://github.com/enzo-project/enzo-e.git and that worked fine.  Charm was built using `./build charm++ netlrts-linux-x86_64   -j4  --with-production --enable-tracing`.

rbuch:
I've been able to reproduce this, I'm currently trying to get to the bottom of what's going on. It seems like load balancing finishes when the hang happens, objects get their `ResumeFromSync` functions called, so it might be an issue in Enzo-E.

One question, it seems like the number of `Block` chares increases with every load balancing period, I assume this is expected because of refinement?

rbuch:
Using the `+LBSyncResume` flag seems to solve this in my tests, so I think there's some kind of object registration / calling `AtSync` race that's happening here.

rbuch:
Should be fixed in #3286.

PR #3286, opened 2021-02-17
LocMgr: Add epoch to location updates to prevent hangs

epmikida:
Without ordering location update messages, we can get incorrect entries in the location
tables when old updates arrive after new ones. In some cases this can cause stale cycles
that can trap messages indefinitely.

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2374,25 +2374,20 @@ void CkLocCache::pup(PUP::er& p)
      * indexes of local elements dont need to be packed since they will be
      * recreated later anyway
      */
-    int count = 0;
-    std::vector<int> pe_list;
-    std::vector<CmiUInt8> id_list;
-    for (const auto& itr : id2pe)
+    std::vector<CkLocEntry> entries;
+    for (const auto& itr : locMap)
     {
-      if (homePe(itr.first) == CmiMyPe() && itr.second != CmiMyPe())
+      if (homePe(itr.first) == CmiMyPe() && itr.second.pe != CmiMyPe())
       {
-        id_list.push_back(itr.first);
-        pe_list.push_back(itr.second);
-        count++;
+        entries.push_back(itr.second);
       }
     }
 
+    int count = entries.size();
     p | count;
-    // syncft code depends on this exact arrangement:
     for (int i = 0; i < count; i++)
     {
-      p | id_list[i];
-      p | pe_list[i];
+      p | entries[i];
     }
   }
   else
@@ -2401,16 +2396,14 @@ void CkLocCache::pup(PUP::er& p)
     p | count;
     for (int i = 0; i < count; i++)
     {
-      CmiUInt8 id;
-      int pe;
-      p | id;
-      p | pe;
-      inform(id, pe);
-      if (homePe(id) != CkMyPe())
+      CkLocEntry e;
+      p | e;
+      updateLocation(e);
+      if (homePe(e.id) != CkMyPe())
       {
-        requestLocation(id, homePe(id));
+        thisProxy[homePe(e.id)].updateLocation(e);
       }
-      CkAssert(getPe(id) == pe);
+      CkAssert(getPe(e.id) == e.pe);
     }
   }
 #endif
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2410,7 +2410,7 @@ void CkLocCache::pup(PUP::er& p)
       {
         requestLocation(id, homePe(id));
       }
-      CkAssert(whichPE(id) == pe);
+      CkAssert(getPe(id) == pe);
     }
   }
 #endif
@@ -2427,7 +2427,7 @@ void CkLocCache::requestLocation(CmiUInt8 id)
 
 void CkLocCache::requestLocation(CmiUInt8 id, const int peToTell)
 {
-  CkAssert(peToTell != CkMyPe());
+  if (peToTell == CkMyPe()) return;
 
   LocationMap::const_iterator itr = locMap.find(id);
   if (itr != locMap.end())
@@ -2438,12 +2438,35 @@ void CkLocCache::requestLocation(CmiUInt8 id, const int peToTell)
 
 void CkLocCache::updateLocation(const CkLocEntry& newEntry)
 {
+  CkAssert(newEntry.pe != -1);
   CkLocEntry& oldEntry = locMap[newEntry.id];
   if (newEntry.epoch > oldEntry.epoch)
   {
     oldEntry = newEntry;
     notifyListeners(newEntry.id, newEntry.pe);
   }
+}
+
+void CkLocCache::moveTo(CmiUInt8 id, int pe)
+{
+  LocationMap::iterator itr = locMap.find(id);
+
+  CkAssert(itr != locMap.end());
+  CkAssert(itr->second.pe == CkMyPe());
+
+  itr->second.pe = pe;
+  itr->second.epoch++;
+}
+
+
+void CkLocCache::insert(CmiUInt8 id, int epoch)
+{
+  CkLocEntry& e = locMap[id];
+  // TODO: This should be > probably, but demand creation needs some fixing up
+  CkAssert(epoch >= e.epoch);
+  e.id = id;
+  e.pe = CkMyPe();
+  e.epoch = epoch;
 }
 
 /*************************** LocMgr: CREATION *****************************/
@@ -2584,7 +2607,7 @@ void CkLocMgr::informHome(const CkArrayIndex& idx, int nowOnPe)
     // TODO: This may not need to be an idx update. Pretty sure the home will always
     // know the idx to id mapping.
     CmiUInt8 id = lookupID(idx);
-    thisProxy[home].updateLocation(idx, cache->locMap[id]);
+    thisProxy[home].updateLocation(idx, cache->getLocationEntry(id));
   }
 }
 
@@ -2596,11 +2619,8 @@ CkLocRec* CkLocMgr::createLocal(const CkArrayIndex& idx, bool forMigration,
 
   CkLocRec* rec = new CkLocRec(this, forMigration, ignoreArrival, idx, id);
   insertRec(rec, id);
-  CkLocEntry e;
-  e.id = id;
-  e.pe = CkMyPe();
-  e.epoch = epoch;
-  updateLocation(idx, e);
+  cache->insert(id, epoch);
+  updateLocation(idx, cache->getLocationEntry(id));
 
   if (notifyHome)
   {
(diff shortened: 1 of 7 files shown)


rbuch (approved):
Looks good to me, and it seems to fix the hang I've been seeing in my test application.

rbuch:
I think this may also fix hangs that have been observed in SpECTRE and Enzo (#3108, #3145). I'll test against them, but it'd be good to get this tested by the application teams as well. @nilsdeppe @jobordner

nilsdeppe:
@moxcodes @fmahebert could you give see if this fixes the hangs?

fmahebert:
I can test this, maybe tomorrow but if not then early next week.

fmahebert:
I've tested commit this branch (commit e97c80f) on the scaled-down SpECTRE problem described in #3145. Unfortunately, I still see the consistent hanging behavior.

As a reminder: this problem sets up a "dummy" evolution on an array chare and uses `AtSync` LB to migrate the elements. I am finding consistent hangs with 8 elements in the chare array and `+p5 +balancer RotateLB`.

If there's any specific information about the setup that would be helpful for me to provide, do let me know.

epmikida:
This PR is mostly ready, although I can't figure out why the oldbuild test is timing out. It runs fine on my machine, and doesn't seem to go any slower than the regular tests, even when looking through the CI logs. I'm working on the Charm4Py issue, but the documentation for building Charm4Py on darwin seems to be out of date.

rbuch:
> This PR is mostly ready, although I can't figure out why the oldbuild test is timing out. It runs fine on my machine, and doesn't seem to go any slower than the regular tests, even when looking through the CI logs. I'm working on the Charm4Py issue, but the documentation for building Charm4Py on darwin seems to be out of date.

How slow is it on your machine in an oversubscribed case? It seems like that's what's really slow in the CI, so perhaps there's something in your change that has caused increased contention or something that's rearing its head as big slowdowns in the oversubscribed case.

epmikida:
@nilsdeppe @fmahebert we addressed some more problems that we suspected may have caused hangs in location management.

fmahebert:
@epmikida I'll test this again as soon as I'm able; hopefully this week.

fmahebert:
I've tested the updated branch (Charm++ commit `6d12a43`) with SpECTRE and LB functionality seems restored.

I tested this using both the scaled-down LB test problem, and a SpECTRE evolution. In each case, turning on LB would cause hangs in Charm++ 6.11-beta1 but no hangs on `6d12a43`.

✅ Thanks!

src/ck-core/cklocation.C
```diff
@@ -2423,21 +2420,50 @@ void CkLocCache::requestLocation(CmiUInt8 id)
...
+  CkLocEntry& oldEntry = locMap[newEntry.id];
+  if (newEntry.epoch > oldEntry.epoch)
+  {
+    oldEntry = newEntry;
+    notifyListeners(newEntry.id, newEntry.pe);
+  }
 }
 
-void CkLocCache::updateLocation(CmiUInt8 id, int nowOnPe) { inform(id, nowOnPe); }
+void CkLocCache::moveTo(CmiUInt8 id, int pe)
```
rbuch, line 2446: Minor, but I don't like the nomenclature of this method, I think `moveTo` implies that it's involved in the moving process, it should be called something closer to `updateLocation` in my mind. Perhaps it can even be merged with the existing `updateLocation` (and have the asserts/assumption of existence be checked elsewhere if necessary).

src/ck-core/cklocation.C
```diff
@@ -2507,6 +2533,7 @@ void CkLocMgr::pup(PUP::er& p)
   p | lbmgrID;
   p | metalbID;
   p | bounds;
+  p | idCounter;
```
rbuch, line 2536: This isn't used elsewhere in this PR, I assume it's a fix for checkpoint/restart?

epmikida: It's related to checkpoint restart, but it is part of the PR. It was previously updated (weirdly) in the old inform method a few lines down. That method was removed with the API updates.

src/ck-core/ckmemcheckpoint.C
```diff
@@ -1064,7 +1064,7 @@ void CkMemCheckPT::updateLocations(int n, CkGroupID *g, CkArrayIndex *idx, CmiUI
 
   for (int i=0; i<n; i++) {
     CkLocMgr *mgr = CProxy_CkLocMgr(g[i]).ckLocalBranch();
-    mgr->updateLocation(idx[i], id[i], nowOnPe);
+    //mgr->updateLocation(idx[i], id[i], nowOnPe);
```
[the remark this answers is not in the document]
epmikida: Yes, must have missed some, I'll remove them.

src/ck-core/cklocation.C
epmikida: Agreed. Both moveTo and insert are a little weird sounding. They are named based on the conditions they are called in rather than what they do to the table which is weird. I'll try to come up with a better name. I'm a bit hesitant to overload updateLocation even more, because frankly the vast majority of operations on the cache are requests or updates, so we'd just end up with a bunch of versions of request and a bunch of versions of update. request/update are also the only entry methods and I was trying to keep them as just entry method calls, not intended for local use, but maybe that's foolish.

rbuch: Given that it's a cache, I think having things mostly being called something to do with requests or updates makes sense. The semantics of different variants should obviously be made clear, but I don't think there's anything inherently bad with multiple functions being called something like `updateLocation` since that's probably the best description of what the function does.

Wingpad: Perhaps something like `recordMovement` would be applicable here? "record" doesn't have the side-effecting conations of "update", at least. Maybe increased verbosity could be another option, like, "updateLoc(ation)Map" or "updateLocationEntry"?

As for "insert", "put" is always another option when dealing with maps.

src/ck-core/cklocation.ci
```diff
@@ -8,15 +8,14 @@ module CkLocation {
   group [migratable] CkLocCache {
     entry CkLocCache();
     entry [expedited] void requestLocation(CmiUInt8 id, int peToTell);
-    entry [expedited] void updateLocation(CmiUInt8 id, int nowOnPe);
+    entry [expedited] void updateLocation(CkLocEntry newEntry);
```
Wingpad, line 11: Can this be a const reference?

src/ck-core/cklocation.C
epmikida: Went with recordEmigration. can't say I love it, but can't say I have anything better either.

src/ck-core/ckmemcheckpoint.C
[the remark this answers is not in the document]
epmikida: The issue here is that some of these commented out things should maybe be returned in some way. They currently don't break anything by being commented out, but there would need to be a lot of restructuring probably to get them to work correctly. Namely, it should probably go through CkArray instead, and leave LocMgr as backend.

src/ck-core/cklocation.h
```diff
@@ -301,25 +311,50 @@ enum CkElementCreation_t : uint8_t
...
+
+  // Query the local location table
+  const CkLocEntry& getLocationEntry(CmiUInt8 id) const
+  {
+    LocationMap::const_iterator itr = locMap.find(id);
+    return itr != locMap.end() ? itr->second : nullEntry;
+  }
+  int getPe(const CmiUInt8 id) const { return getLocationEntry(id).pe; }
+  int getEpoch(const CmiUInt8 id) const { return getLocationEntry(id).epoch; }
+  int homePe(const CmiUInt8 id) const { return id >> CMK_OBJID_ELEMENT_BITS; }
```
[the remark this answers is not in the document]
epmikida: Agreed. This should be further cleaned up in the 64 bit update, but for now I'll use the ObjID method. Especially because it further shields the code from 64bit ID constants/implementation.

src/ck-core/cklocation.h
```diff
@@ -72,6 +72,14 @@ typedef enum : uint8_t
 PUPbytes(CkDeliver_t)
 
     class CkArrayOptions;
+
+struct CkLocEntry {
+  CmiUInt8 id = 0;
```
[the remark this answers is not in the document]
epmikida: Yes and no. When using full 64bit IDs, yes, 0 is the "invalid" ID. However, with the way CkLocMgr and CkArray current deal with 64bit IDs, it is not, because they don't use full IDs, only the element bits. Which is odd. This will change in the upcoming 64bit ID update, but for now I think this is fine. CkLocMgr and CkArray don't rely on checking this value for validity, and any other users of CkLocCache would be using full IDs, so this should work as intended for them.

src/ck-core/cklocation.h
```diff
@@ -301,25 +311,50 @@ enum CkElementCreation_t : uint8_t
...
   // Map of ID to PE
-  using IdPeMap = std::unordered_map<CmiUInt8, int>;
-  IdPeMap id2pe;
+  using LocationMap = std::unordered_map<CmiUInt8, CkLocEntry>;
+  LocationMap locMap;
 
   using Listener = std::function<void(CmiUInt8, int)>;
   std::list<Listener> listeners;
 
+  const CkLocEntry nullEntry = CkLocEntry();
```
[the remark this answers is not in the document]
epmikida: I was having issues getting it to be constexpr, and I also realized it probably makes more sense as a static member of CkLocEntry, so for now I moved it there. Still open for discussion though.

src/ck-core/cklocation.ci
```diff
@@ -8,18 +8,17 @@ module CkLocation {
   group [migratable] CkLocCache {
     entry CkLocCache();
     entry [expedited] void requestLocation(CmiUInt8 id, int peToTell);
-    entry [expedited] void updateLocation(CmiUInt8 id, int nowOnPe);
+    entry [expedited] void updateLocation(const CkLocEntry& newEntry);
   };
 
   group [migratable] CkLocMgr {
```
Wingpad, line 14: More as a future question than something immediately relevant, can CkLocMgr be templated on `Index` and specialized for `CkArrayIndex` by default? What would have to change for this to be possible?

src/ck-core/cklocation.h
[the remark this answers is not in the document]
rbuch: I think your solution is sensible enough, we can probably defer `constexpr`ization of stuff across the codebase for later.

rbuch (approved):

Wingpad (approved):

src/ck-core/cklocation.ci
epmikida: I've been wondering something like that myself. There's probably a lot of templating type stuff that can be done within the CkLocMgr/CkLocCache/CkArray/etc space that could be useful. A lot of it is probably pretty straightforward.

PR #3293, opened 2021-02-23
charmxi: Update list of builtin types

### LINKED ISSUE #3065: charmxi doesn't parse std::size_t (but parses size_t)
I ran into this error while building ExaM2M with charm master. 

The error I see is:
```
[PARSE-ERROR] Unexpected/missing token at line 15. Current token being parsed: 'size_t'.
/scratch/nitin/ExaM2M_2/src/IO/meshwriter.ci:15:11-38: error: invalid entry method attribute list
  const std::vector< std::size_t >& inpoel,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Fatal Error by charmc in directory /scratch/nitin/ExaM2M_2/build_debug/IO
```

A simpler example to reproduce the issue is an entry method like `entry void write(std::size_t abc);`. 
This throws the following error: 
```
../../../../bin/charmc   direct_api.ci
[PARSE-ERROR] Unexpected/missing token at line 7. Current token being parsed: 'size_t'.
direct_api.ci:7:9-32: error: invalid entry method attribute list
  entry void write(std::size_t abc);
      ^^^^^^^^^^^^^^^^^^^^^^^^
Fatal Error by charmc in directory /scratch/nitin/charm_2/tests/charm++/zerocopy/direct_api_test
   Command ../../../../bin/charmxi -orig-file direct_api.ci returned error code 1
charmc exiting...
```

Since this was working previously, by running git bisect, I found that this bug was introduced by 
https://github.com/UIUC-PPL/charm/commit/6498e1752bcb4b64fa5ac03c1fc416ab6dff05bc (PR https://github.com/UIUC-PPL/charm/pull/2775)

evan-charmworks:
- Remove unnecessary size_t and bool (revert from #2775)
- Add some rarer multi-word types

Fixes #3065


nitbhat (approved):

minitu:
I assume passing `size_t` and `bool` to entry methods would still work after the revert?

evan-charmworks:
> I assume passing `size_t` and `bool` to entry methods would still work after the revert?

Yes. For example, `tests/charm++/zerocopy/zc_post_modify_size/zc_post_modify_size.ci` includes `entry void recv_zerocopy(nocopypost int buffer[size], size_t size, bool isBcast);`.

There is no benefit to including them in BuiltinTypes. I searched through the charmxi source and couldn't find any special behavior that such a designation would give. My guess is that it is only necessary for types like `unsigned int` that have a space in the middle.

minitu (approved):

PR #3294, opened 2021-02-25
AMPI: Avoid pooling messages tagged as broadcast

evan-charmworks:
Co-authored-by: Sam White <white67@illinois.edu>

### DIFF
--- a/src/libs/ck-libs/ampi/ampi.C
+++ b/src/libs/ck-libs/ampi/ampi.C
@@ -7557,7 +7557,7 @@ void ampi::irecvBcast(void *buf, int count, MPI_Datatype type, int src,
   // if msg has already arrived, do the receive right away
   if (msg) {
     newreq->receive(this, msg, false);
-    delete msg; // never add bcast msgs to AmpiMsgPool
+    delete msg; // never add bcast msgs to AmpiMsgPool because they are nokeep
   }
   else { // ... otherwise post the receive
     postedBcastReqs.put(newreq);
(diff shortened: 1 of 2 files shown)


matthiasdiener (dismissed):
Out of curiosity:
- Whats the reason for not pooling them?
- New broadcast msgs still seem to be created from the pool

stwhite91:
> * Whats the reason for not pooling them?

Bcast msgs are sent via [nokeep] entry methods, so the same msg is delivered to all ranks across all PEs. So we can't have multiple ranks stick the same msg into a PE-local msg pool.

> * New broadcast msgs still seem to be created from the pool

ampi::makeBcastMsg() already doesn't use the pool and instead creates a new msg. Despite that, I don't see why we wouldn't be able to use the pool when sending a bcast msg? I think it should be safe to do, but the cost of allocating a small msg (the AmpiMsgPool is only for small msgs) is not as bad in comparison to the cost of a bcast as it is for a point-to-point msg so I'm not sure it's worth doing.

matthiasdiener:
> > * Whats the reason for not pooling them?
> 
> Bcast msgs are sent via [nokeep] entry methods, so the same msg is delivered to all ranks across all PEs. So we can't have multiple ranks stick the same msg into a PE-local msg pool.

I think it might be worth adding this as a comment to the code.

> 
> > * New broadcast msgs still seem to be created from the pool
> 
> ampi::makeBcastMsg() already doesn't use the pool and instead creates a new msg. Despite that, I don't see why we wouldn't be able to use the pool when sending a bcast msg? I think it should be safe to do, but the cost of allocating a small msg (the AmpiMsgPool is only for small msgs) is not as bad in comparison to the cost of a bcast as it is for a point-to-point msg so I'm not sure it's worth doing.
> 
> I agree that there's an asymmetry here though, in that we don't call into the AmpiMsgPool when creating a bcast msg but we do when freeing it. We could rectify that either way, either don't call into the AmpiMsgPool ever for bcasts or always do and catch bcasts within the AmpiMsgPool class.

Yeah, in the way the current code is written this does not become super clear. Maybe just add as a comment there as well?

evan-charmworks:
I've addressed the asymmetry.

stwhite91 (approved):

matthiasdiener (approved):

Issue #3295, opened 2021-02-26
memory safety violations in tests/charm++/chkpt


evan-charmworks:
From CI:

```
../../../bin/testrun  ./hello +p4 +restart log ++ppn 4  ++local +setcpuaffinity +CmiSleepOnIdle
Charmrun> scalable start enabled. 
Charmrun> started all node programs in 0.005 seconds.
Charm++> Running in SMP mode: 1 processes, 4 worker threads (PEs) + 1 comm threads per process, 4 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID: v6.11.0-beta1-131-ged334dcff
Charm++ built with internal error checking enabled.
Do not use for performance benchmarking (build without --enable-error-checking to do so).
Charm++> scheduler running in netpoll mode.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled. 
Charm++> Running on 1 hosts (1 sockets x 1 cores x 2 PUs = 2-way SMP)
Charm++> cpu topology info is gathered in 0.008 seconds.
Received 1 arguments: { |./hello| }
Main's MigCtor. a=987(0x97f45c), b[0]=654(0x97f460), b[1]=321, old PE number 2
Main's PUPer. a=123(0x97f45c), b[0]=456(0x97f460), b[1]=789
*** Error in `./hello': double free or corruption (fasttop): 0x000000000097b640 ***
[0] data on Group 0
[0] data on NodeGroup 0
[2] data on Group 0
[3] data on Group 0
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7ffff6ba87e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7ffff6bb137a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7ffff6bb553c]
./hello(_ZN9LBManager3pupERN3PUP2erE+0x17b)[0x577efb]
./hello(_Z14CkPupGroupDataRN3PUP2erE+0x4a3)[0x4f0ad3]
./hello(_Z13CkRestartMainPKcP8CkArgMsg+0x1ad)[0x4f2c6d]
./hello(_Z10_initCharmiPPc+0x19ac)[0x56974c]
./hello[0x6383e3]
./hello[0x638f5d]
[...]

Issue #2613, opened 2019-11-14
Checkpoint/restart hangs in SMP mode


minitu:
With tests/charm++/chkpt, the code hangs when it is checkpointed & restarted using 2 processes each with 2 worker threads on LC Lassen.(`jsrun -n2 -a1 -c2 -g1 -K1 -r2 ./hello ++ppn 2 +setcpuaffinity +showcpuaffinity` -> `jsrun -n2 -a1 -c2 -g1 -K1 -r2 ./hello +restart log ++ppn 2 +setcpuaffinity +showcpuaffinity`)

minitu:
Still hangs on OLCF Summit.

Issue #1955, opened 2018-08-07
tests/charm++/chkpt hangs for mpi-win-x86_64-smp 


nitbhat:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1955*

---

```

../../../bin/testrun  ./hello +p4 +restart log

Running on 4 processors:  ./hello +restart log
charmrun> /cygdrive/c/Program Files/Microsoft MPI/Bin/mpiexec -n 4  ./hello +restart log

Charm++> Running on MPI version: 2.0
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 4 processes, 1 worker threads (PEs) + 1 comm threads per process, 0 PEs total
Charm++> The comm. thread both sends and receives messages
Charm++ warning> fences and atomic operations not available in native assembly
Converse/Charm++ Commit ID: v6.8.2-853-g4146bf788
Charm++> Disabling isomalloc because mmap() does not work.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> Running on 1 hosts (1 sockets x 4 cores x 2 PUs = 8-way SMP)
[1] data on Group 1
[2] data on Group 2
[2] data on NOdeGroup 2
[1] data on NOdeGroup 1
Charm++> cpu topology info is gathered in 0.000 seconds.
Received 1 arguments: { |./hello| }
Main's MigCtor. a=987(0000000000285CF4), b[0]=654(0000000000285CF8), b[1]=321, old PE number 4
Main's PUPer. a=123(0000000000285CF4), b[0]=456(0000000000285CF8), b[1]=789
[3] data on Group 3
[3] data on NOdeGroup 3
CHello's PUPer. step=3.
[0] data on Group 0
[0] data on NOdeGroup 0
[0]CkRestartMain done. sending out callback.
```

Found this bug from manual testing. 

Build charm or LIBS using  ./build LIBS mpi-win-x86_64 smp --with-production --enable-error-checking --without-romio -j4

nitbhat:
*Original date: 2018-08-08 16:33:57*

---
This bug is also seen on charm-6.8.2.

evan-charmworks:
*Original date: 2018-11-19 22:22:46*

---
Was this fixed by #1960?

Issue #1775, opened 2018-01-11
chkpt test hangs for pami-linux-ppc64le-smp & pamilrts-linux-ppc64le-smp build


nitbhat:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1775*

---
On Summitdev, chkpt test (tests/charm++/chkpt) with pami-linux-ppc64le-smp build hangs with the following output:
```

Running on 2 processors:  ./hello +restart log +ppn 2
jsrun -n2 ./hello +restart log +ppn 2
Choosing optimized barrier algorithm name I0:HybridBinomial:SHMEM:P2P
Converse/Charm++ Commit ID: v6.8.2-58-gc448343
Warning> Randomization of virtual memory (ASLR) is turned on in the kernel, thread migration may not work! Run 'echo 0 > /proc/sys/kernel/randomize_va_space' as root to disable it, or try running with '+isomalloc_sync'.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled.
Setting default affinity
Charm++> Running on 1 unique compute nodes (160-way SMP).
Charm++> cpu topology info is gathered in 0.002 seconds.
Received 1 arguments: { |./hello| }
Main's MigCtor. a=987(0x10003b9f5bc), b[0]=654(0x10003b9f5c0), b[1]=321, old PE number 4
Main's PUPer. a=123(0x10003b9f5bc), b[0]=456(0x10003b9f5c0), b[1]=789
[1] data on Group 1
[3] data on Group 3
[2] data on Group 2
CHello's PUPer. step=3.
[0] data on Group 0
[1] data on NOdeGroup 1
[0] data on NOdeGroup 0
[0]CkRestartMain done. sending out callback.
```

To replicate, run with *make test TESTOPTS="++ppn 2"*

nitbhat:
*Original date: 2018-01-12 19:59:54*

---
Similar hang is seen for pamilrts-linux-ppc64le-smp build as well. (~~https://charm.cs.illinois.edu/gerrit/#/c/3141/~~ https://github.com/UIUC-PPL/charm/commit/333d3699234fd0c24ff93eb4e892e98411a0ae15)

PR #3296, opened 2021-02-27
Disable chkpt test in SMP mode

evan-charmworks:
Avoids CI and autobuild failures caused by lack of SMP-awareness in checkpoint/restore code.

See: #3295, #2613, #1955, #1775, #1510

### DIFF
--- a/tests/charm++/Makefile
+++ b/tests/charm++/Makefile
@@ -6,7 +6,6 @@ DIRS = \
   simplearrayhello \
   provisioning \
   load_balancing \
-  chkpt \
   delegation \
   queue \
   sdag \
@@ -27,6 +26,12 @@ DIRS = \
   longIdle \
   bombard \
   varTRAM \
+
+ifneq ($(CMK_SMP),1)
+ifneq ($(CMK_MULTICORE),1)
+  DIRS += chkpt
+endif
+endif
 
 FTDIRS = \
   jacobi3d \


epmikida (approved):
May be worth discussing in core ways to better handle features like chkpt, ft, etc. They are marketed as core features of the RTS, but are so lightly tested if at all. I'm not even sure we need separate tests for them. In theory, many of our tests should already be chkpt/ft compatible, and we just need to run them with these features on. Just like we do with SMP.

trquinn:
ChaNGa has a checkpoint test since we use the feature all the time.

evan-charmworks:
> In theory, many of our tests should already be chkpt/ft compatible, and we just need to run them with these features on. 

I don't believe this is the case.

epmikida:
> > In theory, many of our tests should already be chkpt/ft compatible, and we just need to run them with these features on.
> 
> I don't believe this is the case.

Many might be a strong word, but I would hope its a non-empty subset. And if there are not many, maybe that should also be re-evaluated as well.

evan-charmworks:
I just experienced another spurious CI failure due to this test failing in SMP and multicore, in #3297. We should merge this ASAP.

> ChaNGa has a checkpoint test since we use the feature all the time.

Do you use checkpointing in SMP mode?

trquinn:
> I just experienced another spurious CI failure due to this test failing in SMP and multicore, in #3297. We should merge this ASAP.
> 
> > ChaNGa has a checkpoint test since we use the feature all the time.
> 
> Do you use checkpointing in SMP mode?

All the time!

PR #3301, opened 2021-03-03
CI: Remove Python 2 test for charm4py

rbuch:
Official Python 2 support ended on January 1, 2020 (and that's after a long extension), so I think it's more than fine not to test it or support it anymore.


matthiasdiener (approved):

rbuch:
The charm4py test will fail until https://github.com/UIUC-PPL/charm4py/pull/197 is merged.

ZwFink (approved):

PR #3302, opened 2021-03-04
Print a line to indicate CMA usage for ZC API

nitbhat:


stwhite91 (approved):

src/conv-core/convcore.C
```diff
@@ -4051,6 +4051,9 @@ void ConverseCommonInit(char **argv)
   useCMAForZC = true;
   if (CmiGetArgFlagDesc(argv, "+noCMAForZC", "When Cross Memory Attach (CMA) is supported, the program does not use CMA when using the Zerocopy API")) {
     useCMAForZC = false;
+  } else {
+    if(CmiMyNode() == 0 && CmiMyRank() == 0 && CmiDoesCMAWork())
+      CmiPrintf("Charm++> CMA will be used for same node inter process Zerocopy API transfers\n");
```
evan-charmworks, line 4056: If this line will print during startup of every Charm program, we should clean it up a little bit. Maybe. `CMA active for same-node inter-process Zero Copy transfers`.

matthiasdiener: Also, does this honor `++quiet`?

rbuch: > Also, does this honor `++quiet`?

It should, `++quiet`disables `CmiPrintf` all throughout init.

PR #3298, opened 2021-03-02
Collection of fixes to address CI issues

rbuch:

PR #3305, opened 2021-03-09
ZC MultiplePostedBuffers Example: Fix incorrect callback creation

nitbhat:
This patch fixes the crash seen in this example on UCX CI builds.
Previously, the compReductionCb was constructed in the constructor
of zerocopyObject when the mainProxy data member was uninitialized.
With this patch, the compReductionCb construction call has been moved
to the testZeroCopy method, where the mainProxy is passed as a
parameter.

### DIFF
--- a/.github/workflows/ucx.yml
+++ b/.github/workflows/ucx.yml
@@ -17,7 +17,9 @@ jobs:
     #- name: Tmate session for debugging
     #  uses: mxschmitt/action-tmate@v2
     - name: install-prerequisites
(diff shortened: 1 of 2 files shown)


matthiasdiener (approved):
Merge into the general CI fixes PR?

nitbhat:
> Merge into the general CI fixes PR?

Yes, that seems like a better option to get these changes through.

nitbhat:
I've cherry-picked this change to https://github.com/UIUC-PPL/charm/pull/3298 for smoother integration.

PR #3313, opened 2021-03-25
Move message delivery and element creation to CkArray

epmikida:
Array message deliver and element creation currently bounce around a lot between CkArray CkLocMgr and CkLocRec. This cuts down on some of that overhead by cleaning up the delivery process and moving it into CkArray. CkLocMgr in the end should only be responsible for managing the binding of multiple arrays and their interaction with the location cache and lb database. It also handles the conversion of array indices to IDs. Other runtime system components that want to work with arrays should interface directly with CkArray, rather than going directly into CkLocMgr if possible.


src/ck-core/ckarray.C
```diff
@@ -1028,247 +1152,249 @@ void CkArray::remoteBeginInserting(void)
...
-
-void CkSetMsgArrayIfNotThere(void *msg) {
-  envelope *env = UsrToEnv((void *)msg);
+  msg_prepareSend(msg, ep, ckGetArrayID());
+  if (ckIsDelegated())  // Just call our delegateMgr
+    ckDelegatedTo()->ArraySend(ckDelegatedPtr(), ep, msg, _idx, ckGetArrayID());
+  else
+  {  // Usual case: a direct send
+    CkArray* localbranch = ckLocalBranch();
+    if (localbranch == NULL)
```
stwhite91, line 1209: Enclose within CMK_ERROR_CHECKING?

src/ck-core/ckarray.C
```diff
@@ -1028,247 +1152,249 @@ void CkArray::remoteBeginInserting(void)
...
+void CkSendMsgArray(int entryIndex, void* msg, CkArrayID aID, const CkArrayIndex& idx,
+                    int opts)
 {
-  CkArrayMessage *m=(CkArrayMessage *)msg;
-  msg_prepareSend(m,entryIndex,aID);
-  CkArray *a=(CkArray *)_localBranch(aID);
+  CkArrayMessage* m = (CkArrayMessage*)msg;
+  msg_prepareSend(m, entryIndex, aID);
+  CkArray* a = (CkArray*)_localBranch(aID);
   if (a == NULL)
```
stwhite91, line 1277: CMK_ERROR_CHECKING

src/ck-core/ckarray.C
```diff
@@ -1592,26 +1768,385 @@ void CkArray::flushStates() {
...
+        bufferedIndexMsgs[rec->getIndex()].push_back(msg);
+        return;
+      }
+      else
+      {
+        // Directly demand create. Since we are the location of a sibling, we don't need
+        // to request permission from home.
+        int chareType = _entryTable[msg->array_ep()]->chareIdx;
+        int ctor = _chareTable[chareType]->getDefaultCtor();
+        if (ctor == -1)
```
stwhite91, line 1940: CMK_ERROR_CHECKING

src/ck-core/ckarray.C
```diff
@@ -1592,26 +1768,385 @@ void CkArray::flushStates() {
...
+
+void CkArray::bufferForCreation(CkArrayMessage* msg, const CkArrayIndex& idx)
+{
+  // Only send a request if we haven't already requested
+  if (bufferedCreationMsgs.find(idx) == bufferedCreationMsgs.end())
+  {
+    // Figure out the constructor to call
+    int chareType = _entryTable[msg->array_ep()]->chareIdx;
+    int ctor = _chareTable[chareType]->getDefaultCtor();
+    if (ctor == -1)
```
stwhite91, line 2078: CMK_ERROR_CHECKING

src/ck-core/ckarray.C
epmikida: This is no longer part of the PR

src/ck-core/ckarray.C
epmikida: This is no longer part of the PR

src/ck-core/ckarray.C
```diff
@@ -1743,4 +1778,406 @@ void CkArray::ckDestroy()
...
+      if (msg->array_ifNotThere() == CkArray_IfNotThere_buffer)
+      {
+        // Directly buffer. Since we are the location of a sibling, we don't need to make
+        // a location request.
+        bufferedIDMsgs[id].push_back(msg);
+        return;
+      }
+      else
+      {
+        CkLocRec* rec = locMgr->elementNrec(id);
```
jszaday, line 1933: At the point when `CkArray` needs some awareness of `CkLocRec`, there may be an encapsulation breakdown. For example, at this location, shouldn't `locMgr` offer a direct mechanism for retrieving the index (without going through `CkLocRec`)?

epmikida: Yes, probably. That's a good point. In what was planned to be the next PR in this chain, CkLocRec is removed completely, so I think I just overlooked this piece.

src/ck-core/ckarray.C
```diff
@@ -1034,21 +1045,57 @@ void CkArray::insertElement(CkMarshalledMessage&& m, const CkArrayIndex& idx,
...
   }
-  int ctorIdx = me->array_ep();
+  int ctorIdx = m->array_ep();
   int chareType = _entryTable[ctorIdx]->chareIdx;
-  ArrayElement* elt = allocate(chareType, me, false, listenerData);
-  if (!locMgr->addElement(thisgroup, idx, elt, ctorIdx, (void*)me))
+  ArrayElement* elt = allocate(chareType, m, false, listenerData);
+
+  // Get the ID for this element
+  CmiUInt8 id = locMgr->getNewObjectID(idx);
```
jszaday, line 1064: Instead of (physically) inlining all of this logic, maintaining some degree of separation may be valuable. Particularly, separating the logic for registration and delivery into a helper function like `ckBeforeConstruction` or `ckRegisterElement` would help to keep `insertElement` concise.

jszaday:
PR#3431 is now a use case for these changes since they greatly aided in facilitating it.

src/ck-core/ckarray.C
epmikida: I've made some changes along these lines, although with entry method invocation being routed through CkLocRec, I could not fully remove that relationship yet. In the future though, I want CkLocRec either gone, or just a passive struct holding some small amount of information that is specific to a "location" rather than an "element".

src/ck-core/ckarray.C
epmikida: Got rid of this dependency and made a bunch of these types of methods private to CkLocMgr.

jszaday (approved):

epmikida:
Changa failure is because it is calling `CkLocMgr::getGroupID()`, so I guess I'll have to add that back in. Although I think it should be removed, as well as in other places like `CkArray`.

epmikida:
Fixed the changa issue and lumped the tiny cleanup commits into one. Should be ready for review now.

jszaday (approved):

rbuch (commented):
Looks good mostly.

One high level question: What happened to the previous `CkInRestarting` special casing? Is that not needed anymore?

src/ck-core/cklocation.C
```diff
@@ -2806,12 +2714,25 @@ bool CkLocMgr::addElementToRec(CkLocRec* rec, CkArray* mgr, CkMigratable* elt,
...
+  CmiUInt8 id;
+  if (lookupID(idx, id))
   {
-    thisProxy[home].requestLocation(idx, CkMyPe());
+    requestLocation(id);
+  }
+  else
+  {
+    int home = homePe(idx);
+    if (home != CkMyPe())
```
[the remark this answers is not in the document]
epmikida: When the element hasn't been created yet, and the index space is non-compressible. In that case, no ID would exist for that index.

epmikida:
> Looks good mostly.
> 
> One high level question: What happened to the previous `CkInRestarting` special casing? Is that not needed anymore?

I think those were already outdated. They only triggered during message delivery where the recipient array manager did not exist. But the only way to get to those code-paths was via calls from that very array manager that supposedly didn't exist.

jszaday (approved):

epmikida:
Those minor things I said I fixed...forgot to actually commit and push them. Whoops.

src/ck-core/cklocation.h
```diff
@@ -569,10 +572,15 @@ class CkLocMgr : public IrrGroup
 
   CkArrayIndex lookupIdx(const CmiUInt8& id) const
   {
+    CkLocRec* rec = nullptr;
     if (compressor)
     {
       return compressor->decompress(id);
     }
+    else if (rec = elementNrec(id))
```
evan-charmworks, line 580:
```
cklocation.h:580:18: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
    else if (rec = elementNrec(id))
             ~~~~^~~~~~~~~~~~~~~~~
cklocation.h:580:18: note: place parentheses around the assignment to silence this warning
    else if (rec = elementNrec(id))
                 ^
             (                    )
cklocation.h:580:18: note: use '==' to turn this assignment into an equality comparison
    else if (rec = elementNrec(id))
                 ^
                 ==
```

PR #3319, opened 2021-04-01
LocMgr: Fix syncft builds

epmikida:
This moves pupping of the location table to CkLocCache, where the table is stored. It
also fixes an issue where IDs were being packing but indices were being unpacked.


epmikida:
This should fix the autobuild issues in the syncft tests, however looking at the code makes me a bit suspicious as to how syncft tests were passing before. We were sizing/packing IDs, and unpacking indices, and I have no idea how that wasn't breaking things.

stwhite91 (approved):

evan-charmworks (approved):

PR #3332, opened 2021-04-14
cmake: honor arch in triplet

### LINKED ISSUE #3278: buildcmake ignores OS and arch fields of tuplet, preventing cross-compilation
Current builds which specify `netlrts-linux-i386` silently build x86_64 instead.

matthiasdiener:
Fixes #3278.


evan-charmworks (approved):
I tried `./build charm++ netlrts-linux-i386 -j4 -g3` and got this:

```
CMake Error at src/QuickThreads/CMakeLists.txt:17 (message):
  Unsupported Quickthreads version "i386-gcc"
```

This could be a separate PR. It at least confirms that the arch is honored.

Confirmed that gni-crayxc succeeds.

evan-charmworks:
i386 build still fails for me because CMK_SIZET_64BIT is set to 1.

evan-charmworks (approved):

stwhite91 (approved):

PR #3333, opened 2021-04-14
cmake: allow adding extra opts to end of triplet 

### LINKED ISSUE #3246: buildcmake ignores options embedded in target name
It is very useful to be able to pass the name of a Charm++ build directory to the build script rather than breaking out the hyphenated base and various options into separate arguments. This no longer works with the new build script.

./build charm++ ucx-linux-x86_64-smp-clang ...

... The CXX compiler identification is GNU 8.3.1

Somehow the build completes but you get a non-smp gcc-based build in a directory that suggest otherwise.

The buildold script contains code to "find longest prefix of version argument that exists as a directory in $src" and treat any following hyphen-separated options as if they had been entered separately on the build line.

matthiasdiener:
Fixes #3246.


rbuch (approved):

evan-charmworks:
Needs update

evan-charmworks (approved):

matthiasdiener:
Needs another review.

rbuch (approved):

PR #3340, opened 2021-04-17
LB: Add PUP to TreeLB

rbuch:


src/ck-ldb/TreeLB.C
```diff
@@ -198,6 +202,21 @@ void TreeLB::configure(json& config)
...
+  std::string configString;
+  if (p.isPacking())
+  {
+    configString = config.dump();
+  }
+  p | configString;
+  if (p.isUnpacking())
+  {
+    config = json::parse(configString);
+    init(CkLBOptions(seqno));
```
epmikida, line 216: Where is seqno from?

src/ck-ldb/TreeLB.C
```diff
@@ -198,6 +202,21 @@ void TreeLB::configure(json& config)
...
 }
 
+void TreeLB::pup(PUP::er& p)
+{
+  std::string configString;
+  if (p.isPacking())
+  {
+    configString = config.dump();
+  }
+  p | configString;
```
epmikida, line 212: Isn't this enforcing that you have the same LB config before and after restart? Shouldn't I be able to restart with whatever LB config I want, even if I didn't use any LB before restart?

src/ck-ldb/TreeLB.C
rbuch: `BaseLB`.

src/ck-ldb/TreeLB.C
rbuch: Potentially, that's where this gets a little tricky. On one hand, it perhaps makes sense to save the LBs used in the original run so that they'll be automatically used at restart, but on the other, they're not intrinsically a part of the application and one may want to restart with another LB, as you're saying.

epmikida: Yeah, I very strongly fall into camp #2. The load balancer is like the PE count. Not part of the application, part of the run. And when I rerun the application, I should be able to configure the run how I want w.r.t. number of PEs, whether I'm using LB, etc.

But, I think either way, if this fixes the crash, then we can merge this and discuss in core or something how we actually think this should work. The important thing now is fixing the bug.

epmikida (approved):

evan-charmworks (approved):

PR #3343, opened 2021-04-20
LB: Fix MetisLB crash when there are no edges

rbuch:
The METIS manual says that `adjncy` should be of size `2m` where `m = |E|` and that `adjwgt` can be `NULL` when edges are unweighted (they implicitly are when there are no edges, of course). However, internally when it checks the graph, it accesses element zero of both of these arrays, so they can't actually be `NULL` in practice, so create dummy arrays when there are no elements.


matthiasdiener (approved):
I think this is fine, although I'd personally just declare this as `std::vector<idx_t> adjncy(numEdges+1);`.

stwhite91 (approved):

evan-charmworks (approved):

PR #3350, opened 2021-05-03
Add warning for anytime migrations

PathikritGhosh:
Does not allow anytime migration if corresponding flag is turned off and throws a warning when attempting to migrate.

### DIFF
--- a/src/ck-core/ckarray.h
+++ b/src/ck-core/ckarray.h
@@ -308,7 +308,13 @@ public:
   virtual char *ckDebugChareName(void);
   virtual int ckDebugChareID(char*, int);
 
-  void ckEmigrate(int toPe) {ckMigrate(toPe);}
+  void ckEmigrate(int toPe) {
+    if(!_isAnytimeMigration) {
+      CkPrintf("Charm++> WARNING: Attempted anytime migration is disabled.\n");
+      return;
+    }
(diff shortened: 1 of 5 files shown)


epmikida:
Have you confirmed that the warning actually shows up when expected? I don't think any of the tests turn off anytime migration, and even if they do, they would still pass as this is just a warning.

A test for this change could also be added. One where you call migrateMe from within array elements and confirm they don't move, but that they do still move after load balancing. Not sure if that is necessary though.

PathikritGhosh:
> Have you confirmed that the warning actually shows up when expected? I don't think any of the tests turn off anytime migration, and even if they do, they would still pass as this is just a warning.
> 
> A test for this change could also be added. One where you call migrateMe from within array elements and confirm they don't move, but that they do still move after load balancing. Not sure if that is necessary though.



> Have you confirmed that the warning actually shows up when expected? I don't think any of the tests turn off anytime migration, and even if they do, they would still pass as this is just a warning.
> 
> A test for this change could also be added. One where you call migrateMe from within array elements and confirm they don't move, but that they do still move after load balancing. Not sure if that is necessary though.

Yes, I checked this throws a warning and blocks migrations. Also it allows LB migrations. I wrote a simple code to check this manually.

PR #3352, opened 2021-05-04
Trace communication to inline entry methods for LB

rbuch:
As reported by the SpECTRE team, communication to `[inline]` entry methods was not being added to the comm graph. For some reason, the RTS code explicitly avoided adding this to the comm graph, and a newer optimization that called the target EP directly on the object pointer when it was local without ever going into the RTS code also skipped this step. This change removes the condition excluding remote messages to `[inline]` EPs from having their communication traced and adds an explicit call to record the communication when the `[inline]` EP is directly invoked locally on the object pointer (even though this case doesn't actually involve any communication, it should be in the comm graph so that the LB knows that they are communicating and won't separate them when running communication based strategies).

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -3033,7 +3033,7 @@ int CkLocMgr::deliverMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
 // may be necessary due to migrations, etc.
 void CkLocMgr::sendMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
                        const CkArrayIndex* idx, CkDeliver_t type,
-                       int opts)
+                       int opts /* = 0 */)
 {
   // Trace this send for LB
   recordSend(idx, id, UsrToEnv(msg)->getTotalsize(), opts);
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2959,7 +2959,7 @@ int CkLocMgr::deliverMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
       return true;
     }
     // unknown location
-    deliverUnknown(msg, idx, type, opts);
+    deliverUnknown(msg, idx, opts);
     return true;
   }
 
@@ -2982,7 +2982,7 @@ int CkLocMgr::deliverMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
   {  // That sibling of this object isn't created yet!
     if (msg->array_ifNotThere() != CkArray_IfNotThere_buffer)
     {
-      return demandCreateElement(msg, rec->getIndex(), CkMyPe(), type);
+      return demandCreateElement(msg, rec->getIndex(), CkMyPe());
     }
     else
     {  // BUFFERING message for nonexistent element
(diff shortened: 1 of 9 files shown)


src/ck-core/cklocation.C
```diff
@@ -2935,10 +2929,20 @@ int CkLocMgr::deliverMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
...
+
+/************************** LocMgr: MESSAGING *************************/
+/// Deliver message to this element, going via the scheduler if local
+/// @return 0 if object local, 1 if not
+int CkLocMgr::deliverMsg(CkArrayMessage* msg, CkArrayID mgr, CmiUInt8 id,
+                         const CkArrayIndex* idx, CkDeliver_t type, int opts)
+{
+  CkLocRec* rec = elementNrec(id);
+
+  recordSend(idx, id, UsrToEnv(msg)->getTotalsize(), opts);
```
epmikida, line 2945: I'm pretty sure this will now record a send for the initial send, and every time the message is delivered. If you move the recordSend call to sendMsg, that should fix the issue. Although I'm not sure how it should interact with buffering.

rbuch: Should be fixed in 95cbdf7.

src/ck-core/cklocation.h
```diff
@@ -651,10 +650,13 @@ class CkLocMgr : public IrrGroup
   // int deliverMsg(CkMessage *m, CkArrayID mgr, const CkArrayIndex &idx, CkDeliver_t
   // type, int opts = 0);
   int deliverMsg(CkArrayMessage* m, CkArrayID mgr, CmiUInt8 id, const CkArrayIndex* idx,
-                 CkDeliver_t type, int opts = 0);
+                 CkDeliver_t type, bool firstAttempt = true, int opts = 0);
```
epmikida, line 653: Does true make sense as the default (for all of these defaults)? I would think that the first send is more of the exception? Just a minor thought though.

src/ck-core/cklocation.h
```diff
@@ -481,8 +480,8 @@ class CkLocMgr : public IrrGroup
   IndexMsgBuffer bufferedIndexMsgs;
   IndexMsgBuffer bufferedDemandMsgs;
 
-  void deliverAnyBufferedMsgs(CmiUInt8, MsgBuffer& buffer);
-  void deliverAnyBufferedMsgs(const CkArrayIndex&, CmiUInt8, IndexMsgBuffer&);
+  void deliverAnyBufferedMsgs(CmiUInt8, MsgBuffer& buffer, bool firstAttempt = true);
```
epmikida, line 483: I'm worried that this also presents more issues. How do we know if the buffered messages have had an attempted delivery yet? What if some have, and some haven't? I think this just inherently will lead to potential inconsistencies. What would happen if instead, we record the send when the message is buffered? Even though the send is not carried out at that point, it is still the case that the program is attempting to communicate from elem X to elem Y at that point in execution...

epmikida: Although, I realize with the latter point I mentioned, we wouldn't be able to include the PE, but why is that even a necessary thing to include? Isn't the comm graph object to object?

rbuch: I agree that it's a bit of a flaky situation, the fact that `deliverMsg` is reused on both the send side and receive side makes it inherently a bit of a sticky wicket. I assigned the values for `firstAttempt` based on tracing through the call graph, making it `false` only for things that should have already gone through `deliverMsg` once before. Ideally, we'd have separate send-side and deliver-side functions for this so that we can record the send for LB purposes in a function that is clearly used only for sending a message once its destination and other details are known (as I believe your refactor of the location manager will do).

Yes, the comm graph is generally object to object, but it saves object locations for hops/hopbytes calculations that MetaBalancer uses as one of its decision criteria, so it needs object locations as well.

src/ck-core/cklocation.h
rbuch: My tracing through the function call paths says yes, but I agree that it's a brittle solution that we should ideally solve in another way.

epmikida (approved):
I think some of the function names were initially confusing, but overall it looks good, although it will be largely blown away by #3313. This is good enough to fix release issues though.

evan-charmworks (approved):

epmikida (approved):

stwhite91 (approved):

PR #3355, opened 2021-05-05
Anytime migrations

PathikritGhosh:
- Throw warning when migrations are attempted when the flag is set to false.

### DIFF
--- a/src/ck-core/ckarray.h
+++ b/src/ck-core/ckarray.h
@@ -308,7 +308,11 @@ public:
   virtual char *ckDebugChareName(void);
   virtual int ckDebugChareID(char*, int);
 
-  void ckEmigrate(int toPe) {ckMigrate(toPe);}
+  void ckEmigrate(int toPe) {
+    if(!_isAnytimeMigration) {
+      CkPrintf("Charm++> WARNING: Attempted anytime migration is disabled.\n");
+    }
+    ckMigrate(toPe);}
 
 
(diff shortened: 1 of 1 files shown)


src/ck-core/ckarray.h
```diff
@@ -308,7 +308,11 @@ AllreduceMgr * allredMgr; // for allreduce
   virtual char *ckDebugChareName(void);
   virtual int ckDebugChareID(char*, int);
 
-  void ckEmigrate(int toPe) {ckMigrate(toPe);}
+  void ckEmigrate(int toPe) {
```
epmikida, line 311: Fix formatting: braces on their own lines

epmikida (commented):
Once the formatting is fixed this looks good for now. Once we have a better understanding of what exactly constitutes "anytime migration" and what triggers it, we should probably actually error and not migrate.

src/ck-core/ckarray.h
```diff
@@ -308,7 +308,13 @@ AllreduceMgr * allredMgr; // for allreduce
   virtual char *ckDebugChareName(void);
   virtual int ckDebugChareID(char*, int);
 
-  void ckEmigrate(int toPe) {ckMigrate(toPe);}
+  void ckEmigrate(int toPe)
+  {
+    if(!_isAnytimeMigration) {
```
epmikida, line 313: Another brace

epmikida (approved):

ZwFink (approved):

PR #3359, opened 2021-05-11
GPU-aware communication with UCX in Charm++

minitu:
Extracted from #2791.

### DIFF
--- a/benchmarks/charm++/cuda/gpudirect/Makefile
+++ b/benchmarks/charm++/cuda/gpudirect/Makefile
@@ -1,7 +1,7 @@
 DIRS = \
   latency \
   bandwidth \
-  latency-persistent \
+  latency-tag \
 
 TESTDIRS = $(DIRS)
 
--- a/benchmarks/charm++/cuda/gpudirect/bandwidth/bandwidth.C
+++ b/benchmarks/charm++/cuda/gpudirect/bandwidth/bandwidth.C
@@ -36,7 +36,7 @@ class Main : public CBase_Main {
 
     // Process command line arguments
     int c;
-    while ((c = getopt(m->argc, m->argv, "s:x:i:l:w:d:")) != -1) {
+    while ((c = getopt(m->argc, m->argv, "s:x:i:l:w:d:z")) != -1) {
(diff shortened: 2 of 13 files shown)


minitu:
Ready for review. Tested with the `verify` example (`examples/charm++/cuda/gpudirect/verify`) and OSU latency benchmark (`benchmarks/charm++/cuda/gpudirect/latency`).

src/arch/cuda/hybridAPI/hapi_impl.cpp
```diff
@@ -143,6 +143,13 @@ void hapiInit(char** argv) {
 #endif
   }
 
+  CmiNodeAllBarrier();
+
+  if (CmiInCommThread()) {
+    // FIXME: Comm. thread sets its device to be the same as worker thread 0
```
ericjbohm, line 149: Is this comment still accurate?  The code below implies otherwise.

src/arch/ucx/machine.C
```diff
@@ -133,12 +154,30 @@ static void UcxPrepostRxBuffers();
...
 
+#if CMK_CUDA
+inline void UcxInvokeRecvHandler(DeviceRdmaOp* op, DeviceRecvType type) {
+  switch (type) {
+    case DEVICE_RECV_TYPE_CHARM:
+      CmiInvokeRecvHandler(op);
+      break;
+    // TODO: AMPI and Charm4py
+    default:
+      CmiAbort("Invvalid recv type: %d\n", type);
```
ericjbohm, line 165: Typo in Invvalid

src/arch/cuda/hybridAPI/hapi_impl.cpp
minitu: It is still accurate. If you see the lines 330-332 below, the thread with CmiMyRank 0 saves its device number in GPUManager (which is a CSV variable) which is later used by the comm thread to map to the same device.

ericjbohm (commented):
Fairly long PR, but not sure that there is a good way to split it up.  Minor comments inline.

PR #3364, opened 2021-05-14
Reformat ckarray.{C,h}

epmikida:
This was automated reformating with clang-format

### DIFF
--- a/src/ck-core/ckarray.h
+++ b/src/ck-core/ckarray.h
@@ -26,13 +26,13 @@ Orion Sky Lawlor, olawlor@acm.org
 #ifndef __CKARRAY_H
 #define __CKARRAY_H
 
-#include "cklocation.h"
-#include "ckmulticast.h"
-#include "ckmemcheckpoint.h" // for CkArrayCheckPTReqMessage
 #include "ckarrayindex.h"
+#include "cklocation.h"
+#include "ckmemcheckpoint.h"  // for CkArrayCheckPTReqMessage
+#include "ckmulticast.h"
 
 /***********************************************************
-	Utility defines, includes, etc.
+        Utility defines, includes, etc.
 */
 extern void _registerCkArray(void);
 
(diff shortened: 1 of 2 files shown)


epmikida:
At Ronak's request, this moved reformatting from #3313 to it's own change.

src/ck-core/ckarray.h
```diff
@@ -87,191 +88,242 @@ class CkVerboseListener : public CkArrayListener {
 /*@{*/
 
 /*********************** Proxy Support ************************/
-//Needed by CBase_ArrayElement
-class ArrayBase { /*empty*/ };
+// Needed by CBase_ArrayElement
+class ArrayBase
+{ /*empty*/
```
Wingpad, line 93: Would it be better to simply delete `/*empty*/` under the new formatting guidelines?

rbuch (approved):
We should consider doing something like this for pure reformatting changes, too: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.

src/ck-core/ckarray.h
epmikida: Done

jszaday (dismissed):

epmikida:
> We should consider doing something like this for pure reformatting changes, too: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.

Yeah this seems like it would be a good idea

epmikida:
I can't tell why this keeps failing. It's timing out occasionally on runs where Charm++ doesn't even finish initialization. I don't think it is specific to this PR, but it seems to be happening more and more frequently. Almost always in ZC tests, but this time it looks like it was in util.

rbuch (approved):

epmikida:
> > We should consider doing something like this for pure reformatting changes, too: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.
> 
> Yeah this seems like it would be a good idea

I tried this out and it works. I'll merge this PR and then if we think its a good idea, we can set up a separate PR to add these revs to a blame file that ignores them.

jszaday (approved):

rbuch (approved):

PR #3367, opened 2021-05-18
Improve Assert/Enforce Ergonomics

rbuch:


matthiasdiener (approved):

src/conv-core/converse.h
```diff
@@ -1104,15 +1104,19 @@ extern void __cmi_assert(const char *);
      (__cmi_assert ("Assertion \"" __CMK_STRING(expr) \
                     "\" failed in file " __FILE__ \
                     " line " __CMK_XSTRING(__LINE__) "."), 0)))
-
-#if ! CMK_ERROR_CHECKING
-#define CmiAssert(expr) ((void) 0)
-#else
-#define CmiAssert(expr) \
+#define CmiEnforceMsg(expr, msg)                    \
```
epmikida, line 1107: Will this not allow for formatted strings? In a lot of cases then, we'd still probably want to do `if (condition) CkAbort(...)` instead.

rbuch: Ah, yeah, good point, this should support format strings, I'll add support for that.

rbuch:

src/conv-core/converse.h
```diff
@@ -1098,21 +1098,33 @@ void  CmiError(const char *format, ...);
...
                     " line " __CMK_XSTRING(__LINE__) "."), 0)))
-
-#if ! CMK_ERROR_CHECKING
-#define CmiAssert(expr) ((void) 0)
+#define CmiEnforceMsg(expr, msg, ...) \
+  ((void) ((expr) ? 0 : \
+     (__cmi_assert_msg ("Assertion \"" __CMK_STRING(expr) \
+                        "\" failed in file " __FILE__ \
+                        " line " __CMK_XSTRING(__LINE__) ".", \
+                        msg, __VA_ARGS__), 0)))
```
matthiasdiener, line 1120: Could you pass the whole thing to `CmiAbort`?

rbuch: No, as the comment on line 1101 says, the issue there is that the stringified `expr` may have a `%` in it (e.g. `var%d == 0` to assert that `var` is a multiple of `d`), which can then cause problems when it is treated as a format string. Calling `CmiAbort` through this `__cmi_assert_msg` wrapper lets us handle the macro generated and user provided strings separately.

rbuch: Ah, I see, I think your edited version makes sense. I'll update this branch with something of that ilk.

rbuch: Added in a0d0933.

rbuch: Actually, it seems the answer is no, calling `CmiAbort` directly is an issue because it can be called in any scope and there are other implementations of `CmiAbort`, namely in `CkArray` and `CkMigratable`, etc. This causes issues when there are asserts from a static context because `CmiAbort` resolves to the version in the inheritance tree, which then fails because it can't be called statically, so we need to call it through some kind of wrapper to make sure that this always resolves correctly.

rbuch: Also, I think it's worthwhile to create a function to pass the user message into the `CmiAbort` to ensure that the user message is printed at the end of the output; if we just print it out and then call abort, then it'll be above the stack trace and other printed errors, so it can be hard to find.

epmikida (approved):

matthiasdiener (approved):

epmikida:
This probably needs a follow up. It requires 3 args to CkEnforce/CkAssert, so doing something simple like: `CkAssert(ctor != -1, "Cannot demand create an element without a default ctor");` fails to compile.

rbuch:
> This probably needs a follow up. It requires 3 args to CkEnforce/CkAssert, so doing something simple like: `CkAssert(ctor != -1, "Cannot demand create an element without a default ctor");` fails to compile.

Fixed in #3391.

Issue #2969, opened 2020-07-18
Anomalous initialization messages in SMP build


minitu:
The following anomalous initialization statements show up with `1darray hello` when run with multiple PEs per process (e.g. 4 processes with 10 PEs per process) on PAMILRTS SMP build on LLNL Lassen:

```
choi18@lassen7:1darray$ jsrun -n4 -a1 -c10 -g1 -K2 -r4 ./hello 80 +ppn 10 +pemap L0-160:4
Choosing optimized barrier algorithm name I0:HybridBinomial:SHMEM:P2P
Charm++> Running in SMP mode: 4 processes, 10 worker threads (PEs) + 0 comm threads per process, 40 PEs total
Charm++> There's no comm. thread. Work threads both send and receive messages
Choosing optimized barrier algorithm name I0:HybridBinomial:SHMEM:P2P
Charm++> Running in SMP mode: 4 processes, 10 worker threads (PEs) + 0 comm threads per process, 40 PEs total
Charm++> There's no comm. thread. Work threads both send and receive messages
Choosing optimized barrier algorithm name I0:HybridBinomial:SHMEM:P2P
Charm++> Running in SMP mode: 4 processes, 10 worker threads (PEs) + 0 comm threads per process, 40 PEs total
Charm++> There's no comm. thread. Work threads both send and receive messages
Choosing optimized barrier algorithm name I0:HybridBinomial:SHMEM:P2P
Charm++> Running in SMP mode: 4 processes, 10 worker threads (PEs) + 0 comm threads per process, 40 PEs total
Charm++> There's no comm. thread. Work threads both send and receive messages
Converse/Charm++ Commit ID: v6.11.0-devel-283-ge4a6e7b3f
Isomalloc> Synchronized global address space.
CharmLB> Load balancer assumes all CPUs are same.
Charm++> cpu affinity enabled.
Charm++> cpuaffinity PE-core map (logical indices): 0-160:4
Charm++> Running on 1 hosts (2 sockets x 20 cores x 4 PUs = 160-way SMP)
Charm++> cpu topology info is gathered in 0.005 seconds.
Running Hello on 40 processors for 80 elements
...
```

I did a git bisect and found the offending commit to be #2531.

Before this commit:
```
choi18@lassen7:1darray$ jsrun -n4 -a1 -c10 -g1 -K2 -r4 ./hello 80 +ppn 10 +pemap L0-160:4
[...]

Issue #2814, opened 2020-04-21
glibc heap-allocates internal FILE buffers on first print


evan-charmworks:
glibc internally allocates buffers for stdout and stderr the first time something is written to those file descriptors. UCX builds of #2531 expose an issue where these first print statements happen inside an AMPI rank with `-memory isomalloc` enabled, so these internal buffers are allocated locally for the rank, and if that rank migrates away, the pointers become invalid and the next write to the file descriptor will segfault inside glibc. It is unknown why this behavior is only exposed by this PR.

Here are example stack traces:

```gdb
Thread 1 "pgm" received signal SIGSEGV, Segmentation fault.
__memmove_sse2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:288
288     ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory.
(gdb) bt
#0  __memmove_sse2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:288
#1  0x00007ffff77f0792 in _IO_new_file_xsputn (n=1, data=0x555555ae8610, f=<optimized out>) at fileops.c:1240
#2  _IO_new_file_xsputn (f=0x7ffff79496a0 <_IO_2_1_stdout_>, data=0x555555ae8610, n=1) at fileops.c:1201
#3  0x00007ffff77d7a42 in __vfprintf_internal (s=0x7ffff79496a0 <_IO_2_1_stdout_>, format=0x555555ae8610 "[%d] migrated from %d to %d\n", ap=ap@entry=0x2aaaa010fbf0, mode_flags=mode_flags@entry=0) at ../libio/libioP.h:948
#4  0x00007ffff77c2e1f in __printf (format=<optimized out>) at printf.c:33
#5  0x00005555558247fe in MPI_Tester::testMigrate (this=0x2aaaa010ff24) at test.C:302
#6  0x0000555555825171 in AMPI_Main_cpp (argc=1, argv=0x2aaaa00003c0) at test.C:506
#7  0x000055555589f540 in AMPI_threadstart (data=0x555555d09550) at ampi.C:1096
#8  0x0000555555825cb0 in startTCharmThread (msg=0x555555d09530) at tcharm.C:163
#9  0x0000555555a529dc in CthStartThread (arg=...) at threads.C:1836
#10 0x0000555555a52e4f in make_fcontext ()
[...]

PR #3368, opened 2021-05-20
Revert "Add _IO_file_overflow call to ConverseCommonInit"

evan-charmworks:
See #2814, #2531, #2969


evan-charmworks:
Fails with the same error that this code originally intended to solve.

trquinn (commented):
This does fix the problems I reported in #2969

PR #3382, opened 2021-06-05
Tracing: Fix begin/endExecute mismatch for threaded EPs

rbuch:
Modify charmxi to emit a beginExecute before starting a [threaded] EP
to match the one emitted by CthSuspend. Without this, there is only
one beginExecute (emitted when Charm invokes an EP), but two
endExecutes (one for CthSuspend and one from Charm when the invoked EP
finishes). This only matters when a [threaded] EP is invoked as an EP,
all subsequent suspensions and resumptions of that EP are traced
correctly.


slm960323:
Tested with paratreet. All good.

epmikida (approved):

PathikritGhosh (approved):

PR #3387, opened 2021-06-07
ZC Post Async Posting: charmxi changes

nitbhat:


src/ck-core/ckarray.C
```diff
@@ -1370,6 +1370,27 @@ bool CkArrayBroadcaster::deliver(CkArrayMessage* bcast, ArrayElement* el, bool d
   }
 }
 
+bool CkArrayBroadcaster::deliverAlreadyDelivered(CkArrayMessage *bcast, ArrayElement *el, bool doFree)
```
nitbhat, line 1373: @epmikida: I added this method for delivering a message to an element after the posting is completed. It is similar to the previous version of `CkArrayBroadcaster::deliever`. I'm guessing this needs to be updated now?

epmikida: This looks fine. It may change slightly with new delivery code, but at least for now I believe it is accurate.

src/ck-core/ckarray.h
```diff
@@ -703,6 +703,15 @@ class CkArray : public CkReductionMgr
     }
   }
 
+ inline size_t getNumLocalElems() {
+    return localElemVec.size();
+  }
+
+  inline unsigned int getEltLocalIndex(const CmiUInt8 id) {
```
epmikida, line 710: This doesn't seem like it should ever be part of the public facing interface. How the elements are stored internally could change in the future.

src/ck-core/ckrdma.C
```diff
@@ -2093,3 +2059,510 @@ void zcPupIssueRgets(CmiUInt8 id, CkLocMgr *locMgr) {
...
+  int numops = 0;
+  int bufsize = 0;
+  int rootNode;
+  PUP::fromMem up((void *)((CkMarshallMsg *)EnvToUsr(env))->msgBuf);
+  up|numops;
+  up|rootNode;
+  for(int i=0; i<numops; i++){
+    CkNcpyBuffer w;
+    up|w;
+    w.ncpyEmInfo->tagArray[CmiMyRank()].resize(localElems * numops);
```
evan-charmworks, line 2218: Try `(size_t)localElems * numops` to avoid the LGTM warning.

src/ck-core/ckrdma.C
```diff
@@ -2093,3 +2059,510 @@ void zcPupIssueRgets(CmiUInt8 id, CkLocMgr *locMgr) {
...
+  std::vector<int> *tagArray = NULL;
+  NcpyBcastRecvPeerAckInfo *peerAckInfo = NULL;
+
+  if(emMode == ncpyEmApiMode::BCAST_RECV) {
+    if(env->getMsgtype() == ArrayBcastFwdMsg) {
+      CkArray *mgr = getArrayMgrFromMsg(env);
+      int numElems = mgr->getNumLocalElems();
+
+      tagArray = new std::vector<int>[CmiMyNodeSize()];
+      tagArray[CmiMyRank()].resize(numElems * numops);
```
evan-charmworks, line 2342: `(size_t)numElems * numops`

src/ck-core/ckarray.h
nitbhat: I am internally maintaining the state of the zerocopy posting at the node level and for that reason, I need to find out the local index to appropriately change the state of the corresponding element in the data structure that I am maintaining. Even if the element storage is changed in the future, as long as they are indexable from 0...n-1, this method could be appropriately changed. As far as keeping it in the public-facing API, I agree. I could change the accessing functions to friend functions if you think that's a better solution?

src/ck-core/ckrdma.C
```diff
@@ -1602,69 +1609,10 @@ void CkReplaceSourcePtrsInBcastMsg(envelope *env, NcpyBcastInterimAckInfo *bcast
...
-  p|source;
-  CmiSpanningTreeInfo &t = *(getSpanningTreeInfo(rootNode));
-  if(peerAckInfo->decNumPeers() - 1 == 0) {
-    QdCreate(1);
-    CMI_ZC_MSGTYPE(env) = CMK_ZC_BCAST_RECV_ALL_DONE_MSG;
-    CmiPushPE(CmiRankOf(peerAckInfo->peerParentPe), env);
-  }
-}
 
+#if CMK_SMP || CMK_UTH_VERSION
```
evan-charmworks, line 1613: `CMK_UTH_VERSION` is no longer part of Charm

src/ck-core/ckrdma.C
```diff
@@ -2093,3 +2059,510 @@ void zcPupIssueRgets(CmiUInt8 id, CkLocMgr *locMgr) {
...
+  ncpyPost[index].postAsync = false;
+}
+
+void setPostStruct(CkNcpyBufferPost *ncpyPost, int index, CkNcpyBuffer &buffObj, CmiUInt8 elemIndex) {
+  ncpyPost[index].srcBuffer = (void *)buffObj.ptr;
+  ncpyPost[index].srcSize = buffObj.cnt;
+  ncpyPost[index].ncpyEmInfo->tagArray = buffObj.ncpyEmInfo->tagArray;
+  ncpyPost[index].opIndex = index;
+  ncpyPost[index].arrayIndex = elemIndex;
+}
```
evan-charmworks, line 2206: It might be good to factor out the array indexing into `CkNcpyBufferPost & post = ncpyPost[index];`

src/ck-core/ckrdma.h
```diff
@@ -424,4 +469,97 @@ void zcPupIssueRgets(CmiUInt8 id, CkLocMgr *locMgr);
...
+static inline constexpr size_t safe_sizeof(T * ptr)
+{
+    return sizeof(T);
+}
+
+template <>
+inline constexpr size_t safe_sizeof(void * ptr)
+{
+    return 1;
+}
```
evan-charmworks, line 549: Since this is in a .h file, it would be good to put these in `namespace details` to avoid potential conflicts since this is a fairly generic name. There is an example of this in `src/util/pup.h`.

src/ck-core/ckrdma.C
nitbhat: Fixed.

src/ck-core/ckrdma.C
nitbhat: Fixed.

src/ck-core/ckrdma.h
nitbhat: That's a good point, added in the namespace. There are a lot of other global functions in the ckrdma implementation, I'll add another patch to wrap up those in a namespace.

PR #3391, opened 2021-06-09
fixup! Add {Cmi/Ck}{Enforce/Assert} variants that take a custom message

rbuch:


evan-charmworks:
Why not just replace `(expr, msg, ...)` with `(expr, ...)`?

epmikida (approved):

rbuch:
> Why not just replace `(expr, msg, ...)` with `(expr, ...)`?

The added empty string then becomes an unused argument, so it gets warned about when using `-Wformat-extra-args`, for example.

epmikida (approved):

stwhite91 (approved):

PR #3392, opened 2021-06-12
LLVM OpenMP Runtime 12.0.0

ZwFink:
- Updates the LLVM OpenMP Runtime used for OMP Charm to the latest stable, v12.0.0


stwhite91:
Any advice on reviewing this? Are there any changes to files that we have altered for our Charm OpenMP implementation?

rbuch:
This doesn't seem to build properly. With the CMake build, this doesn't seem to be built at all (likely because of #3396, I don't think that's to do with the changes here), but using `buildold` gives me the following error:
```
CMake Error at CMakeLists.txt:12 (message):
  Direct configuration not supported, please use parent directory!
```

This may be as simple as removing that check from the beginning of `CMakeLists.txt`, I'll try doing that and see if that fixes this.

rbuch:
> This doesn't seem to build properly. With the CMake build, this doesn't seem to be built at all (likely because of #3396, I don't think that's to do with the changes here), but using `buildold` gives me the following error:
> 
> ```
> CMake Error at CMakeLists.txt:12 (message):
>   Direct configuration not supported, please use parent directory!
> ```
> 
> This may be as simple as removing that check from the beginning of `CMakeLists.txt`, I'll try doing that and see if that fixes this.

That simple fix didn't work, this requires more structural changes, namely the `include/` folder hierarchy no longer matches what the new version of the library uses, some CMake changes seem to be needed to support the way it's built, etc.

PR #3393, opened 2021-06-14
Fix threaded ep

slm960323:
Assign the right ep to CthThread and use it with traceResume.


rbuch:
As discussed in Core, we should remove the `ep` field from Converse since it's a Charm++-specific thing, so we shouldn't cross the abstraction boundary, instead using something like `Ctv` or `CkAddThreadListeners` to cross that boundary.

PR #3400, opened 2021-06-22
Tests: Add AtSync test

rbuch:


tests/charm++/atsync/atsync.C
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  void done(void)
+  {
+    CkPrintf("All done\n");
+    CkExit();
+  };
+};
+
+class atsync : public CBase_atsync
```
epmikida, line 46: Why the weird style for a class/chare name?

tests/charm++/atsync/atsync.C
```diff
@@ -0,0 +1,98 @@
...
+  {
+    CkPrintf("atsync (%d) created on %d\n", thisIndex, CkMyPe());
+    counter = 0;
+    originalPE = CkMyPe();
+    usesAtSync = true;
+  }
+
+  atsync(CkMigrateMessage* m) {}
+
+  void ResumeFromSync()
```
epmikida, line 63: The test doesn't really seem to "test" anything other than does it crash or hang. Should there be something maybe in ResumeFromSync to ensure that all elements are even actually hitting the AtSync. As written, the test would still pass even if ResumeFromSync is only ever called on 0.

epmikida (commented):
I'm a little confused as to what this is explicitly meant to be testing that differs from, for example, LB tests.

tests/charm++/atsync/atsync.C
rbuch: I was trying to match an existing example, but obviously this should be cleaned up, I'll do that.

tests/charm++/atsync/atsync.C
rbuch: Yeah, that's all it really does, I originally wrote this to catch issues with `AtSync` where the internal state was getting messed up, but it was also useful in the unordered location updates issue, so failure is meant to be expressed as a hang (either deadlock or livelock). Adding more stuff is a good idea, though, I'll at the very least add a counter incremented in `ResumeFromSync` and then return that as a min reduction before exiting to make sure that all the objects are resumed the correct number of times.

epmikida (approved):

PR #3404, opened 2021-07-02
Fix handling of spaces in bash script command line arguments

evan-charmworks:
This affects charmc, shell script charmruns, AMPI toolchain wrappers, and
unix2nt scripts.

This commit extends space-safe idioms to buildcmake and buildold, though
arguments to them are not space-safe due to the handoff to configure. The
conv-mach scripts have not been fixed because configure runs them but does
not use bash. Building Charm from a path containing spaces will remain
unsupported until the above is fixed.


matthiasdiener:
This looks good, but might be a bit risky close to the release? Also, can we add some of those scripts to the shellcheck test?

rbuch:
> This looks good, but might be a bit risky close to the release? Also, can we add some of those scripts to the shellcheck test?

Note that we've already created a release branch, so things merged now won't go into the release unless they explicitly placed onto the release branch.

stwhite91 (approved):
Agree that this shouldn't be put on the 7.0 release branch

evan-charmworks:
> Also, can we add some of those scripts to the shellcheck test?

They don't pass. Fixing unrelated issues and also replacing `[` and `test` with `[[` would be good as followup.

matthiasdiener (approved):

Issue #3402, opened 2021-06-29
Infinite invocations of singleton chare constructors


epmikida:
If a singleton chare contains a constructor, where the only argument to that constructor is a proxy to the same type of singleton chare, it may get end up getting called infinitely.

The following program should be able to reproduce the bug every time. As long as the constructor with the single proxy argument is included, and at least one constructor taking a proxy is called, the bug will occur. If the single proxy constructor is removed, then the other constructor can be called just fine and will behave as expected.

[test.zip](https://github.com/UIUC-PPL/charm/files/6736445/test.zip)

epmikida:
Another way to reproduce is to add `entry Fib(CProxy_Fib f)` and `Fib(CProxy_Fib f) { CkPrintf("Error\n"); }` to fib.ci and fib.C in the fibonacci example in the repository.

jszaday:
Still working on debugging this but wanted to quickly share that, interestingly, this is actually generating new messages for each new invocation. 

![image](https://user-images.githubusercontent.com/7907791/124606779-d2b8f580-de32-11eb-9793-b75987ae613e.png)

jszaday:
The recurrence seems to be occuring because it overrides the copy constructor of the `CProxy_` class. For some reason (that is unknown to me), for singleton chares, calling `CProxy_[...]::CProxy_[...](args)` when the `args` matches one of the chare's constructors -- it actually creates a new chare. This behavior seems problematic to me since, in my opinion, chare creation should only occur through `ckNew` (or `insert`, `create[here|home]`, etc.).

    #0  CmiAlloc (size=size@entry=112) at /home/szaday2/workspace/charm/src/conv-core/convcore.C:3207
    #1  0x00005555556f6585 in envelope::alloc (incEvent=true, groupDepNumRequest=..., prio=0, size=<optimized out>, type=4 '\004') at /home/szaday2/workspace/charm/src/ck-core/envelope.h:308
    #2  _allocEnv (groupDepNum=..., prio=0, size=<optimized out>, msgtype=4) at /home/szaday2/workspace/charm/src/ck-core/envelope.h:478
    #3  CkAllocMsg (msgIdx=7, msgBytes=<optimized out>, prioBits=0, groupDepNum=...) at /home/szaday2/workspace/charm/src/ck-core/msgalloc.C:37
[...]

PR #3405, opened 2021-07-06
Remove Implicit Singleton Chare Creation w/ CProxy

jszaday:
This PR resolves issue #3402, removing the relevant parts of Charmxi that generate code that overrides the copy constructor. It eliminates implicit creation of singleton chares via `CProxy_[...]::CProxy_[...]`, forcing users to use `ckNew`. This seems to be an unexamined legacy behavior.


epmikida (approved):

epmikida:
The failure here looks to be that CI issue that we've been dealing with. A hang before charm initialization even completes. Probably unrelated to the patch itself.

jszaday:
> The failure here looks to be that CI issue that we've been dealing with. A hang before charm initialization even completes. Probably unrelated to the patch itself.

Agreed. The only thing that changed between this commit and the successful one was deleting comments.

evan-charmworks (approved):

PR #3406, opened 2021-07-07
CkArray: Fix semantics of initCallback/doneInserting

epmikida:
Array maps previously automatically called doneInserting on the array after initial insertion. This occurred even for arrays with 0 elements. For cases where users then did dynamic insertion and called doneInserting themselves, doneInserting was essentially a no-op (with some messaging). This change makes it so doneInserting is only automatically called for arrays with static insertion.

### DIFF
--- a/doc/charm++/manual.rst
+++ b/doc/charm++/manual.rst
@@ -6226,6 +6226,13 @@ where the number of initial elements is 0. Applications can call
 ``opts.setStaticInsertion(false)`` to override this behavior for cases where there are a
 non-zero number of initial insertions, but more dynamic insertions will follow.
 
+If the application needs to know when an array has been fully constructed, CkArrayOptions
+provides ``CkArrayOptions::setInitCallback(CkCallback)``. The callback passed will be
+invoked once every element in the initial set of elements has been created. This works
+both for bulk insertion, and if dynamic insertion is used to create the initial elements.
+In the latter case, after the initial elements have been inserted and ``doneInserting`` is
+called, the initialization callback will be invoked once all insertions have completed.
+
 .. _array map:
 
 Initial Placement Using Map Objects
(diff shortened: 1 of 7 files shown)


epmikida:
Note that this doesn't resolve the doneInsertion synchronization issues that are showing up in AMPI and elsewhere, but it's a step in the right direction, and should in theory at least make those errors less common.

stwhite91 (approved):

epmikida:
So it turns out this patch has the following issue: static insertion is actually set to false by default, regardless of how the array is created. During core, I'll propose the following changes:

Get rid of the _isStaticInsertion flag, and instead use the method of array creation as the way to set the default value of `CkArrayOptions::staticInsertion`. If I construct my array as follows:
```CProxy_Foo::ckNew(numElements);```
or
```CkArrayOptions opts; opts.setNumInitial(numElements); // numElements != 0```
```CProxy_Foo::ckNew(opts);```
Then we will assume staticInsertion is true. If instead, I use `CProxy_Foo::ckNew();` or use `numElements == 0`, we will assume dynamic insertion.

Both of these assumptions can be overridden by using the existing `CkArrayOptions::setStaticInsertion(bool)` as normal.

With the additional changes in already in this PR, this would cause our semantics to match up with what is already in use and documented in the manual regarding the need/purpose of `doneInserting`.

I don't think this should cause any major issues since it aligns with our current documentation, but discussion here would be helpful as well.

ericjbohm (dismissed):

stwhite91 (approved):

epmikida:
With the latest commit, this makes it so both doneInserting and initCallback work as expected for most cases. The one remaining issue is for cases where initCallback is used, and there are multiple phases of inserting this will behave weirdly. Granted, even in the old implementation it behaved weirdly in that initCallback meant nothing and would be called multiple times, once for each doneInserting call.

epmikida:
This is ready for review now.

evan-charmworks (approved):

stwhite91 (approved):

PR #3418, opened 2021-07-09
CkArray: Change defaults for static insertion

### LINKED ISSUE #3412: CkArray: Set better defaults regarding static insertion
By default, newly created arrays have static insertion set to false. This can be changed by the user directly via CkArrayOptions, or via a global variable which sets the default, _isStaticInsertion, for the whole application. This doesn't match up with common use cases, and also causes conflict with how dynamic insertion and doneInserting work in the manual.

Proposed change is to make the default value of static insertion set based on how the array is constructed. The global variable will be done away with. Then, for arrays which are created with a non-zero number of initial elements, static insertion will default to true. For those created with 0 initial elements, it will default to false. The user can still override these defaults by setting it directly in CkArrayOptions as normal, but that should only be needed for cases where there is an initial number of elements inserted at construction, but there will also be more added dynamically later on.

It will also be easy to add checks that throw errors if these rules are disobeyed. Currently the system gives no feedback if elements are inserted dynamically into a static array.

This change is also necessary to have #3411 work effectively.

epmikida:
Closes #3412


evan-charmworks (dismissed):

stwhite91 (approved):
Any documentation or test/example changes needed for this?

epmikida:
> Any documentation or test/example changes needed for this?

Good question. I'll check on that. I know the startupTest creates arrays weirdly, and could use some cleanup, but maybe as a separate patch.

evan-charmworks (dismissed):

src/ck-core/ckarrayoptions.h
```diff
@@ -82,10 +82,11 @@ class CkArrayOptions {
   CkCallback initCallback; // Callback to be invoked after chare array creation is complete
   bool anytimeMigration;                            // Elements are allowed to move freely
   bool disableNotifyChildInRed;  // Child elements are not notified when reduction starts
-  bool staticInsertion;          // Elements are only inserted at construction
   bool broadcastViaScheduler;    // broadcast inline or through scheduler
   bool sectionAutoDelegate;      // Create a mCastMgr and auto-delegate all sections
 
+  int staticInsertion; // -1 = unset, 0 = false, 1 = true
```
rbuch, line 88: Maybe make this into an enum?

src/ck-core/init.C
```diff
@@ -342,9 +342,9 @@ static inline void _parseCommandLineOpts(char **argv)
 	  _isNotifyChildInRed = false;
 	}
 
-	_isStaticInsertion = false;
-	if (CmiGetArgFlagDesc(argv,"+staticInsertion","Array elements are only inserted at construction")) {
-	  _isStaticInsertion = true;
+	if (CmiGetArgFlagDesc(argv,"+staticInsertion", "DEPRECATED")) {
+    CmiPrintf("WARNING: +staticInsertion has been deprecated.\n"
```
rbuch, line 346: Indentation?

src/ck-core/ckarrayoptions.h
```diff
@@ -235,6 +236,24 @@ class CkArrayOptions {
...
+    if (numInitial.dimension == 0) return false;
+    bool empty = false;
+    bool shorts = numInitial.dimension > 3;
+    for (int i = 0; i < numInitial.dimension; i++) {
+      if (shorts && ((short*)numInitial.data())[i] == 0) {
+        empty = true;
+        break;
+      } else if (!shorts && numInitial.data()[i] == 0) {
+        empty = true;
+        break;
```
rbuch, line 250: Perhaps I'm missing something, but why can't a statically inserted element have an index value of 0 in any dimension?

epmikida:
> Any documentation or test/example changes needed for this?

As far as I can tell, the command line argument this change removes was completely undocumented. Documentation to CkArrayOptions may need updating. Currently though, it counts .insert calls as being fine with static insertion. Basically according to the manual right now, everything that isn't demand creation is static insertion. Or at least, that's my interpretation.

src/ck-core/ckarrayoptions.h
epmikida: This is just the default. If there are 0 initial elements, then we assume that the array will be created by dynamic insertion.

rbuch: Aha, I misread that as an array index, but it's the size of the initial elements. Makes sense.

stwhite91 (approved):

evan-charmworks (approved):

src/ck-core/ckarrayoptions.h
```diff
@@ -82,10 +82,16 @@ class CkArrayOptions {
...
-  bool staticInsertion;          // Elements are only inserted at construction
   bool broadcastViaScheduler;    // broadcast inline or through scheduler
   bool sectionAutoDelegate;      // Create a mCastMgr and auto-delegate all sections
 
+  enum class InsertionType : char {
+    UNSET,
+    DYNAMIC,
+    STATIC
+  };
+  InsertionType insertionType; // -1 = unset, 0 = false, 1 = true
```
rbuch, line 93: This comment is now potentially wrong and should be removed.

rbuch: Fixed in https://github.com/UIUC-PPL/charm/pull/3428.

PR #3420, opened 2021-07-15
doc: say that CMake is the default build system

### LINKED ISSUE #3089: Update CMake build documentation
The current documentation about building Charm++ with CMake still says that it is experimental, doesn't work on Windows, etc. This should be updated to reflect the current state of the build, seeing as it's the default when running `./build` in v6.11.0.

matthiasdiener:
and other various fixes

Fixes #3089.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,13 +13,13 @@
 cmake_minimum_required(VERSION 3.4)
 
 # Print warning
-message("#################################################################")
-message("# The CMake build system may not support advanced build options.    #")
-message("# Please let us know if you experience any problems by opening  #")
-message("# an issue on our GitHub (https://github.com/UIUC-PPL/charm),   #")
-message("# with the below configuration information.                     #")
-message("# You can run './buildold' to use the previous build system.    #")
-message("#################################################################")
+message("##################################################################")
+message("# The CMake build system may not support advanced build options. #")
+message("# Please let us know if you experience any problems by opening   #")
+message("# an issue on our GitHub (https://github.com/UIUC-PPL/charm),    #")
+message("# with the below configuration information.                      #")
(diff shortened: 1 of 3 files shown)


rbuch:
This should remove the "beta version" header from `CMakeLists.txt`, too.

matthiasdiener:
> This should remove the "beta version" header from `CMakeLists.txt`, too.

It's still not really finished though - is there a better way we can express this?

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ cmake_minimum_required(VERSION 3.4)
 
 # Print warning
 message("#################################################################")
-message("# The CMake build system is currently a beta version.           #")
+message("# The CMake build system does not support all build options.    #")
```
rbuch, line 17: I think the current change sounds a little too negative, maybe something like this is a bit smoother:
```suggestion
message("# The CMake build system may not support advanced build options.    #")
```

matthiasdiener: That's a good idea, thanks!

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ cmake_minimum_required(VERSION 3.4)
 
 # Print warning
 message("#################################################################")
-message("# The CMake build system does not support all build options.    #")
+message("# The CMake build system may not support advanced build options.    #")
 message("# Please let us know if you experience any problems by opening  #")
 message("# an issue on our GitHub (https://github.com/UIUC-PPL/charm),   #")
 message("# with the below configuration information.                     #")
```
evan-charmworks: The alignment of the right border is broken now.

matthiasdiener: Thanks, fixed.

evan-charmworks (approved):

rbuch (approved):

PR #3424, opened 2021-07-17
Add CthIsMainThread to Consolidate Logic

jszaday:
This adds a library-accessible function for detecting whether the currently executing thread is the main thread. The motivations for this are two-fold: (a) to determine whether or not the currently executing thread should be suspended, and (b) to determine whether the currently executing entry method is `[threaded]`. For libraries that overload aspects of Charm++'s initialization routines (i.e., group creation), (a) pertains to writing threading-agnostic yield routines like:

```cpp
inline void yield(void) {
  auto th = CthSelf();
  if (CthIsMainThread(th)) {
    CsdScheduler(0);
  } else {
    CthYield();
  }
}
```

While the complex machinations in `convcore.C` suggest that the main thread can be suspended (and that other threads will inherit its role of running the scheduler while it is suspended), I have found that, in practice, the main thread should _not_ be suspended. Rudimentary screening suggests that crashes can occur and, more generally, there are concerns wrt migrations since it is not bound to a particular chare.

Previously, this information was restricted to `convcore.C`; this means it had to be inferred (through known properties of the main thread) in `threads.C`. This PR breaks this encapsulation, making the information publicly available. Whether or not this is "harmful" is a matter for discussion; however, I believe that it is benign since it is only useful to expert users.

### DIFF
--- a/src/conv-core/convcore.C
+++ b/src/conv-core/convcore.C
@@ -2126,11 +2126,15 @@ void CthResumeNormalThread(CthThreadToken* token)
 #endif
 }
 
+int CthIsMainThread(CthThread t) {
+  return t == CpvAccess(CthMainThread);
+}
+
 void CthResumeSchedulingThread(CthThreadToken  *token)
 {
   CthThread t = token->thread;
   CthThread me = CthSelf();
-  if (me == CpvAccess(CthMainThread)) {
+  if (CthIsMainThread(me)) {
     CthEnqueueSchedulingThread(CthGetToken(me),CQS_QUEUEING_FIFO, 0, 0);
   } else {
     CthSetNext(me, CpvAccess(CthSleepingStandins));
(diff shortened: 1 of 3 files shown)


rbuch (approved):

rbuch:
LGTM, but can you explain how this accomplishes (b), identifying if EPs are `[threaded]` or not?

jszaday:
> LGTM, but can you explain how this accomplishes (b), identifying if EPs are `[threaded]` or not?

Since anybody can spawn a `CthThread` at any point, even a non-threaded EP, this isn't a hard guarantee. That said, one could reasonably write something like and get their point across: 

```cpp
void myThreadSpecificFn(void) {
  auto me = CthSelf();
  CkAssert(!CthIsMainThread(me) && "myThreadSpecificFn cannot be called from the main thread; is your EP [threaded]?");
  ...
}
```

slm960323 (approved):
LGTM.

PR #3425, opened 2021-07-17
Improve CkLocalBranchOther and Node/Group Creation Consistency

jszaday:
This PR attempts to make the code paths for node/group creation more consistent. It also ensures that the object has been constructed before it is returned from `CkLocalBranchOther`, otherwise, there is a risk it may return a pointer to an uninitialized memory block. I encountered difficulties with this while experimenting while using `CkLocalBranchOther` on chare-arrays.


src/ck-core/ck.C
```diff
@@ -722,22 +744,12 @@ void CkCreateLocalNodeGroup(CkGroupID groupID, int epIdx, envelope *env)
...
-  CksvAccess(_nodeGroupIDTable).push_back(groupID);
-
-  PtrQ *ptrq = CksvAccess(_nodeGroupTable)->find(groupID).getPending();
-  if(ptrq) {
-    void *pending;
-    while((pending=ptrq->deq())!=0) {
-      _CldNodeEnqueue(CkMyNode(), pending, _infoIdx);
-    }
-    CksvAccess(_nodeGroupTable)->find(groupID).clearPending();
-  }
```
rbuch, line 739: Why separate this into two different critical sections?

src/ck-core/ck.C
```diff
@@ -656,6 +657,25 @@ void CkCreateChare(int cIdx, int eIdx, void *msg, CkChareID *pCid, int destPE)
...
+inline void CkReadyEntry(TableEntry& entry, bool nodeLevel) {
+  // the ready flag is set first to expedite the unblocking of
+  // node-level peers (that depend on the current [node]group)
+  entry.setReady();
+
+  PtrQ *ptrq = entry.getPending();
+  if(ptrq) {
+    void *pending;
+    while((pending=ptrq->deq())!=nullptr) {
+      if (nodeLevel) {
```
rbuch, line 669: Minor formatting stuff, make add spaces around operators, make spacing consistent.

src/ck-core/ck.C
jszaday: Well, it puts the object into the table so it can access itself during its creation (e.g., array reduction managers call ckLocalBranch and expect to get their associated group), but doesn't notify the pending listeners until its construction has finished (i.e., after the EP is invoked).

src/ck-core/ck.C
jszaday: Will do, this was copy-pasted so it needs adjustment.

src/ck-core/ck.C
jszaday: Also, if the dependencies can individually specify whether or not they're node-level (which would be a trivial addition from this PR) -- both node/groups would require two critical sections.

rbuch: Ah, I see, makes sense. I think it'd be useful to have that explanation enshrined as a comment in the code.

epmikida (approved):

rbuch (approved):

PR #3431, opened 2021-07-25
Add Mechanism for Message Forwarding

jszaday:
This PR adds a mechanism for forwarding messages from one chare-array element to another, primarily intended for use after the deletion of an element. This was motivated by my work on distributed section creation, wherein facilitating message forwarding can eliminate synchronization. In the following diagram, when CH1 is deleted, the RTS begins forwarding its messages to CH0. This avoids potential concerns like, "what if CH3 contributes to CH1 before it is updated to use CH0?"

image

Mechanically speaking, forwarding requests are instantiated on the home PE of the "old" index; only one copy of each request is kept in the system. Messages that arrive for the "old" index are updated and resent. Updating entails setting a boolean flag (e.g., "forwarded") and overriding the "old" index (see `getDestination`). Chare-array elements can implement a virtual method for screening inbound forwarded messages. By default, all forwarded messages are accepted but, via screening, they can only accept forwarded messages for certain EPs, etc.

When an element is recreated at a forwarded index, the forwarding record is obliterated and normal delivery resumes. This exploits the fact that, when an object is created, it always sends an `updateLocation` to the home PE.

This feature is primarily intended for internal use and is currently left undocumented until it is more robustly tested, merging now would be akin to a "semi-secret beta".


jszaday:
This PR is no longer necessary for my work, suspending its integration until mainstream motivations come along.

PR #3356, opened 2021-05-05
AMPI: Add pieglobals privatization method

evan-charmworks:

PR #3434, opened 2021-07-27
AMPI: Add prerequisites for pieglobals

evan-charmworks:


evan-charmworks:
I'm moving these commits into their own PR so they can be reviewed and merged without pieglobals itself (which is where the existing review comments were), and so that I don't have to keep force-pushing #3356.

src/conv-core/isomalloc.C
```diff
@@ -970,6 +973,8 @@ struct isommap
     pup_raw_pointer(p, allocated_extent);
     p | use_rdma;
 
+    p | protect_regions;
+
```
stwhite91, line 977: maybe add a comment somewhere here about not pupping heap_record and use_recording because they reset after migration

stwhite91 (approved):

src/conv-core/isomalloc.C
evan-charmworks: I feel like the existing comment by their definitions in the struct would be sufficient.

matthiasdiener (approved):

PR #3438, opened 2021-08-03
Add documentation summary logs

PathikritGhosh:


doc/projections/manual.rst
```diff
@@ -1735,3 +1736,29 @@ that we have not resolved at this time.
...
+
+-   The EP count is given in the first line of each individual processor logs along with the size of the interval.
+
+    eg: -  ``ver:7.1 0/4 count:6665 ep:270 interval:1.000000e-03 phases:0``
+
+-   The data is one encoded. So for decoding, the following line decodes to 14 consecutive zeros, followed by 65, followed by 0 and 3 consecutive 32767.
+
+    eg: ``0+14 65+1 0+1 32767+3``
+
+-   The data is in order of EP followed by bin. So, for the above example the first 270 values are for 270 EPs in the 0th bin. The next 270 values are for the 1st bin and so on.
```
rbuch, line 1760: I think this could use some more elucidation, to make it clearer that the entries are grouped and ordered by bin and then each bin consists of entries that are ordered by EP index. Also, each data line of the log files should be described here (with separate descriptions for Summary and SumDetail logs).

PathikritGhosh:
Waiting on #3448

PR #3442, opened 2021-08-05
Include information about uncaught exceptions in user code

ZwFink:
Currently, uncaught C++ exceptions in user code cause the following error message to be created:
```
------------- Processor 6 Exiting: Called CmiAbort ------------
Reason: Unhandled C++ exception in user code.
```

This patch attempts to retrieve the exception that was thrown and includes the result of ```exception.what()```, as shown below:
```
------------- Processor 6 Exiting: Called CmiAbort ------------
Reason: Unhandled C++ exception in user code: vector::_M_default_append.
```

The hope is that this information will be useful in debugging user code (it was useful for me). If the exception cannot be retrieved, the first message above is printed.
Further information about exception_ptr: https://en.cppreference.com/w/cpp/error/exception_ptr

### DIFF
--- a/src/ck-core/init.C
+++ b/src/ck-core/init.C
@@ -1277,7 +1277,24 @@ void _initCharm(int unused_argc, char **argv)
 	int inCommThread = (CmiMyRank() == CmiMyNodeSize());
 
 	DEBUGF(("[%d,%.6lf ] _initCharm started\n",CmiMyPe(),CmiWallTimer()));
-	std::set_terminate([](){ CkAbort("Unhandled C++ exception in user code.\n");});
+	std::set_terminate([](){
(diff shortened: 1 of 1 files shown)


evan-charmworks:
The code LGTM. The formatting needs fixes: the leading whitespace, space between `if` and `(`, and Allman style curly braces where they line up with the beginning of the statement that leads into them.

ZwFink:
> The code LGTM. The formatting needs fixes: the leading whitespace, space between `if` and `(`, and Allman style curly braces where they line up with the beginning of the statement that leads into them.

I think I've fixed the issues you've mentioned, please let me know if that is not the case.

evan-charmworks (approved):

matthiasdiener (approved):

PR #3446, opened 2021-08-09
AMPI: Reorganize privatization test

evan-charmworks:
- Allows testing of globals and thread-locals simultaneously 
- Cleans up file naming
- Removes unnecessary/redundant tests
- Decreases amount of terminal output


stwhite91 (approved):
Commit message would be better if it mentioned why the reorganization is being done

evan-charmworks:
> Commit message would be better if it mentioned why the reorganization is being done

Done

matthiasdiener (approved):

PR #3448, opened 2021-08-10
Summary/Sum Details changes

PathikritGhosh:


src/ck-perf/trace-summary.h
```diff
@@ -30,16 +30,93 @@ class BinEntry {
...
+    BinEntry(double t, double idleT, unsigned long msgSize, unsigned long msgCount,
+             CkVec<unsigned long> msgSizePerEP, CkVec<unsigned long> msgCountPerEP,
+             unsigned long recvSize, unsigned long recvCount,
+             CkVec<unsigned long> recvSizePerEP, CkVec<unsigned long> recvCountPerEP,
+             unsigned long extRecvSize, unsigned long extRecvCount,
+             CkVec<unsigned long> extRecvSizePerEP, CkVec<unsigned long> extRecvCountPerEP
+             ): _time(t), _idleTime(idleT),
+                 _msgSize(msgSize), _msgCount(msgCount),
+                 _recvSize(recvSize), _recvCount(recvCount),
+                 _extRecvSize(extRecvSize), _extRecvCount(extRecvCount)
```
rbuch, line 60: All of the vectors can be added in this initializer list rather than being initialized in the body.

rbuch (changes requested):

src/ck-perf/trace-summary.h
PathikritGhosh: These are not vectors values, but integer values of the total counts. These are not printed anywhere but are useful to debug if something does not match. But this part can be removed entirely without affecting the logs.

src/ck-perf/trace-summary.C
```diff
@@ -476,9 +498,76 @@ void SumLogPool::write(void)
...
+        } else if(event == 5) {
+            vec = pool[k].getExtRecvCountPerEP();
+        } else if(event == 6) {
+            vec = pool[k].getExtRecvSizePerEP();
+        }
+
+        //need to take min as we store _entryTable + 10 elements
+        //and in the first few bins, entryTable.size() does not show all the elements
+        int len = std::min(vec.size(), _entryTable.size());
+        for(int l = 0; l < len; ++l) {
```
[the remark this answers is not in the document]
PathikritGhosh: I can change this or change the earlier to follow the format of comm logs. But the earlier method was not a linear access of memory, which is why I changed this in the commlogs. But maybe the logs will be smaller if done that way. Let me know what you make of this.

PathikritGhosh:
> Note that this change should update the version number of logs since it is updating the format. The corresponding change in Projections should check this version number to detect if it has communication data or not.

Done

Issue #2850, opened 2020-05-12
NAMD occasionally hangs on Frontera with mpi-smp builds


nitbhat:
I stumbled upon this issue while doing some performance runs. 

This looks like a startup hang probably caused during RTS initialization. 
```

c191-001.frontera(1060)$ ibrun /scratch1/03808/nbhat4/namd/Linux-x86_64-g++-smp-impi-prod/namd2 ++ppn 13 +pemap 4-55:2,5-55:2 +commap 0,2,1,3 julio_input/runZIKV-50M-atoms.namd
TACC:  Starting up job 821440
TACC:  Starting parallel tasks...
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 64 processes, 13 worker threads (PEs) + 1 comm threads per process, 832 PEs total
Charm++> The comm. thread both sends and receives messages
Charm++> Using recursive bisection (scheme 3) for topology aware partitions
Converse/Charm++ Commit ID: v6.11.0-devel-225-g4504e8531
```

On running git bisect, I found that the bug was introduced by https://github.com/UIUC-PPL/charm/commit/ecea95f9938e35f32144fd0e465f9628d62d8ba0

stwhite91:
Can you try running with `+no_isomalloc_sync`?

nitbhat:
I don't see the hangs when I run with `+no_isomalloc_sync`.

evan-charmworks:
I'm not surprised there is a hang in isomalloc_sync considering I had to fight the code a bit to get something that worked with netlrts, netlrts-smp, and multicore.

Could #2838 be related?

Maybe something about MPI's use of thread 0 as a comm thread instead of thread N is affecting this, though I used `CmiInCommThread()` instead of hardcoding an ID check, and we don't encounter this issue on other mpi-smp builds.

It could be that I am not driving the comm thread properly in a way that only causes an issue with this machine.

The log also mentions partitions. Maybe that is related. isomalloc_sync does no special handling for partitions.

For now NAMD can pass `+no_isomalloc_sync` if they have the issue.

stwhite91:
I would not be surprised if partitions caused the issue, or one of the things you mentioned. We could also add a build-time option, if there isn't one already, to disable isomalloc_sync entirely

evan-charmworks:
[...]

PR #3449, opened 2021-08-12
Clean up cputopology initialization global communication pattern

evan-charmworks:
- Replaces error-prone locking and improper scheduler driving with well-tested pattern from Isomalloc-Sync
- Replaces bespoke "empty reduction" with CmiBarrier
- Replaces Broadcast with NodeBroadcast and moves send outside of handler
- Replaces unnecessary privatization of Converse handler index variables with CmiAssignOnce

### DIFF
--- a/src/conv-core/cputopology.C
+++ b/src/conv-core/cputopology.C
@@ -196,10 +196,10 @@ static void printTopology(int numNodes)
     CmiPrintf("Charm++> Running on %d hosts\n", numNodes);
 }
 
-static std::atomic<char> cpuTopoSyncHandlerDone;
+static std::atomic<bool> cpuTopoSyncHandlerDone{};
 #if CMK_SMP && !CMK_SMP_NO_COMMTHD
 extern void CommunicationServerThread(int sleepTime);
-static std::atomic<char> cpuTopoSyncCommThreadDone;
+static std::atomic<bool> cpuTopoSyncCommThreadDone{};
 #endif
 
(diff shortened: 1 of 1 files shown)


evan-charmworks:
@trquinn Does this resolve the hang on init from #2850?

ericjbohm (dismissed):
Commit comment should note motive for switching from locked Cpv to atomic.

evan-charmworks:
> Commit comment should note motive for switching from locked Cpv to atomic.

Updated.

src/conv-core/cputopology.C
```diff
@@ -198,6 +196,32 @@ static void printTopology(int numNodes)
     CmiPrintf("Charm++> Running on %d hosts\n", numNodes);
 }
 
+static std::atomic<char> cpuTopoSyncHandlerDone;
+#if CMK_SMP && !CMK_SMP_NO_COMMTHD
+extern void CommunicationServerThread(int sleepTime);
+static std::atomic<char> cpuTopoSyncCommThreadDone;
```
matthiasdiener, line 202: Is there a reason this is not a `std::atomic<bool>`? Should this be explicitly initialized? (same  question for cpuTopoSyncHandlerDone)

evan-charmworks: Done

matthiasdiener (approved):

evan-charmworks:
Needs one more review.

ericjbohm (approved):

PR #3453, opened 2021-08-17
Fix LGTM alert for array element proxy assignment operator

jszaday:


stwhite91 (approved):

evan-charmworks:
Could you change the PR name and squashed commit title to succinctly say what the fixed warning is? Also, commit titles should use `Sentence case` and not `Title Case`.

evan-charmworks (approved):

PR #3455, opened 2021-08-17
hwloc: Specify full set of compiler options to configure script

### LINKED ISSUE #3441: buildold fails on machines without /lib/cpp
As reported by @trquinn:

> v7.0.0-rc1 doesn't compile on a machine without /lib/cpp.  hwloc seems to cause the problem.
> v6.11.0-beta1 does compile.
> The final error is:
> ```
> configure:13996: error: in `/home/trq/src/charm/mpi-linux-x86_64-smp-mpicxx/tmp/hwloc':
> configure:13998: error: C++ preprocessor "/lib/cpp" fails sanity check
> See `config.log' for more details
> ```
> Here is the bad commit according to git bisect: b55cc8ca7a3295b2c474c9a3e6213f5a42b95a09

It would be useful to know what `tmp/hwloc/config.log` says. If buildold fails, it should print the contents of this file to stdout. It would be useful to know the exact command and input is used for the test that fails, as well as for the test preceding it for the C preprocessor as opposed to C++. The environment variables `$CPP`, `$CPPFLAGS`, and `$CXXCPP` would also be good to know.

evan-charmworks:
Fixes #3441

### DIFF
--- a/cmake/hwloc.cmake
+++ b/cmake/hwloc.cmake
@@ -33,8 +33,15 @@ ExternalProject_Add(hwloc
         --without-x \
         HWLOC_FLAGS=\"$CHARM_CC_FLAGS\" \
         CFLAGS=\"$CHARM_CC_FLAGS\" \
-        CC=$CHARM_CC \
-        CC_FOR_BUILD=$CHARM_CC \
+        CXXFLAGS=\"$CHARM_CXX_FLAGS\" \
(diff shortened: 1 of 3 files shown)


matthiasdiener (approved):

trquinn (approved):
This branch builds on the machine I was previously having trouble with.

rbuch (approved):

PR #3462, opened 2021-08-25
Avoid hang during Isomalloc-Sync

evan-charmworks:


evan-charmworks:
CPU affinity has now been cleaned up, resolving the CI hang introduced by the first commit. @trquinn Ready for testing.

trquinn (changes requested):
It still hangs.

evan-charmworks:
> It still hangs.

Okay, I can reproduce this, but now it happens at program exit instead of during initialization. I will debug further.

evan-charmworks:
There are still two hangs.

One is somewhere in the implementation of `CmiNodeReduceStruct`. I have verified that Isomalloc-Sync does everything correctly, but the reduction never completes and calls `CmiIsomallocSyncReductionHandler`. `CmiAddressSpaceRegionMerge` and `CmiAddressSpaceRegionPup` are called fewer times than normal before the hang (6 vs. 8, 60 vs. 62; `+n32 ++ppn 4`).

The other is at exit. The worker threads on most nodes are in the `CmiYield` loop in `realConverseExit`, while the comm thread is stuck in an `MPI_Barrier` inside the block of code that handles exiting. I still need to find what node is not participating in the barrier and where.

It's rare to reproduce them, taking a few minutes of iterations before hitting, and it's a coin flip for which hang is encountered.

evan-charmworks:
@trquinn I fixed a data race in the CmiNodeReduce API that Isomalloc-Sync uses and have not seen any of the hangs or crashes since. Ready for another round of testing.

trquinn (approved):
It doesn't hang!

stwhite91 (approved):
Avoid squashing at merge, or else make the commit message more descriptive.

Good work tracking this down Evan, and thanks Toms for the testing!

PR #3467, opened 2021-08-31
OFI: Change return type of registerDirectMemory to void

nitbhat:
This commit fixes the ofi autobuild crashes seen on Stampede2.
Previously, registerDirectMemory was defined to return a fid_mr
struct pointer but did not return any argument. This
caused a mysterious crash with SIGSEGV on Stampede2 with gcc 9.1.0.
This commit changes the return type to void, which is the correct
return type for that function, and this in turn fixes the crash.

### DIFF
--- a/src/arch/ofi/machine-onesided.C
+++ b/src/arch/ofi/machine-onesided.C
@@ -1,12 +1,12 @@
 void registerDirectMemory(void *info, const void *addr, int size) {
   CmiOfiRdmaPtr_t *rdmaInfo = (CmiOfiRdmaPtr_t *)info;
   uint64_t requested_key = 0;
-  int ret;
+  int err;
 
   if(FI_MR_SCALABLE == context.mr_mode) {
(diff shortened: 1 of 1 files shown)


src/arch/ofi/machine-onesided.C
```diff
@@ -1,4 +1,4 @@
-struct fid_mr* registerDirectMemory(void *info, const void *addr, int size) {
+void registerDirectMemory(void *info, const void *addr, int size) {
   CmiOfiRdmaPtr_t *rdmaInfo = (CmiOfiRdmaPtr_t *)info;
   uint64_t requested_key = 0;
   int ret;
```
stwhite91, line 4: Would be better to change 'ret' to 'err' or something more fitting now

nitbhat: Yes, I think that's a good idea. Done!

evan-charmworks (approved):

matthiasdiener (approved):

PR #3475, opened 2021-09-02
charmc: Remove assumptions about compilation output file extensions

evan-charmworks:
Recently, HYPRE changed their build system to output object files with `.obj` extensions instead of `.o`. This broke building with AMPI because charmc made assumptions about object files having `.o` extensions and files with `.o` extensions being object files. These assumptions are completely invalid because toolchains do not care about file extensions, so remove them entirely and instead test for the presence of the `-c` flag where needed.


stwhite91 (approved):

stwhite91 (approved):

jszaday:
I have always been curious about why `charmc` behaved this way -- so I support these changes. That said, I have concerns about how this will affect the build scripts of our collaborators given its widespread effects on our own tests. I wonder whether we should maintain a bit of detection for `-o *.o` and print a warning like, "WARNING: Charm++ 7.1 changes Charmxi's default for *.o files; if you mean to build an object-file, please add `-c`".

evan-charmworks:
@jszaday I added warnings, thanks for the suggestion.

stwhite91 (approved):

matthiasdiener (approved):

PR #3476, opened 2021-09-02
AMPI: Try `main` as an entry point with fs/pip/pieglobals

evan-charmworks:
This allows programs that do not include mpi.h before defining `main` to
run successfully with these methods, such as in some configure tests.

It also allows running Fortran programs with these methods without
renaming the `program` statement to `subroutine mpi_main`.


matthiasdiener (approved):

stwhite91 (approved):
Great idea, we could change the documentation saying that you must rename main, though it is dependent on fs/pie/pip-globals

PR #3481, opened 2021-09-14
Converse: Strengthen locking in node reductions

evan-charmworks:
It is possible for node reduction messages to arrive at the root before
it has completed its own CmiNodeReduce* call. Therefore, lock around
all accesses of _nodereduce_info.

### DIFF
--- a/src/conv-core/convcore.C
+++ b/src/conv-core/convcore.C
@@ -2736,7 +2736,6 @@ void CmiResetGlobalReduceSeqID(void) {
 }
 
 void CmiResetGlobalNodeReduceSeqID(void) {
-	CmiAssert(CmiMyRank() == 0);
 	CsvAccess(_nodereduce_seqID_global) = 0;
 }
 
@@ -2894,8 +2893,6 @@ void CmiGroupReduceStruct(CmiGroup grp, void *data, CmiReducePupFn pupFn,
 }
 
 void CmiNodeReduce(void *msg, int size, CmiReduceMergeFn mergeFn) {
-  CmiAssert(CmiMyRank() == 0);
-
   const CmiReductionID id = CmiGetNextNodeReductionID();
 #if CMK_SMP
   CmiNodeReduction & nodered = CsvAccess(_nodereduce_info)[CmiGetRedIndex(id)];
@@ -2913,8 +2910,6 @@ void CmiNodeReduce(void *msg, int size, CmiReduceMergeFn mergeFn) {
 void CmiNodeReduceStruct(void *data, CmiReducePupFn pupFn,
                          CmiReduceMergeFn mergeFn, CmiHandler dest,
                          CmiReduceDeleteFn deleteFn) {
-  CmiAssert(CmiMyRank() == 0);
-
   const CmiReductionID id = CmiGetNextNodeReductionID();
 #if CMK_SMP
   CmiNodeReduction & nodered = CsvAccess(_nodereduce_info)[CmiGetRedIndex(id)];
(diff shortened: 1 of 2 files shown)


src/conv-core/convcore.C
```diff
@@ -2888,27 +2893,65 @@ void CmiGroupReduceStruct(CmiGroup grp, void *data, CmiReducePupFn pupFn,
 }
 
 void CmiNodeReduce(void *msg, int size, CmiReduceMergeFn mergeFn) {
-  CmiReduction *red = CmiGetNextNodeReduction(CmiNumNodeSpanTreeChildren(CmiMyNode()));
+  const CmiReductionID id = CmiGetNextNodeReductionID();
```
rbuch, line 2896: Does the call to `CmiGetNextNodeReductionID()` need to be inside the lock, too? Won't the reads and writes to `_nodereduce_seqID_global` race? I see that `id` is used to get the right lock, so perhaps I'm missing something here.

evan-charmworks: `CmiGetNextNodeReductionID()` is only called from `CmiNodeReduce()` and `CmiNodeReduceStruct()`, which are documented as only valid to call on rank 0. I've added assertions to that effect.

src/conv-core/convcore.C
```diff
@@ -2510,19 +2514,19 @@ static CmiReduction* CmiGetReductionCreate(int id, short int numChildren) {
...
   auto red = redref;
   redref = nullptr;
   free(red);
 }
 
-static CmiReduction* CmiGetNextReduction(short int numChildren) {
-  int id = CpvAccess(_reduce_seqID_global);
-  int newid = id + CmiReductionID_multiplier;
-  if (id > 0xFFF0) newid = CmiReductionID_globalOffset;
+static CmiReductionID CmiGetNextReductionID() {
```
jszaday, line 2523: Minor nitpick, file convention seems to be `(void)` vs. `()`.

evan-charmworks: That's a holdover from when this file was `convcore.c`. I'm not eager to add the `void` because the file is C++ and the function has no declaration in a header that could be included from C. There are other inconsistencies in this file too, like whether pointers and references have a space to the right of their respective symbols.

evan-charmworks: I'll fix it anyway since I'm having to make more changes to this PR.

src/conv-core/convcore.C
evan-charmworks: That ended up being incorrect. These may be called from any rank so long as the node only calls them once. I've made `_nodereduce_seqID_global` `std::atomic` to compensate.

src/conv-core/convcore.C
```diff
@@ -2471,15 +2483,16 @@ CpvStaticDeclare(CmiReductionID, _reduce_seqID_request);
...
+CsvStaticDeclare(CmiNodeReductionID, _nodereduce_seqID_request);
+CsvStaticDeclare(CmiNodeReductionID, _nodereduce_seqID_dynamic);
 
 enum : CmiReductionID {
   CmiReductionID_globalOffset = 0, /* Reductions that involve the whole set of processors */
   CmiReductionID_requestOffset = 1, /* Reductions IDs that are requested by all the processors (i.e during intialization) */
   CmiReductionID_dynamicOffset = 2, /* Reductions IDs that are requested by only one processor (typically at runtime) */
-  CmiReductionID_multiplier = 3
+
+  CmiReductionID_multiplier = 4 /* MUST be a power of two because the ID counter may overflow and wrap to 0 */
```
rbuch, line 2495: Maybe add a `static_assert` to check this at compile time?

rbuch (dismissed):

evan-charmworks:
This is needed to fix autobuild.

stwhite91 (approved):

evan-charmworks:
I set autobuild to run on this branch and issues seen previously are resolved.

matthiasdiener (approved):

evan-charmworks:
Something went wrong with LGTM, like it didn't receive any notification from GitHub when I updated the branch from the web interface, and it's blocking the merge. I can't find anything to help with this on lgtm.com.

matthiasdiener:
> Something went wrong with LGTM, like it didn't receive any notification from GitHub when I updated the branch from the web interface, and it's blocking the merge. I can't find anything to help with this on lgtm.com.

I made LGTM optional for now.

PR #3484, opened 2021-09-23
Channel API in Charm++ for direct GPU-GPU communication with UCX

minitu:
Implements the Channel API in Charm++ that allows the user to create a channel between a pair of chares and send/receive data using that channel (MPI-like).
This aims at reducing the overheads from sending an extra metadata message & delaying the posting of the receive for incoming GPU data in the Zero Copy API based mechanism, by exchanging only data and not the flow of execution.
This could be seen as trading off message-driven execution for better communication performance.


minitu:
Moving this to #3507.

PR #3497, opened 2021-10-11
Add support for MPI Cray Shasta build

rbuch:
Tested with Cray, AMD (both based on `clang`), and GNU compilers on OLCF's testbed machine Spock. I'll add OFI support as another PR, that'll require additional work.


evan-charmworks:
Maybe it's worth adding autodetection to `smart-build.pl` too?

rbuch:
> Maybe it's worth adding autodetection to `smart-build.pl` too?

I think I'll add autodetection for the `ofi-crayshasta` build instead (which I'm working on now), since that should be the preferred layer.

PR #3515, opened 2021-10-26
ZC Post API: Set pointer to nullptr before passing it to Post EM

nitbhat:
Previously, the pointer was set to the source buffer's pointer, which
is invalid when used directly on another node. For this reason, this
commit sets it to nullptr. Using the nullptr value, this commit also
adds error checking to ensure that the CkMatchBuffer call was made or
the reference to the pointer was set inside the Post Entry Method.


minitu (dismissed):

minitu:
Created another PR #3525 to handle the charmxi issue, which should be merged in before this.

minitu (approved):
Tests pass for GPU messaging.

stwhite91 (approved):

PR #3518, opened 2021-10-26
Doc: Document the new features of the ZC Post API

nitbhat:
This change describes the Match and Post API, which include
CkMatchBuffer, CkPostBuffer, CkMatchNodeBuffer, and
CkPostNodeBuffer.


stwhite91 (approved):
In the future I think it would be nice to make the post/match "tag" into a typedef that the application can specify at Charm build-time, like how we have CMK_REFNUM_TYPE now.

stwhite91 (approved):

evan-charmworks (approved):

PR #3523, opened 2021-11-09
Tracing: Use vector for Projections LogPool

rbuch:
Switches from raw buffer used via `placement new` to a `std::vector` used via `emplace_back`. Performance is probably slightly worse due to overheads, but probably worth it for safety.

### DIFF
--- a/src/ck-perf/trace-projections.C
+++ b/src/ck-perf/trace-projections.C
@@ -154,11 +154,10 @@ void LogPool::closeLog(void)
 }
 
 LogPool::LogPool(char *pgm) {
-  pool = new LogEntry[CkpvAccess(CtrLogBufSize)];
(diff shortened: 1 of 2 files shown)


stwhite91 (dismissed):

rbuch:
Converting to draft since a user reported that memory usage is increasing across the run (starting from a lower point that the original implementation, so there must be something wonky going on with reserving the capacity).

rbuch:
Converting from draft to ready; I did a bit of testing, I think the user saw that memory usage was increasing because they were tracking resident memory, not virtual memory. The `reserve` call allocates the memory upfront, but it's purely virtual until the pages that comprise it get touched.

jszaday (approved):

matthiasdiener (approved):

PR #3527, opened 2021-11-12
Reduce Copying in Node Broadcast Routines

jszaday:


src/arch/util/machine-broadcast.C
```diff
@@ -120,14 +120,20 @@ void CmiForwardMsgToPeers(int size, char *msg) {
 }
 #endif
 
+static char *_lazyCopy(char* &dst, const char *src, int size, int rankToAssign) {
```
stwhite91, line 123: Maybe better named _lazyCopyMsg?

stwhite91 (dismissed):

src/arch/util/machine-broadcast.C
rbuch: Also, it's a bit of a semantic misnomer to me, I'd expect a function called lazy copy to defer copies to point of use.

rbuch:
1. Just to check, I ran into issues when I was doing something similar to do this because of message packing; I was sending to some arbitrary list of PEs, but using the same message for local and remote destinations caused issues, so I had to do this instead:
```
    const int myNode = CkMyNode();
    std::vector<TreeLBMessage*> decisions(children.size() + 1);

    // Avoid allocating a new message until we get to a non-local child
    TreeLBMessage* remoteMsg = nullptr;

    // The first element is always for the local PE, the caller must delete it manually if
    // it is not used
    decisions[0] = decision;
    for (int i = 1; i < decisions.size(); ++i)
    {
      if (CkNodeOf(children[i-1]) == myNode)
      {
        CkReferenceMsg(decision);
        decisions[i] = decision;
      }
      // Use separate message for messages outside the process because it will get packed
      // and mess up local uses of the message
      else
      {
        if (remoteMsg == nullptr)
        {
          remoteMsg = (TreeLBMessage*)CkCopyMsg((void**)&decision);
        }
        else
        {
          CkReferenceMsg(remoteMsg);
        }
        decisions[i] = remoteMsg;
      }
    }

    return decisions;
```
Something similar may be necessary here unless you can guarantee that all destinations are all either local or remote. (This may be moot at the Converse level, I think packing happens at the Charm level, but I'd double check just in case.)

2. It might be clearer just to do a copy at the top of the function and then call `CmiReference` directly rather than through an additional function, I think the reason for an extra call to `CmiFree` would be more obvious that way.

jszaday:
> 1. Just to check, I ran into issues when I was doing something similar to do this because of message packing; I was sending to some arbitrary list of PEs, but using the same message for local and remote destinations caused issues, so I had to do this instead:
> 
> ```
>     const int myNode = CkMyNode();
>     std::vector<TreeLBMessage*> decisions(children.size() + 1);
> 
>     // Avoid allocating a new message until we get to a non-local child
>     TreeLBMessage* remoteMsg = nullptr;
> 
>     // The first element is always for the local PE, the caller must delete it manually if
>     // it is not used
>     decisions[0] = decision;
>     for (int i = 1; i < decisions.size(); ++i)
>     {
>       if (CkNodeOf(children[i-1]) == myNode)
>       {
>         CkReferenceMsg(decision);
>         decisions[i] = decision;
>       }
>       // Use separate message for messages outside the process because it will get packed
>       // and mess up local uses of the message
>       else
>       {
>         if (remoteMsg == nullptr)
>         {
>           remoteMsg = (TreeLBMessage*)CkCopyMsg((void**)&decision);
>         }
>         else
>         {
>           CkReferenceMsg(remoteMsg);
>         }
>         decisions[i] = remoteMsg;
>       }
>     }
> 
>     return decisions;
> ```
> 
> Something similar may be necessary here unless you can guarantee that all destinations are all either local or remote. (This may be moot at the Converse level, I think packing happens at the Charm level, but I'd double check just in case.)
> 
> 1. It might be clearer just to do a copy at the top of the function and then call `CmiReference` directly rather than through an additional function, I think the reason for an extra call to `CmiFree` would be more obvious that way.

1. This function is, semantically, guaranteed to only send to other address spaces (i.e., nodes) so I don't think the sharing issue is a problem here. 

2. I deferred copying until it's needed since I'm unsure whether this function is always called with at least one destination. If it's called with zero destinations (i.e., single node situation), then the copy->free would be wasteful.

stwhite91 (approved):

rbuch (approved):

PR #3530, opened 2021-12-02
Jacobi: Replace dynamically allocated objects with std::vector

### LINKED ISSUE #1314: Replace widespread dynamic allocated arrays with std::vector and std::string
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1314*

---
There are loads of places in the runtime system and our examples where the code manually allocates an array member using `malloc` or `new`. This results in the potential for leaked memory, and really ugly cumbersome code in PUP methods that have to use the `if (p.isUnpacking()) foo = new T[n]` idiom. All of this goes away if those members are just defined as `std::vector<T>`

NK-Nikunj:
This PR partially resolves #1314 by porting examples (`examples/charm++`) of Jacobi type to use `std::vector` instead of dynamically allocated arrays.

The following files are touched in this PR:
- [x] jacobi2d-1d-decomposition
- [x] jacobi2d-2d-decomposition
- [x] jacobi3d-2d-decomposition
- [x] load_balancing/stencil3d
- [x] shrink_expand/jacobi2d-iter

Random Flyby:
- [x] Resolves a warning generated by incorrect use of string format (`%d` instead of `%ld`) in `shrink_expand/jacobi2d-iter` when using `CmiMemoryUsage`

### DIFF
--- a/examples/charm++/jacobi2d-2d-decomposition/jacobi2d.ci
+++ b/examples/charm++/jacobi2d-2d-decomposition/jacobi2d.ci
@@ -17,7 +17,7 @@ mainmodule jacobi2d {
   array [2D] Jacobi {
     entry Jacobi(void);
     entry void begin_iteration(void);
-    entry void receiveGhosts(int iter, int dir, int size, double ghosts[size]);
+    entry void receiveGhosts(int iter, int dir, int size, std::vector<double> ghosts);
     entry [reductiontarget] void recvConvergeResult(bool result);
 
     entry void run() {
@@ -27,7 +27,7 @@ mainmodule jacobi2d {
         }
         for(imsg = 0; imsg < neighbors; imsg++) {
           // "iterations" keeps track of messages across steps
-          when receiveGhosts[iterations] (int iter, int dir, int size, double ghosts[size])
+          when receiveGhosts[iterations] (int iter, int dir, int size, std::vector<double> ghosts)
             serial "process ghosts" {
               processGhosts(dir, size, ghosts);
             }
--- a/examples/charm++/jacobi2d-2d-decomposition/jacobi2d.ci
+++ b/examples/charm++/jacobi2d-2d-decomposition/jacobi2d.ci
@@ -17,7 +17,7 @@ mainmodule jacobi2d {
   array [2D] Jacobi {
     entry Jacobi(void);
     entry void begin_iteration(void);
-    entry void receiveGhosts(int iter, int dir, int size, std::vector<double> ghosts);
+    entry void receiveGhosts(int iter, int dir, int size, double ghosts[size]);
     entry [reductiontarget] void recvConvergeResult(bool result);
 
     entry void run() {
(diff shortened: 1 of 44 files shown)


evan-charmworks:
1. Where possible, can you change array1d/array2d/vector_t function parameters to pass by const reference in the C++ instead of by value, to avoid unnecessary copies? I believe the .ci files should remain unchanged.

2. Can the `tmp = temperature; temperature = new_temperature; new_temperature = tmp;` constructs be replaced with `std::swap`?

NK-Nikunj:
1. I should've checked if the functions were marked [entry]. You're right, let me mark local calls to `const&`.
2. `std::swap` is conceptually the same as 3 moves. When you `std::move` to another vector all you do is copy the 3 pointers of the source vector and assign them to the target vector (so that's O(1) operations). So, both semantics will perform at a similar level (see [this](https://www.quick-bench.com/q/BWfmZqVVZZxBeGu_L-AON-arSkw)). But you're right, for readability purposes, using `std::swap` makes more sense.

evan-charmworks:
> 1. I should've checked if the functions were marked [entry]. You're right, let me mark local calls to `const&`.

Even for entry methods, you should still try the change, because it will save a copy between the runtime's generated code and your code.

> 2. `std::swap` is conceptually the same as 3 moves. When you `std::move` to another vector all you do is copy the 3 pointers of the source vector and assign them to the target vector (so that's O(1) operations). So, both semantics will perform at a similar level (see [this](https://www.quick-bench.com/q/BWfmZqVVZZxBeGu_L-AON-arSkw)). But you're right, for readability purposes, using `std::swap` makes more sense. I didn't do it to keep the changes to a minimum.

Yes, that makes sense. Since the lines of source code in question are already changing in this PR, I would recommend upgrading them to use the proper `std::swap` idiom, which serves the overall goal of the changeset, and is just as clear, if not clearer, than three successive `std::move` calls.

examples/charm++/jacobi2d-2d-decomposition/jacobi2d.ci
```diff
@@ -17,7 +17,7 @@ mainmodule jacobi2d {
   array [2D] Jacobi {
     entry Jacobi(void);
     entry void begin_iteration(void);
-    entry void receiveGhosts(int iter, int dir, int size, double ghosts[size]);
+    entry void receiveGhosts(int iter, int dir, int size, std::vector<double> ghosts);
```
jszaday, line 20: Sending an `std::vector` via entry method call will result in generally worse performance; it results in an unnecessary/unavoidable copy. It would be better to preserve `receiveGhosts` as it is, and simply send the `.data()` of the `std::vector` in user-code.

evan-charmworks (approved):
Thanks for looking into my suggestions, Nikunj.

Before merge, could you change "object" to "objects" in the PR title and add a "Jacobi:" prefix at the beginning?

jszaday (approved):

PR #3536, opened 2021-12-09
Add support for Intel oneAPI DPC++/C++ compiler

### LINKED ISSUE #3535: Add support for Intel oneAPI DPC++/C++ compiler
Builds for using DPC++ on Intel GPU platforms (e.g., Intel DevCloud) currently uses `clang` (e.g., `./build charm++ multicore-linux-x86_64 clang -j --with-production`), but there is word this might no longer work in the near future.
`dpcpp` should be enabled to be used as the compiler for building Charm++.

Notes from ALCF Aurora Workshop:
> If invoking "clang" or "clang++" directly, please try to use icx/icpx/dpcpp instead. Direct use of clang as a driver is not supported. The "clang" drivers do not have the same behavior as either icx or upstream LLVM, and are not widely tested.

minitu:
Uses `icx` as the C compiler since `dpcpp` is a C++ compiler. They both use `clang` underneath.
Tested on Intel DevCloud.

Closes #3535.


src/arch/common/cc-dpcpp.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ "$CMK_COMPILER" = "msvc" ]; then
+  CMK_AR='ar q'
+  CMK_LIBS='-lws2_32 -lpsapi -lkernel32'
+  CMK_SEQ_LIBS="$CMK_LIBS"
+
+  CMK_NATIVE_CC="$CMK_CC"
+  CMK_NATIVE_LD="$CMK_LD"
+  CMK_NATIVE_CXX="$CMK_CXX"
+  CMK_NATIVE_LDXX="$CMK_LDXX"
+fi
```
rbuch, line 19: Will this block ever get hit? Is it meaningful since line 24 sets it to `icx` anyway?

minitu: Maybe on Windows (I think that's where `msvc` is used), this file has been mostly copied over from `cc-clang.sh` since `dpcpp` uses `clang` underneath.
We can remove it for now and revisit it when we have the need for dpcpp builds on Windows.

jszaday (commented):
Intel is viewing ICC as "deprecated" with the dawn of ICX, so this PR is prudent.

One quick question -- do you know which debugger is preferred for ICX? I always run into the issue manually changing charmrun's definition of `DEBUGGER` to `gdb-oneapi` with ICC, and I'm curious whether there's a corresponding `lldb-oneapi`, etc.

rbuch (approved):
Looks good to me, assuming it tests cleanly on with these new compilers. (Can we add them to CI or are there availability/licensing issues?)

minitu:
@jszaday I think it's still gdb-oneapi, that is the only debugger I could find on DevCloud.
@rbuch Charm++ built cleanly with icx and dpcpp on DevCloud. To add this to the CI, I think we need the whole oneAPI base toolkit (https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit.html) which is about 3 GB.

jszaday:
> @jszaday I think it's still gdb-oneapi, that is the only debugger I could find on DevCloud. 
I see -- in that case, can we update `charmrun` to use that for this build? Manually editing it to make `charmrun ++debug` work correctly can be annoying.

minitu:
@jszaday I've never used a debugger with `charmrun` - do you know if it is sufficient to add another `arg_debugger` for the DPC++ compiler after line 1052 in `src/util/charmrun-src/charmrun.C`?

jszaday:
> @jszaday I've never used a debugger with `charmrun` - do you know if it is sufficient to add another `arg_debugger` for the DPC++ compiler after line 1052 in `src/util/charmrun-src/charmrun.C`?

On second thought, since ICX targets LLVM, it's likely preferable to use `lldb`. Intel provided `gdb-oneapi` since it modified GCC's code generation vis-a-vis ICC; however, LLVM is somewhat "immutable" in this regard.

I cannot find much documentation about Intel's preferred debugger for ICX. For now, we can leave it as `lldb` and revisit this matter if new evidence presents itself.

jszaday (approved):

PR #3537, opened 2021-12-10
CcdCallbacks: only call timers when necessary

stwhite91:
Added CcdCondFn to vaoid timers in condition-based callbacks and
changed the return type of CcdRaiseCondition to enable only calling
timers when necessary, decreasing the scheduling overhead. Also only
call CcdCallBacks() when there are periodic callbacks registered.

Define a macro for backwards compatibility, i.e. users can check
"#if CCD_COND_FN_EXISTS" in application code.


rbuch (approved):

jszaday (approved):

evan-charmworks (approved):

jszaday (approved):

rbuch (dismissed):

stwhite91:
This latest commit changes the type of condition-based (vs. periodic) callbacks so that we can avoid timer calls entirely for them. This requires changing the function signature of those callbacks, so merging this will require applications (such as NAMD) to update their code.

stwhite91:
I introduced an error copy/pasting Cpv variable names in conv-conds. Now that that is fixed I'm hopeful all builds will pass except NAMD, which will need to be updated to use CcdCondFn instead of CcdVoidFn.

rbuch:
> I introduced an error copy/pasting Cpv variable names in conv-conds. Now that that is fixed I'm hopeful all builds will pass except NAMD, which will need to be updated to use CcdCondFn instead of CcdVoidFn.

I've created an MR for the change in NAMD here: https://gitlab.com/tcbgUIUC/namd/-/merge_requests/101

src/conv-core/conv-conds.C
```diff
@@ -10,21 +10,32 @@ CpvDeclare(double, idleBeginWalltime); // used for determining the conditon for
...
   int pe;			/* the pe that sets the callback */
-} ccd_callback;
+} ccd_periodic_callback;
 
 
 
 /**
  * An element (a single callback) in a list of callbacks
  */
 typedef struct _ccd_cblist_elem {
```
rbuch, line 37: Maybe the typename of this struct (and for `ccd_heap_elem`) should now reflect that they only accept callback of their respective types? Not a big deal, though.

rbuch (approved):

stwhite91:
The corresponding patch to NAMD has been merged, so this is ready for final review and merge now.

jszaday (approved):

stwhite91:
UCX build failure was unrelated to this patch:

E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libe/libevent/libevent-extra-2.1-7_2.1.11-stable-1_amd64.deb  Connection failed [IP: 52.250.76.244 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

PR #3540, opened 2021-12-15
Add LB Position API

rbuch:
- LB: Add position API
- LB: Add ORBLB


src/ck-ldb/TreeLevel.h
```diff
@@ -1065,19 +1070,25 @@ class PELevel : public LevelLogic
     if (nobjs > 0) lbmgr->GetObjData(allLocalObjs.data());  // populate allLocalObjs
     myObjs.clear();
     LBRealType nonMigratableLoad = 0;
+    size_t minPosDimension = std::numeric_limits<size_t>::max();
```
slm960323, line 1073: use a mixture of **size_t** and **unsigned int** for min/maxPosDimensions among this PR

slm960323 (approved):
Left a comment for some inconsistency of variable type.

src/ck-ldb/TreeLevel.h
rbuch: Should be resolved in 2306daf.

PR #3554, opened 2022-01-12
AMPI: Privatize Fortran runtime libraries under globals methods

evan-charmworks:
Tested with gfortran and ifort.
May resolve OFI autobuild failure.


matthiasdiener (approved):

stwhite91 (approved):
What are the implications for users? Anything that should be documented?

evan-charmworks:
> What are the implications for users? Anything that should be documented?

I don't anticipate any drawbacks, only improved turnkey functionality. This may resolve #2932 under the globals methods.

PR #3557, opened 2022-01-13
Fix for broken AMPI LB statistics

jszaday:


stwhite91 (dismissed):

rbuch (approved):
Are these needed just for AMPI or does this issue exist in tcharm more generally?

stwhite91:
> Are these needed just for AMPI or does this issue exist in tcharm more generally?

Should be TCharm generally, so the comment could be updated to that

rbuch (approved):

evan-charmworks (approved):

PR #3567, opened 2022-02-07
AMPI: pieglobals updates for migration under Fortran

evan-charmworks:


stwhite91 (approved):

evan-charmworks:
I set autobuild to use this branch last night, and it did not resolve the errors seen in the OFI build, and also introduced a new error with PiPglobals-f90 on i386. I did not notice until now that the OFI build fails in PIEglobals with the C++ test too, so more debugging is needed.

PR #3571, opened 2022-02-23
Fix non-marshaled inline entry methods

jszaday:
This PR fixes `[inline]` non-marshaled entry methods. They do not build currently since Charmxi generates code that attempts to size messages using PUP. @NK-Nikunj demonstrated this problem with this example:
https://gist.github.com/NK-Nikunj/41bb17c37a4f9199405ded106589aeb3

This PR simultaneously ensures that `[nokeep, inline]` entry methods behave as expected (i.e., they call `CkFreeMsg`).


stwhite91 (approved):

NK-Nikunj (approved):
It works for me locally. Thanks for looking into it.

rbuch (approved):

PR #3578, opened 2022-03-03
Allow Multiple System Libraries in Standalone Build

ZwFink:
The current use of CMK_SYSLIBS doesn't allow multiple values in the standalone build. For example, currently ```zlib``` and ```shmem``` cannot be used together. This patch fixes this issue by correcting the way the variable is accessed in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -283,14 +283,14 @@ endif()
 
 if(${CMK_USE_ZLIB})
   find_package(ZLIB REQUIRED)
-  set(CMK_SYSLIBS "$CMK_SYSLIBS -lz")
+  set(CMK_SYSLIBS "${CMK_SYSLIBS} -lz")
 endif()
 
 if(${CMK_USE_SHMEM})
   if(${CMK_HAS_XPMEM})
-    set(CMK_SYSLIBS "$CMK_SYSLIBS -lxpmem")
+    set(CMK_SYSLIBS "${CMK_SYSLIBS} -lxpmem")
   elseif(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    set(CMK_SYSLIBS "$CMK_SYSLIBS -lrt")
+    set(CMK_SYSLIBS "${CMK_SYSLIBS} -lrt")
   endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,16 +285,19 @@ endif()
 
 if(${CMK_USE_ZLIB})
   find_package(ZLIB REQUIRED)
-  set(CMK_SYSLIBS "$CMK_SYSLIBS ${CMK_SYSLIBS} -lz")
+  set(CMK_SYSLIBS "${CMK_SYSLIBS} -lz")
 endif()
 
 if(${CMK_USE_SHMEM})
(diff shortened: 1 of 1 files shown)


jszaday (approved):

minitu (approved):

CMakeLists.txt
```diff
@@ -283,14 +283,14 @@ endif()
 
 if(${CMK_USE_ZLIB})
   find_package(ZLIB REQUIRED)
-  set(CMK_SYSLIBS "$CMK_SYSLIBS -lz")
+  set(CMK_SYSLIBS "${CMK_SYSLIBS} -lz")
```
matthiasdiener, line 286: I don't think this is correct. I think we want to (also) access the bash variable `$CMK_SYSLIBS` here. Maybe:

```suggestion
  set(CMK_SYSLIBS "${CMK_SYSLIBS} $CMK_SYSLIBS -lz")
```

(same below)

matthiasdiener (changes requested):

CMakeLists.txt
matthiasdiener: (just to clarify, `$CMK_SYSLIBS` will be accessed through `charmc`)

rbuch: Can you explain a bit more why it's necessary to access this variable twice? I'm a bit confused by this issue in general, I thought the fact that the variable is quoted would be sufficient.

matthiasdiener: > Can you explain a bit more why it's necessary to access this variable twice? I'm a bit confused by this issue in general, I thought the fact that the variable is quoted would be sufficient.

If I understand correctly, the trick here is that CMK_SYSLIBS are two variables, one for bash (`$CMK_SYSLIBS`), and one for CMake (`${CMK_SYSLIBS}`). The generated files *need* to reference `$CMK_SYSLIBS` since that gets also set in other shell files of the build system. 

Maybe it is better to write the code as follows:

```cmake
set(CMK_SYSLIBS "${CMK_SYSLIBS} -lz") # same for the other places where CMK_SYSLIBS gets set 
[...]
# after final modification of CMK_SYSLIBS
set(CMK_SYSLIBS "$CMK_SYSLIBS ${CMK_SYSLIBS}")
```

CMakeLists.txt
```diff
@@ -285,17 +285,20 @@ endif()
...
   if(${CMK_HAS_XPMEM})
-    set(CMK_SYSLIBS "$CMK_SYSLIBS -lxpmem")
+    set(CMK_SYSLIBS "${CMK_SYSLIBS} -lxpmem")
   elseif(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    set(CMK_SYSLIBS "$CMK_SYSLIBS -lrt")
+    set(CMK_SYSLIBS "${CMK_SYSLIBS} -lrt")
   endif()
 endif()
 
+# Ensure parity of bash and CMake variables
```
matthiasdiener, line 299:
```suggestion
# $CMK_SYSLIBS also gets updated in the charmc shell config files, therefore we need to include
# a reference to $CMK_SYSLIBS here.
```

rbuch (approved):

matthiasdiener (approved):

PR #3582, opened 2022-03-19
Fix detection of `offsetof` in Clang/ICX builds

jszaday:
The previous detection method for `offsetof` was incompatible with Clang/ICX. I noticed this while using [parallel-hashmap](https://github.com/greg7mdp/parallel-hashmap), since our redefinition of `offsetof` is not `constexpr` safe and interferes with its build. Thanks to @matthiasdiener for this suggestion!


matthiasdiener (approved):
Thanks!

stwhite91 (approved):

PR #3583, opened 2022-03-19
Remove CDash integration

matthiasdiener:
This has been dormant for a year now. CDash does not work well for us since most systems that we are interested in do not seem to support execution via cron. A better solution might be to use a github-actions runner.


rbuch (dismissed):

matthiasdiener:
This needs another review.

rbuch (dismissed):

jszaday (dismissed):

PR #3586, opened 2022-03-24
Remove need for +LBCommOff

kavithachandrasekar:
Change-Id: I7d3fb02c7dd389b4e45743497b8ecdb91e844d2f


src/ck-ldb/LBManager.C
```diff
@@ -484,7 +485,7 @@ void LBManager::initnodeFn()
   _registerCommandLineOpt("+LBSameCpus");
   _registerCommandLineOpt("+LBUseCpuTime");
   _registerCommandLineOpt("+LBOff");
-  _registerCommandLineOpt("+LBCommOff");
```
stwhite91, line 487: We may want to leave the +LBCommOff option as well (as a no-op) so that if anyone has a script passing it it will still get parsed out by the runtime and not passed through to the application with a warning about 'unknown CLA beginning with a +'

src/ck-ldb/LBManager.C
```diff
@@ -397,9 +397,10 @@ void _loadbalancerInit()
   _lb_args.statsOn() =
       !CmiGetArgFlagDesc(argv, "+LBOff", "Turn load balancer instrumentation off");
 
-  // turn instrumentation of communicatin off at startup
-  _lb_args.traceComm() = !CmiGetArgFlagDesc(
-      argv, "+LBCommOff", "Turn load balancer instrumentation of communication off");
+  // turn instrumentation of communicatin on at startup
```
rbuch, line 400: Might as well fix this typo, too.
```suggestion
  // turn instrumentation of communication on at startup
```

stwhite91 (approved):

rbuch (approved):

PR #3588, opened 2022-03-26
Add tests useful for load balancing

PathikritGhosh:


examples/charm++/load_balancing/kNeighbor/kNeighbor.C
```diff
@@ -238,6 +305,13 @@ class Block: public CBase_Block {
       p(curIterMsgSize);
       p(internalStepCnt);
       p(sum);
+      p(msg_var);
+      p(compute_var);
+      p(compute_work);
```
rbuch, line 310: Minor, but it'd be good to maintain the same ordering as these are declared and initialized in (`compute_work` and then `compute_var`).

examples/charm++/load_balancing/kNeighbor/kNeighbor.C
```diff
@@ -265,14 +339,17 @@ class Block: public CBase_Block {
...
+
       for (int i=0; i<N; i++)
 	for (int j=0; j<N; j++) {
-	  sum += (thisIndex * i + j);
+	  sum += (thisIndex * i + j)%100;
 	}
 
       /* 2. send msg to K neighbors */
-      int msgSize = curIterMsgSize;
+      int msgSize = curIterMsgSize*(rand()%msg_var + 1);
```
rbuch, line 352: Where is this variable used? I don't see it anywhere.

tests/charm++/jacobi3d/jacobi3d.C
```diff
@@ -99,14 +103,15 @@ class Main : public CBase_Main {
 	std::vector<std::pair<double,int> > times;
 
     Main(CkArgMsg* m) {
-      if ( (m->argc != 3) && (m->argc != 7) && (m->argc != 5) && (m->argc != 9) ) {
+      if ( (m->argc != 3) && (m->argc != 7) && (m->argc != 5) && (m->argc != 9) && (m->argc != 10) ) {
```
rbuch, line 106: Shouldn't the addition of `msg_overhead` add an `m->argc != 8` case, too (when `maxIter` and `ckptFreq` are not specified)?

tests/charm++/jacobi3d/jacobi3d.C
```diff
@@ -256,16 +275,18 @@ class Jacobi: public CBase_Jacobi {
 		// pupping properties of this class
 		p | iterations;
 		p | imsg;
+                p | msg_overhead;
```
rbuch, line 278: Indentation (here and below, in lines 284 and 289).

tests/charm++/jacobi3d/jacobi3d.C
```diff
@@ -217,6 +232,9 @@ class Jacobi: public CBase_Jacobi {
     double *temperature;
     double *new_temperature;
 
+    int msg_overhead;
+    char* temp;
```
rbuch, line 236: I think these names could be more descriptive, maybe something like `msg_overhead_size` and `msg_overhead_data` or similar.

examples/charm++/load_balancing/kNeighbor/kNeighbor.C
```diff
@@ -265,14 +339,17 @@ class Block: public CBase_Block {
 
       numNborsRcvd = 0;
       /* 1: pick a work size and do some computation */
-      int N = (thisIndex * thisIndex / num_chares) * 100;
+      //TODO
+      int N = (thisIndex * thisIndex / num_chares) * compute_work * (rand()%compute_var + 1) + l_factor*l_frequency;
```
rbuch, line 343: It's hard to figure out what this does, a comment and some renaming of variable would help (namely `compute_work` and `compute_var` are too similar).

rbuch (changes requested):

examples/charm++/load_balancing/kNeighbor/Makefile
```diff
@@ -34,3 +34,21 @@ testp: all
...
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 +balancer MetisLB +LBDebug 1)
+
+testscale: all
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer DummyLB +LBDebug 1)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer GreedyLB +LBDebug 1)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer GreedyRefineLB +LBDebug 1)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer RefineLB +LBDebug 1)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer DistributedLB +LBDebug 1)
+	$(call run, +p4 ./kNeighbor 4 50 128 10 10 10 10 1 10 5 5 +balancer MetisLB +LBDebug 1)
```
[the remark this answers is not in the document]
PathikritGhosh: I can change it, but the other tests in the Makefile have the same format(no newline character) so there will be two different ones within the same Makefile if that is okay?
I am assuming you mean to add a newline character at the end of the line?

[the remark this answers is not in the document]
rbuch: I mean just adding one at the end of the file.

tests/charm++/jacobi3d/jacobi3d.C
rbuch: Yes, but if one were to provide the parameters as printed in the message on line 108, `argc` would be 8, so it seems like that should be a supported case.

tests/charm++/jacobi3d/jacobi3d.C
```diff
@@ -137,7 +142,17 @@ class Main : public CBase_Main {
...
+	} else if (m->argc == 10) {
+                arrayDimX = atoi(m->argv[1]);
+                arrayDimY = atoi(m->argv[2]);
+                arrayDimZ = atoi(m->argv[3]);
+                blockDimX = atoi(m->argv[4]);
+                blockDimY = atoi(m->argv[5]);
+                blockDimZ = atoi(m->argv[6]);
+                maxIter = atoi(m->argv[7]);
+                ckptFreq = atoi(m->argv[8]);
+                overhead = atoi(m->argv[9]);
```
rbuch, line 154: Should `overhead` be 7 instead of 9? In the other case, the optional `maxIter` and `ckptFreq` parameters come at the end, so it's a little inconsistent for them not to be here.

tests/charm++/jacobi3d/jacobi3d.C
PathikritGhosh: My bad. Misunderstood your comment. Ideally this case was not considered, since I did not take a case where we could pass overhead without passing maxiter and checkpoint frequency as parameter. See the if else lines that follows. But will fix this in the next commit.

tests/charm++/jacobi3d/jacobi3d.C
PathikritGhosh: This is mapped correctly, the corresponding print statements before this is where this change has not been updated hence the confusion. Will update that in the next commit.

rbuch (approved):

examples/charm++/load_balancing/kNeighbor/kNeighbor.C
```diff
@@ -170,20 +222,35 @@ class Block: public CBase_Block {
...
 
   public:
     Block() {
-      //srand(thisIndex.x+thisIndex.y);
+      srand(thisIndex);
       usesAtSync = true;
 
-      numNeighbors = 2*STRIDEK;
+      numNeighbors = 2*comm_neigh;
+      msg_var = comm_var;
```
stwhite91, line 246: Are we changing these values somewhere per chare array element? If not then we can just refer directly to the readonlies

tests/charm++/jacobi3d/jacobi3d.C
```diff
@@ -238,8 +266,9 @@ class Jacobi: public CBase_Jacobi {
 		iterations = 0;
       	imsg = 0;
       	constrainBC();
-
-		usesAtSync = true;
+        msg_overhead_size = overhead;
+        temp = new char[overhead];
```
stwhite91, line 270: What is the purpose of adding temp? I don't see it used (or freed) anywhere

examples/charm++/load_balancing/kNeighbor/kNeighbor.C
PathikritGhosh: This is was done with the idea that we could change these values in the next iterations of this code. As of now, these values are not being changed.

stwhite91: Ok, can you add some comments in the code, probably where these variables are defined or initialized, explaining that these values can be changed in the future for different LB testing.

In general, I think our plan when separating out the examples directory from the tests and benchmarks directories was so that examples/ would be exemplary Charm++ programs demonstrating different features or common patterns, tests/ would be correctness-oriented programs and benchmarks/ would be performance-oriented tests.

PR #3589, opened 2022-03-28
Feature: Migratablelun

ericjbohm:
!     Handle the details of migrating LUNs across ranks
!     basic idea:
MigratableLUN is a utility to simplify migration for FORTRAN I/O

To use it the application creates the registry using createRegistry()

Application registers each LUN with the API using the same parameters it would use to open the file using registerLUN().

Application calls closeRegisteredLUNSForMigration() in about_to_migrate

Application calls reopenRegisteredLUNs() in just_migrated

Thats is all there is to it, otherwise the normal AMPI rules apply

Implementation assumes it is being used for LUNs associated with files.
YMMV if you need to migrate other LUNs.

MigratableLUN does **not** solve LUN uniqueness across virtual ranks.  It is compatible with CK_LUN, which can be used to address that.


evan-charmworks:
I rebased this from the pieglobals-testing branch onto the main branch for review.

PR #3590, opened 2022-03-30
Add functionality to transform a group/nodegroup bcast callback

epmikida:
This allows the manipulation of a group or a nodegroup bcast
callback to a local element callback targeted on the local PE for
a group bcast callback and the local node for a nodegroup bcast
callback.

### DIFF
--- a/src/ck-core/ckcallback.h
+++ b/src/ck-core/ckcallback.h
@@ -295,6 +295,18 @@ public:
           d.group.refnum = 0;
 	}
 
+  void transformBcastToLocalElem() {
+    if(type == bcastGroup) {
+      type = sendGroup;
+      d.group.onPE = CkMyPe();
+    } else if(type == bcastNodeGroup) {
+      type = sendNodeGroup;
+      d.group.onPE = CkMyNode();
+    } else {
+      CkAbort("CkCallback type needs to be either bcastGroup or bcastNodeGroup to be transformed!");
+    }
+  }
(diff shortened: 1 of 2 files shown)


epmikida:
This is required by some changes to the collision detection library we are making for Quinoa, and I think is also generally useful for passing broadcast callbacks to libraries which want to use the callback without performing the broadcast.

src/ck-core/ckcallback.h
```diff
@@ -295,6 +295,18 @@ class CkCallback {
           d.group.refnum = 0;
 	}
 
+  void transformBcastToLocalElem() {
```
jszaday, line 298: Perhaps this can be converted to:
`void transformBcastToLocalElem(int elem = -1) {`
Then, do something like:
`d.group.onPE = (elem == -1) ? CkMyPe() : elem`
and, same with:
`d.group.onPE = (elem == -1) ? CkMyNode() : elem`

Just in the off-chance they want to route it elsewhere.

rbuch (approved):
Justin's suggestion sounds like it's worth adding, but otherwise lgtm.

jszaday (approved):

stwhite91 (commented):
This should be documented in section 2.3.2 of the manual, otherwise LGTM.

https://charm.readthedocs.io/en/latest/charm++/manual.html?highlight=CkCallback#callbacks

PR #3591, opened 2022-04-01
Collide: Add a distributed client option to the collision detection library

epmikida:
The current client for the collision detection library combines all collisions to a single list on PE 0 before returning the results to the calling program. This works fine for use cases with few collisions. For use cases with many collisions, a distributed scheme avoids bottlenecks.


stwhite91 (approved):

epmikida:
I have not seen this error in CI before but it looks unrelated to the patch.

ericjbohm (approved):
Looks good, probably worth adding a note in the commit that these changes include switching from the old custom implemented reduction to the general framework.

PR #3592, opened 2022-04-09
Use mallinfo2 over mallinfo when available

rbuch:


evan-charmworks:
In the memory OS wrapper code, interception of `mallinfo2` should not affect the interception of `mallinfo`. If both of their configure tests pass, both symbols should be intercepted independently.

PR #3594, opened 2022-04-14
MPI layer: assert that Charm does not require msg ordering

stwhite91:
The MPI-4.0 standard specifies the "mpi_assert_allow_overtaking" info
which users can pass to a communicator creation call in order to tell the
MPI implementation that they don't need msg ordering.

### DIFF
--- a/src/arch/mpi/machine.C
+++ b/src/arch/mpi/machine.C
@@ -1423,9 +1423,16 @@ void LrtsInit(int *argc, char ***argv, int *numNodes, int *myNodeID) {
 
     largc = *argc;
     largv = *argv;
+
+    /* Create our communicator, with info hints (such as us not requiring
+     * message ordering) to enable possible MPI runtime optimizations. */
+    MPI_Info hints;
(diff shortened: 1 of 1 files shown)


src/arch/mpi/machine.C
```diff
@@ -1423,9 +1423,16 @@ void LrtsInit(int *argc, char ***argv, int *numNodes, int *myNodeID) {
 
     largc = *argc;
     largv = *argv;
+
+    /* Create our communicator, with info hints (such as us not requiring
+     * message ordering) to enable possible MPI runtime optimizations. */
+    MPI_Info hints;
+    MPI_Info_create(&hints);
+    MPI_Info_set(hints, "mpi_assert_allow_overtaking", "true");
     if(!CharmLibInterOperate || userDrivenMode) {
```
matthiasdiener, line 1432: Would it make sense to put this into the `if()`?

rbuch (approved):

matthiasdiener (approved):

PR #3603, opened 2022-05-04
Add options to `CkCallback::send` and fix demand-creation

jszaday:
This PR adds more correctness and customization options for calls to `CkCallback::send`.

Currently, sends to array elements via `CkCallback::send` always use a `CkArray_IfNotThere_buffer` policy. Subsequently, sends to `[createhere]` or `[createhome]` EPs do not trigger demand-creation. This PR adds an `ifNotThere` policy field to `EntryInfo` and sets it using Charmxi; `CkCallback::send` now looks up this information and sets the policy accordingly.

A simple example that would fail is:
https://gist.github.com/jszaday/573276ada1b17258caab568c1aea7463

### DIFF
--- a/src/ck-core/charm.h
+++ b/src/ck-core/charm.h
@@ -192,9 +192,9 @@ typedef enum{
 // What to do if an entry method is invoked on
 // an array element that does not (yet) exist:
 typedef enum{
-	CkArray_IfNotThere_buffer = 0,			// Wait for it to be created
-	CkArray_IfNotThere_createhere = 1,	// Make it on sending Pe
-	CkArray_IfNotThere_createhome = 2		// Make it on (a) home Pe
+	CkArray_IfNotThere_buffer = 0,     // Wait for it to be created
+	CkArray_IfNotThere_createhere = 1, // Make it on sending Pe
+	CkArray_IfNotThere_createhome = 2  // Make it on (a) home Pe
 } CkArray_IfNotThere;
 
 /** A "call function" to invoke a method on an object. See EntryInfo */
--- a/src/ck-core/charm.h
+++ b/src/ck-core/charm.h
@@ -189,6 +189,15 @@ typedef enum{
 	TypeArray
 } ChareType;
 
+// What to do if an entry method is invoked on
+// an array element that does not (yet) exist:
+typedef enum : uint8_t
+{
+  CkArray_IfNotThere_buffer = 0,      // Wait for it to be created
+  CkArray_IfNotThere_createhere = 1,  // Make it on sending Pe
+  CkArray_IfNotThere_createhome = 2   // Make it on (a) home Pe
+} CkArray_IfNotThere;
+
 /** A "call function" to invoke a method on an object. See EntryInfo */
 typedef void  (*CkCallFnPtr) (void *msg, void *obj);
 /** Register this entry point, with this call function and flags.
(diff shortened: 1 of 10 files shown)


src/ck-core/ckcallback.h
```diff
@@ -534,6 +534,27 @@ class CkCallback {
...
+    case sendNodeGroup:
+    case bcastNodeGroup:
+    case bcastGroup:
+      return d.group.ep;
+    case isendArray:
+    case sendArray:
+    case bcastArray:
+      return d.array.ep;
+    default:
+      return -1;
```
stwhite91, line 555: Will this function actually return -1 for any real case, or should it just abort like the above method?

stwhite91 (commented):
I think this should be documented in the manual

src/ck-core/ckcallback.h
jszaday: This will return `-1` for all non-EP targeting callbacks (e.g., futures) on `ckcallback.C:333`.

src/ck-core/charm.h
```diff
@@ -189,6 +189,14 @@ typedef enum{
 	TypeArray
 } ChareType;
 
+// What to do if an entry method is invoked on
+// an array element that does not (yet) exist:
+typedef enum{
+	CkArray_IfNotThere_buffer = 0,			// Wait for it to be created
+	CkArray_IfNotThere_createhere = 1,	// Make it on sending Pe
+	CkArray_IfNotThere_createhome = 2		// Make it on (a) home Pe
+} CkArray_IfNotThere;
```
rbuch, line 198: This should preserve the type that was given in the old version and should be formatted to match the surrounding context. Also, maybe make this into an `enum class`?

src/ck-core/ckcallback.C
```diff
@@ -367,69 +372,68 @@ void CkCallback::send(void *msg) const
...
-		CkSetMsgArrayIfNotThere(msg);
-		CkSendMsgArray(d.array.ep, msg, d.array.id, d.array.idx.asChild());
+		CkSetMsgArrayIfNotThere(msg, policy);
+		CkSendMsgArray(d.array.ep, msg, d.array.id, d.array.idx.asChild(), opts);
 		break;
 	case isendArray: //inline send-to-array element
 		if (!msg) msg=CkAllocSysMsg();
                 if (d.array.hasRefnum) CkSetRefNum(msg, d.array.refnum);
-		CkSendMsgArrayInline(d.array.ep, msg, d.array.id, d.array.idx.asChild());
+		CkSetMsgArrayIfNotThere(msg, policy);
```
rbuch, line 411: Why does this (and the `bcastArray` case below) now set the policy where it previously didn't? Should it have been there in the past for inline and broadcast sends but it was just never added here?

src/ck-core/charm.h
jszaday: What would you like me to change formatting-wise?

Unfortunately, the type cannot be preserved nor can this be made an `enum class` since it's all inside an `extern "C"` block. Since `CkArray_IfNotThere` should never be sent via a message, the underlying type may not matter?

src/ck-core/ckcallback.C
jszaday: Unsure, could not determine a reason; seems like a curious omission.

src/ck-core/charm.h
stwhite91: 'ifNotThere' actually has to be sent in chare array msgs so that the runtime knows what to do with messages for chare array elements that aren't created yet. It is defined as part of the message envelope as a 'uchar' in src/ck-core/envelope.h's u_type union. It looks like we're already implicitly casting from int to uchar inside envelope::setArrayIfNotThere() though

jszaday: Yes, the type-cast of the policy information to put it in the `envelope` should be safe-ish here.

Alternatively, I can move the declaration to `charm++.h`, whose contents are not confined to an `extern "C"` block.

rbuch (approved):

stwhite91 (approved):

PR #3607, opened 2022-05-10
Add support for directing node-branch sends to PEs

jszaday:
This PE adds support for directing sends to particular PEs for node-groups via `CkEntryOptions::setNodeGroupPe` and/or `CkSendMsgNodeBranchPe`. Example use-case shown here:
[https://gist.github.com/jszaday/486ee195180c42eed2a6f7b14462b123](https://gist.github.com/jszaday/486ee195180c42eed2a6f7b14462b123)

TODOs:

- [ ] Decide on names for all added methods.
- [ ] Decide whether we need support for inline/immediate sends via this new procedure? 
- [ ] Update documentation.

### DIFF
--- a/src/ck-core/ck.C
+++ b/src/ck-core/ck.C
@@ -2087,6 +2087,24 @@ void CkSendMsgNodeBranch(int eIdx, void *msg, int node, CkGroupID gID, int opts)
   CkpvAccess(_coreState)->create();
 }
 
+void CkSendMsgNodeBranchPe(int eIdx, void *msg, int pe, CkGroupID gID, int opts) {
+  // TODO ( expand support for these options at some point... )
+  CkAssertMsg(!((opts & CK_MSG_INLINE) || (opts & CK_MSG_IMMEDIATE)),
+              "pe-level inline and immediate sends to node branchs unsupported");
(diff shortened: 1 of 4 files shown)


src/ck-core/ck.C
```diff
@@ -2087,6 +2087,24 @@ void CkSendMsgNodeBranch(int eIdx, void *msg, int node, CkGroupID gID, int opts)
   CkpvAccess(_coreState)->create();
 }
 
+void CkSendMsgNodeBranchPe(int eIdx, void *msg, int pe, CkGroupID gID, int opts) {
+  // TODO ( expand support for these options at some point... )
+  CkAssertMsg(!((opts & CK_MSG_INLINE) || (opts & CK_MSG_IMMEDIATE)),
+              "pe-level inline and immediate sends to node branchs unsupported");
+  int numPes = (pe == CK_PE_ALL) ? CkNumPes() : 1;
```
jszaday, line 2094: Maybe this should fail if given `CK_PE_ALL`?

Alternatively, should it send to all _PEs_? If so, this would currently break since it `...ClEnqueue` only does node broadcasts.

PR #3612, opened 2022-06-07
Mitigate C++11 issues in charmxi

rbuch:
4fd8c4b's use of `nullptr` in charmxi code broke autobuild because the older compilers used don't support C++11 natively without a flag. This PR fixes that for `buildold` by ensuring that `configure` adds the right flag for the native compiler. CMake is a stickier wicket in that it doesn't really have any awareness of the native compiler; I've spent a couple hours trying to find a way around it, but I wasn't able to come up with anything that's not very flaky, so I've just reverted the use of `nullptr` for now to get autobuild working in the meantime (Note that CI uses newer compilers that don't suffer from this, which is how this slipped in in the first place. I've verified that this PR fixes the issue on a lab machine.).

Do you have any ideas on how to address this, @matthiasdiener? I think the ultimate solution may be to make a second CMake project specifically for configuring the native compiler environment since it may be different from the main compiler and CMake doesn't seem to support configuring multiple compilers in the same file/project.

### DIFF
--- a/src/scripts/configure.ac
+++ b/src/scripts/configure.ac
@@ -768,8 +768,8 @@ for i in "-std=c++11" "--c++11" "-h std=c++11" "-hstd=c++11"; do
     test_cxx "whether C++ compiler supports C++11 with '$i'" "yes" "no" "$i"
     if test $strictpass -eq 1
     then
-      add_flag "$(echo OPTS_CXX=\"\$OPTS_CXX $i\")" "Enable C++11 support"
-      OPTS_CXX="$OPTS_CXX $i"
+      add_flag "$(echo CMK_CXX_FLAGS=\"\$CMK_CXX_FLAGS $i\")" "Enable C++11 support"
+      CMK_CXX_FLAGS="$CMK_CXX_FLAGS $i"
       got_cpp11="true"
       break
     fi
--- a/src/xlat-i/xi-Entry.C
+++ b/src/xlat-i/xi-Entry.C
@@ -2460,7 +2460,7 @@ void Entry::genReg(XStr& str) {
   str << "  " << epIdx(0) << ";\n";
   if (isReductionTarget()) str << "  " << epIdx(0, true) << ";\n";
 
-  const char* ifNot = nullptr;
+  const char* ifNot = NULL;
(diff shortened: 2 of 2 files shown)


matthiasdiener:
I can't test this at the moment, but could you try adding the line `target_compile_options(charmxi PRIVATE ${CXX_OPTS}) ` somewhere in `src/xlat-i/CMakeLists.txt`?

rbuch:
> I can't test this at the moment, but could you try adding the line `target_compile_options(charmxi PRIVATE ${CXX_OPTS}) ` somewhere in `src/xlat-i/CMakeLists.txt`?

Adding that line still fails with the same error (just to note, this is with gcc 5.5.0 on a lab machine)
```
src/xlat-i/xi-Entry.C: In member function ‘virtual void xi::Entry::genReg(xi::XStr&)’:
src/xlat-i/xi-Entry.C:2463:23: error: ‘nullptr’ was not declared in this scope
   const char* ifNot = nullptr;
                       ^
Fatal Error by charmc in directory netlrts-linux-x86_64-smp/src/xlat-i
```

matthiasdiener:
> > I can't test this at the moment, but could you try adding the line `target_compile_options(charmxi PRIVATE ${CXX_OPTS}) ` somewhere in `src/xlat-i/CMakeLists.txt`?
> 
> Adding that line still fails with the same error (just to note, this is with gcc 5.5.0 on a lab machine)
> 
> ```
> src/xlat-i/xi-Entry.C: In member function ‘virtual void xi::Entry::genReg(xi::XStr&)’:
> src/xlat-i/xi-Entry.C:2463:23: error: ‘nullptr’ was not declared in this scope
>    const char* ifNot = nullptr;
>                        ^
> Fatal Error by charmc in directory netlrts-linux-x86_64-smp/src/xlat-i
> ```
> 
> Also, even if that were to work, wouldn't those opts be for potentially a different compiler that takes different opts?

Thank you for testing this. In this case, I'd recommend just merging this PR as it is, maybe with adding a comment to `src/xlat-i/CMakeLists.txt` that this doesn't enable C++11 currently.

stwhite91 (approved):

PR #3623, opened 2022-06-22
AMPI: turn AMPI_Migrate_to_pe off by default

stwhite91:
AMPI_Migrate_to_pe requires Charm++ anytime migration support, which
incurs overheads related to buffering chare array broadcasts. So disable
support for AMPI_Migrate_to_pe by default, allowing users to run with
+ampiEnableMigrateToPe if they want to use it.


stwhite91:
"Sphinx-doc / sphinx-build" failure appears to be spurious

rbuch (approved):
LGTM, thanks for the fix!

rbuch:
> "Sphinx-doc / sphinx-build" failure appears to be spurious

It looks like it isn't; the new version of Sphinx seems to require that the language be specified, I fixed it in https://github.com/UIUC-PPL/charm/pull/3624.

PR #3625, opened 2022-07-05
Make simplepmi the explicit default PMI

adityapb:
`simplepmi` should be the default PMI according to the documentation. Not setting this explicitly causes build failures for OFI and UCX builds on some machines (stampede2, campuscluster).


adityapb:
See #3626

PR #3626, opened 2022-07-05
Make simplepmi the explicit default PMI

adityapb:
`simplepmi` should be the default PMI according to the documentation. Not setting this explicitly causes build failures for OFI and UCX builds on some machines (stampede2, campuscluster).


rbuch:
Has this been tested on some UCX or OFI machine with `simplepmi`?

adityapb:
> Has this been tested on some UCX or OFI machine with `simplepmi`?

Yes, I've tested this on stampede2 with OFI

buildcmake
```diff
@@ -471,6 +471,11 @@ for c in "$opt_lrts_pmi" "${opt_compiler[@]}"; do
   [[ -n "$c" ]] && builddir_extra+="-$c"
 done
 
+is_ofi_ucx="+(ofi|ucx)*"
```
rbuch, line 474: I'm a bit confused by this regex, perhaps I'm misunderstanding it. What is the opening `+` supposed to do? Also, won't the `*` allow it to match against zero or more occurrences of `(ofi|ucx)`, rather than the one occurrence we are looking for? I was assuming the regex would be something like `^(ofi|ucx).*` instead.

adityapb: You're right. I was using bash's extended pattern matching syntax and didn't change it when I switched to using regex matching. I'm not sure why this is working correctly, but I'm changing the regex to what you suggested.

rbuch: Ah, gotcha, sounds good.

stwhite91 (approved):

PR #3627, opened 2022-07-27
Allow using `buildcmake` in another directory

xmjiao:
The `cmake` options are not as well documented as the `build` command. It is helpful to allow users to use the `build` command when cmaking `charm` in different folders by reusing the same source tree.


matthiasdiener (approved):
LGTM, thanks!

rbuch (approved):

Issue #3628, opened 2022-07-27
checkpoint/restart of readonlys overflow a buffer


trquinn:
The following was caught with address sanitizer running the ChaNGa "teststep":
In ckcheckpoint.C:bdcastRO(), a PUP::sizer is used to calculated the size of the RODataMessage from the _readonlyTable, but after the message is allocated an extra pup of "numZerocopyROops" is done to the PUP::toMem before going through the _readonlyTable again.  Hence the last readonly will be pupped beyond the end of the allocated message.

PR #3629, opened 2022-07-27
Trq/readonly pup fix

trquinn:
This patch is against v7.0.0, and addresses issue #3628


trquinn:
This is replaced by #3631

PR #3630, opened 2022-07-28
Rename NAMD CI so it can be optional

matthiasdiener:
GitHub branch protection rules use the name of the build (`build` in most of our CI cases) to require it to pass before being able to merge a PR. Since NAMD requires a secret token to download it from their Gitlab, it will fail on PRs that are opened from outside PPL. By renaming the NAMD build, it will become optional to pass before it can be merged.

### DIFF
--- a/.github/workflows/namd.yml
+++ b/.github/workflows/namd.yml
@@ -3,6 +3,9 @@ name: NAMD
 on: pull_request
 
 jobs:
+  # Not named 'build' since 'build' is used by our CI for builds that must succeed before merge.
+  # Since NAMD requires a secret to download its source from Gitlab, CI builds from outside PPl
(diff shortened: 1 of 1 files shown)


stwhite91 (approved):

rbuch (approved):

.github/workflows/namd.yml
```diff
@@ -3,7 +3,7 @@ name: NAMD
 on: pull_request
 
 jobs:
-  build:
+  build-and-test-namd:
```
rbuch, line 6: Maybe add a comment here saying that the name makes it optional?

matthiasdiener: Good idea, added.

PR #3633, opened 2022-08-03
CI: Make NAMD a neutral test for external forks

matthiasdiener:


matthiasdiener:
It does not work in the current state
image

matthiasdiener:
This should hopefully work now (we can only really test it if someone else outside opens a new PR).

evan-charmworks (approved):

evan-charmworks:
Seems like this works, tested with #3712

stwhite91 (approved):

PR #3637, opened 2022-08-09
Fix overflow of ID home bits and improve asserts

rbuch:
Please merge via rebase.

- objid: Fix overflow of home bits
- objid: Add extra assertions and improve messages


dkokron:
I have tested the fix for HomeBits overflowing on Perlmutter using up to 55,296 MPI ranks.  No issues found up to that point.  Please expedite this PR.

stwhite91 (approved):

ericjbohm (approved):
Ideally I think we should abstract the ID bit manipulation operations more, but this solves the immediate problem.

PR #3641, opened 2022-09-12
Doc: Add CkIO documentation

rbuch:


doc/libraries/manual.rst
```diff
@@ -787,3 +787,126 @@ Example
...
+     (default: ``4 * writeStripe``)
+  - ``activePEs`` - Number of PEs to use for I/O (default: min(32, number of
+    PEs))
+  - ``basePE`` - Index of first participating PE (default: 0)
+  - ``skipPEs`` - Gap between participating PEs (default : ``CkMyNodeSize()``)
+
+
+- Starting a session:
+
+  Note there are two variants of the ```startSession`` function, a regular one
```
matthiasdiener, line 856:
```suggestion
  Note there are two variants of the ``startSession`` function, a regular one
```

matthiasdiener (approved):
LGTM, thanks!

ZwFink:
What happens when a full stripe is never written? Is the write flushed on file close? Is there a way to flush manually?

rbuch:
> What happens when a full stripe is never written?

The only contract with the end user is that they write the same number of bytes as they said they were going to when they created the session. That amount can be less than a stripe; if it is, then it will just write that amount.

> Is the write flushed on file close? Is there a way to flush manually?

Internally, when CkIO closes the session, it calls one of `fdatasync`, `fsync`, `FlushFileBuffers` (Windows-specific), or `sync`, in that order of preference. If none of those calls are available, then it warns at compile time. There is no way to flush manually, I think that's too low-level for this library.

ZwFink (approved):

PR #3642, opened 2022-09-14
Fix default PMI on Cray OFI

rbuch:
#3626 ordered the check to default to `simplepmi` before the check for using `slurmpmi2` on Cray OFI platforms, meaning that `simplepmi` was used by default on those platforms rather than `slurmpmi2`. Instead, do the check for Cray platforms first, and only use the `simplepmi` default if that fails.

### DIFF
--- a/buildcmake
+++ b/buildcmake
@@ -471,13 +471,13 @@ for c in "$opt_lrts_pmi" "${opt_compiler[@]}"; do
   [[ -n "$c" ]] && builddir_extra+="-$c"
 done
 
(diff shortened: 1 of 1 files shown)


matthiasdiener (approved):
Could you add a brief comment explaining the default choice on Cray and non-Cray machines?

matthiasdiener (approved):
LGTM, thanks!

namehta4:
Hi All,

May I please request an update on this?
Thanks!

adityapb (approved):

Issue #3645, opened 2022-09-24
Does the cmake build work with clang?


jbakosi:
On debian linux, I do
```sh
cmake -DTARGET=LIBS -DNETWORK=mpi -DSMP=off -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DCMAKE_Fortran_COMPILER=gfortran <charm-src>
```
and I get
```sh
<...>/charmc: line 236: eval: -I: invalid option
eval: usage: eval [arg ...]
Fatal Error by charmc in directory <...>/b
Command -I<...>/b/include -DAMPI_ERRHANDLER_RETURN=0 -DNDEBUG -O3 -O3 -O2 -c <...>/b/pup_f.f90 -o CMakeFiles/conv-utilf.dir/pup_f.f90.o returned error code 2
```
Seems like a command is missing in charmc:236
```sh
DoNoErrCheck() {
	[ -n "$VERBOSE" ] && echo "charmc: Executing $@" 1>&2
	eval "$@"
}
```
Does someone have a workaround?

matthiasdiener:
In general, I would suggest specifying compilers via environment variables, not inside the cmake options when using cmake directly:

```
$ CC=mpicc CXX=mpicxx FC=gfortran cmake -DTARGET=LIBS -DNETWORK=mpi -DSMP=off <charm-src>
```

Does this make a difference for your build?

jbakosi:
As I have multiple mpi installs besides the system-wide one with gcc, e.g., with clang, I usually point cmake to non-system mpi wrappers via either env vars (as you suggest) or via cmake's args. Neither of the above helps though with the error if I have clang behind the mpi wrappers. (Though, I'm also happy with the workaround in #3636.)

matthiasdiener:
That is interesting. Could you please paste the complete `cmake` output in a gist and add a link here?

jbakosi:
https://gist.github.com/jbakosi/a827cbdcab0d3bcfdc4df5874703d4e9

It only errors with `-DNETWORK=mpi`.

jdmaia:
@matthiasdiener  @ebohm I have a couple of colleagues here at AMD who are trying to get Charm++ built with our clang-based AOCC compiler and are snagging in the same error described here. Do you guys have an ETA for a fix on this issue?

evan-charmworks:
[...]

PR #3646, opened 2022-09-26
CMake: Add option to disable Fortran

jbakosi:
This allows skipping fortran using cmake, e.g.,
```
cmake -DENABLE_FORTRAN=off ...
```
A workaround for #3645.


evan-charmworks:
Looks good to me, except I am unsure if there are side effects of moving `find_package(MPI REQUIRED)` above the Fortran enablement section.

evan-charmworks (approved):

stwhite91 (approved):

matthiasdiener:
> Looks good to me, except I am unsure if there are side effects of moving `find_package(MPI REQUIRED)` above the Fortran enablement section.

I _think_ that is fine, comparing the log files does not seem to show a difference as far as I could see. CMake does not find the `MPI_Fortran` bindings, but we are probably not using them?

CMakeLists.txt
```diff
@@ -31,6 +31,15 @@ project(Charm++ LANGUAGES CXX C ASM VERSION 7.0.0)
...
 find_package(OpenMP) # Do this before Fortran, in case we don't have a Fortran compiler
 
+set(CMK_BUILD_MPI 0)
+set(CMK_BUILD_ON_MPI 0)
+set(CMK_CXX_MPI_BINDINGS 1) # Always true, this just checked for a conflict with old mpich versions.
+if(${NETWORK} STREQUAL "mpi")
+  set(CMK_BUILD_MPI 1)
+  set(CMK_BUILD_ON_MPI 1)
+  find_package(MPI REQUIRED)
+endif()
```
matthiasdiener, line 41: Could you mention why you moved the MPI detection to before the Fortran detection?

jbakosi: If `find_package(MPI)` is called downstream where it originally was _and_ if a fortran compiler is really not available on the system, cmake's `FindMPI.cmake` fails with trying to invoke the fortran compiler as
```
CMake Error: CMAKE_Fortran_COMPILER not set, after EnableLanguage
CMake Error at /usr/share/cmake-3.18/Modules/FindMPI.cmake:1213 (try_compile):
  Failed to configure test project build system.        
Call Stack (most recent call first):                    
  /usr/share/cmake-3.18/Modules/FindMPI.cmake:1240 (_MPI_try_staged_settings)
  /usr/share/cmake-3.18/Modules/FindMPI.cmake:1427 (_MPI_check_lang_works)
  CMakeLists.txt:425 (find_package)
```

When I just reproduced this (to remind myself that this was indeed the case), using
```sh
cmake -DTARGET=LIBS -DNETWORK=mpi -DSMP=off -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DCMAKE_Fortran_COMPILER=mpif90 -DENABLE_FORTRAN=off <charm-src>
```
I found that the problem fortran is enabled upstream is because of this code:
```cmake
# No Fortran support with MSVC
if(NOT CMK_COMPILER STREQUAL "msvc")
  enable_language(Fortran OPTIONAL)
endif()
```
which turns on fortran support. Even though this is optional, FindMPI.cmake now tries to invoke the fortran compiler because on linux I have no msvc and fortran as a language is now enabled in cmake for the project. I see in `FindMPI.cmake` that it will test for all languages of (C CXX Fortran) that are enabled.

I'm not sure about the intent but the above "msvc" code appears to be a bug for me: just because the CMK_COMPILER is not msvc, I don't think that that should *add* (optional) fortran support. Perhaps something like the following would be better:
```cmake
option(ENABLE_FORTRAN "Enable Fortran" ON)
if(ENABLE_FORTRAN AND CMAKE_Fortran_COMPILER)

   # No Fortran support with MSVC
   if(NOT CMK_COMPILER STREQUAL "msvc")
     enable_language(Fortran OPTIONAL)
   endif()

  include(FortranCInterface)
  FortranCInterface_VERIFY()
  ...
else()
  message(STATUS "Charm++: No Fortran support.")
endif()
```
or maybe even
```cmake
if(ENABLE_FORTRAN AND CMAKE_Fortran_COMPILER AND NOT CMK_COMPILER STREQUAL "msvc")
  enable_language(Fortran OPTIONAL)
  include(FortranCInterface)
  FortranCInterface_VERIFY()
  ...
else()
  message(STATUS "Charm++: No Fortran support.")
endif()
```

While I have been using this patch with`-DENABLE_FORTRAN=off`, I have now also tested the above with enabling fortran (and with a fortran compiler present on the system) as
```sh
cmake -DTARGET=LIBS -DNETWORK=mpi -DSMP=off -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DCMAKE_Fortran_COMPILER=mpif90 -DENABLE_FORTRAN=on <charm-src>
```
but that (still) fails during make:
```
/home/jbakosi/code/charm-jbakosi/b/bin/charmc: line 232: -I/home/jbakosi/code/charm-jbakosi/b/include: No such file or directory
Fatal Error by charmc in directory /home/jbakosi/code/charm-jbakosi/b
   Command -I/home/jbakosi/code/charm-jbakosi/b/include -DAMPI_ERRHANDLER_RETURN=0 -DNDEBUG -O3 -O3 -O2 -c /home/jbakosi/code/charm-jbakosi/b/pup_f.f90 -o CMakeFiles/conv-utilf.dir/pup_f.f90.o returned error code 127
charmc exiting...
```
This seems to be some other problem, maybe only with mpi and/or clang?

matthiasdiener (approved):

PR #3663, opened 2022-11-10
Option to Disable TLS

### LINKED ISSUE #3600: Unable to use python bindings that depend on Charm++
Hi folks!

We use python bindings to make some data visualization work easier. Unfortunately the use of `-ftls-model=initial-exec` is incompatible with loading shared libraries after executable initialization (assuming I've understood all the errors correctly!). This means the only way we can use Charm++ 7 is to patch it:

```
diff --git a/cmake/detect-features-cxx.cmake b/cmake/detect-features-cxx.cmake
index d3aa6ab94..830da9cf1 100644
--- a/cmake/detect-features-cxx.cmake
+++ b/cmake/detect-features-cxx.cmake
@@ -38,12 +38,12 @@ endif()
 
 # Needed so that tlsglobals works correctly with --build-shared
 # See https://github.com/UIUC-PPL/charm/issues/3168 for details.
-check_cxx_compiler_flag("-ftls-model=initial-exec" CMK_COMPILER_KNOWS_FTLS_MODEL)
-if(CMK_COMPILER_KNOWS_FTLS_MODEL)
-  set(OPTS_CC "${OPTS_CC} -ftls-model=initial-exec")
-  set(OPTS_CXX "${OPTS_CXX} -ftls-model=initial-exec")
-  set(OPTS_LD "${OPTS_LD} -ftls-model=initial-exec")
-endif()
+# check_cxx_compiler_flag("-ftls-model=initial-exec" CMK_COMPILER_KNOWS_FTLS_MODEL)
+# if(CMK_COMPILER_KNOWS_FTLS_MODEL)
+#   set(OPTS_CC "${OPTS_CC} -ftls-model=initial-exec")
+#   set(OPTS_CXX "${OPTS_CXX} -ftls-model=initial-exec")
+#   set(OPTS_LD "${OPTS_LD} -ftls-model=initial-exec")
+# endif()
 
 # Allow seeing own symbols dynamically, needed for programmatic backtraces
 check_cxx_compiler_flag("-rdynamic" CMK_COMPILER_KNOWS_RDYNAMIC)
diff --git a/src/scripts/configure.ac b/src/scripts/configure.ac
index f6c0f311b..41a7c9f46 100644
--- a/src/scripts/configure.ac
+++ b/src/scripts/configure.ac
@@ -820,15 +820,15 @@ then
 fi
 
 # Needed so that tlsglobals works correctly with --build-shared
-# See https://github.com/UIUC-PPL/charm/issues/3168 for details.
-test_cxx "whether C++ compiler accepts -ftls-model=initial-exec" "yes" "no" "-ftls-model=initial-exec"
-if test $strictpass -eq 1
-then
-    add_flag 'CMK_COMPILER_KNOWS_FTLS_MODEL="1"' "tlsglobals"
-    OPTS_CC="$OPTS_CC -ftls-model=initial-exec"
-    OPTS_CXX="$OPTS_CXX -ftls-model=initial-exec"
-    OPTS_LD="$OPTS_LD -ftls-model=initial-exec"
-fi
+# # See https://github.com/UIUC-PPL/charm/issues/3168 for details.
+# test_cxx "whether C++ compiler accepts -ftls-model=initial-exec" "yes" "no" "-ftls-model=initial-exec"
+# if test $strictpass -eq 1
+# then
+#     add_flag 'CMK_COMPILER_KNOWS_FTLS_MODEL="1"' "tlsglobals"
+#     OPTS_CC="$OPTS_CC -ftls-model=initial-exec"
+#     OPTS_CXX="$OPTS_CXX -ftls-model=initial-exec"
+#     OPTS_LD="$OPTS_LD -ftls-model=initial-exec"
+# fi
 
 # Test for a flag important for shared linking
 test_cxx "whether C++ compiler accepts -fvisibility=hidden" "yes" "no" "-fvisibility=hidden"
```
It would be nice to have a solution that doesn't require us to patch Charm++. From https://github.com/UIUC-PPL/charm/issues/3168 it wasn't clear to me why changing the TLS behavior was the right solution. We've had a lot of issues with allocators like jemalloc and tcmalloc if the TLS module was global-dynamic, and using the faster allocators can actually make a difference. Having said all that, the real issue is not being able to use python bindings anymore. Since all of our classes need to be serializable they all have a `pup` function and so they all depend on Charm++, so any build flags Charm++ uses basically percolate through our entire build system.

Happy to discuss more and provide any additional details that would be useful in finding a solution that works for everyone!

Cheers,

Nils

ZwFink:
Closes #3600. Adds the option ``--disable-tls```  to the Charm++ build CLI that disables TLS, which can cause problems in certain contexts (#3600).

If Charm++ or AMPI are built with ```--disable-tls``` and try to compile an AMPI program with `-tlsglobals`, the compiler will return the following:
```
Error: Charm was built with '--disable-tls' and -tlsglobals is not available.
```


stwhite91 (commented):
If a user builds Charm/AMPI with this flag, then tries to compile an AMPI program with -tlsglobals, will charmc/ampicc fail and tell the user that it's not supported in their build? I'm not sure without checking, but we don't want it to compile and then fail at runtime because of the missing compiler flag.

Also the commit message should reference a github issue # if there is one motivating this change

ZwFink:
@nilsdeppe Can you check if this fixes your issue with Python?

nilsdeppe:
Yep, we'll give it a try! Probably @knelli2 will do it and report back. Thanks so much!

knelli2:
Building charm with this branch and the `--disable-tls` option, I was able to build the SpECTRE python bindings and they work fine. No need to patch charm. So this looks good!

ericjbohm (approved):

src/scripts/configure.ac
```diff
@@ -2582,14 +2591,21 @@ then
   fi
 fi
 
+if test "$enable_tls" = "yes"
+then
+    add_flag 'CMK_USER_DISABLED_TLS="0"' "tlsglobals"
+else
+    add_flag 'CMK_USER_DISABLED_TLS="1"' "glsglobals"
```
rbuch, line 2598: Is this supposed to be `glsglobals`?

src/scripts/configure.ac
```diff
@@ -2582,14 +2591,21 @@ then
...
+fi
+
 AC_DEFINE_UNQUOTED(CMK_SUPPORTS_SWAPGLOBALS, $CMK_SUPPORTS_SWAPGLOBALS, [whether supports swapglobals])
 add_flag "CMK_SUPPORTS_SWAPGLOBALS=$CMK_SUPPORTS_SWAPGLOBALS" 'swapglobals'
 add_make_flag "CMK_SUPPORTS_SWAPGLOBALS:=$CMK_SUPPORTS_SWAPGLOBALS" 'swapglobals'
 AC_DEFINE_UNQUOTED(CMK_SUPPORTS_TLSGLOBALS, $CMK_SUPPORTS_TLSGLOBALS, [whether supports tlsglobals])
 add_flag "CMK_SUPPORTS_TLSGLOBALS=$CMK_SUPPORTS_TLSGLOBALS" 'tlsglobals'
 add_make_flag "CMK_SUPPORTS_TLSGLOBALS:=$CMK_SUPPORTS_TLSGLOBALS" 'tlsglobals'
 
-
```
rbuch, line 2592: Extraneous newline.

src/scripts/charmc
```diff
@@ -1472,7 +1472,12 @@ if [ "$TLSGLOBALS" = "1" ]
 then
     if [[ "$CMK_SUPPORTS_TLSGLOBALS" = '0' ]]
     then
-      echo "Error: -tlsglobals is not supported on $CMK_VDIR"
+	    if [[ "$CMK_USER_DISABLED_TLS" = '1' ]]
+	    then
+        echo "Error: Charm was built with '--disable-tls' and -tlsglobals is not available."
+	    else
+        echo "Error: -tlsglobals is not supported on $CMK_VDIR"
```
rbuch, line 1479: Inconsistent indentation (the conditional lines have tabs, the statement lines have spaces, it looks like).

rbuch (changes requested):
The functionality looks good, I'm only requesting minor stylistic changes.

src/scripts/charmc
ZwFink: Changed all to spaces for this range of lines

rbuch (approved):

PR #3664, opened 2022-11-15
CMake: Do not substitute empty CMK_LUSTREAPI with 0

rbuch:
CMK_LUSTREAPI holds the library argument for linking with the Lustre API, "-llustreapi" when it is detected, otherwise "". Special case the replacement that sets empty variables to 0 so that CMK_LUSTREAPI should remain set to "" when Lustre is not detected, otherwise using -module CkIO causes a spurious attempt to link with "0".


matthiasdiener (approved):
LGTM, thanks!

ZwFink (approved):

PR #3666, opened 2022-11-23
CMake: Quote parameter expansion for finding extra opts

rbuch:
Not quoting can cause the expansion to be matched as a pattern rather
than a literal. Reported by ShellCheck SC2295.


matthiasdiener (approved):
LGTM, thanks!

NK-Nikunj (approved):

PR #3672, opened 2022-11-23
    CI: Install UCX from apt

rbuch:
Use the packaged version of UCX rather than pinning to a specific version and manually installing it.


matthiasdiener (approved):

matthiasdiener:
We could also just try using `sudo apt install libucx-dev ucx`

matthiasdiener (approved):

ZwFink (approved):

PR #3675, opened 2022-12-02
CkIO read

mjacob1002:
This PR is for the reading capabilities of the CkIO library built on top of Charm++. The additions of this PR are as follows:

1) Added a ReadSession chare array (alternatively called IO chare array) that read the data of the relevant session into the chare array and serve it to the user.

2) Added the infrastructure to create an actual read session with helpers in the impl::Director class

3) implemented the read(Session session, size_t offset, size_t bytes) that actually serves the user the data they want by serving the data to them via a ReadCompleteMsg*. 

4) Added a read_stripe field that will determine how many chares to be used in the IO chare array. 

Also might have to add a sort of "tagging" system for each read so SDAG can be utilized by the user to order the processing of the reads.


mjacob1002:
@ZwFink @rbuch If you guys could take a look I would appreciate it!

src/libs/ck-libs/io/ckio.C
```diff
@@ -389,6 +437,90 @@ namespace Ck { namespace IO {
...
+				CkExit();
+			}
+			ifs.seekg(_my_offset); // jump to the point where the chare should start reading
+			char* buffer = new char[_my_bytes];
+			ifs.read(buffer, _my_bytes);
+			_buffer.resize(_my_bytes, 'z'); // resize it and init with 'z' to denote what hasn't been changed
+			_buffer.shrink_to_fit(); // get rid of any extra capacity 
+			for(size_t i = 0; i < _my_bytes; ++i){
+				_buffer[i] = (buffer[i]);
+			}
```
rbuch, line 481: Why not just copy directly into `_buffer`?

src/libs/ck-libs/io/ckio.C
```diff
@@ -398,6 +530,57 @@ namespace Ck { namespace IO {
...
+				_data_buffer[start_idx + counter] = ch;
+		//		ckout << "Read Assembler " << read_chare_offset << "has set " << (start_idx + counter) << " to " << _data_buffer[start_idx + counter] << endl;
+				++counter;	
+			}
+			_bytes_left -= data.size(); // decrement the number of remaining bytes to read
+			if(_bytes_left) return; // if there are bytes still to read, just return
+			char* buffer = data.data(); 
+			// ckout << "Offset " << _read_offset << " belongs to Pe " << CkMyPe() << endl;
+			ReadCompleteMsg* msg = new (_data_buffer.size()) ReadCompleteMsg();
+			memcpy(msg -> data, buffer, _data_buffer.size());
```
rbuch, line 574: This doesn't look right, it's copying from `buffer`, but using `_data_buffer`'s size. Has this been tested with a read that requires data from multiple readers?

src/libs/ck-libs/io/ckio.C
```diff
@@ -25,6 +27,7 @@ using std::max;
 using std::map;
 using std::string;
 
+
```
rbuch, line 30: Take out this and other extraneous newlines in the PR.

src/libs/ck-libs/io/ckio.h
```diff
@@ -70,11 +75,40 @@ namespace Ck { namespace IO {
...
+  void startReadSession(File file, size_t bytes, size_t offset, CkCallback ready);
+
+  /**
+   * Used to end the current read session and will then invoke the after_end callback that takes a CkReductionMsg* with nothing in it
+   * Will effectively call ckDestroy() on the CProxy_Reader of the associated FileInfo
+   *
+   *
+   *
+   */
+
```
rbuch, line 95: Extra space should be removed. Also, match the commenting style of the other functions here (`///` instead of `/**`).

rbuch (changes requested):
- [ ] Remove the binaries that were added in this PR (`lb_test`, `period_selection`, and `ping`). 
- [ ] What is the empty file `src/libs/ck-libs/io/headers` supposed to do? If nothing, it should also be removed.
- [ ] Cleanup extraneous comments and output.
- [ ] Cleanup spacing and style.

src/libs/ck-libs/io/Makefile
```diff
@@ -18,7 +18,7 @@ $(LIB): $(LIBOBJ)
 
 headers: $(HEADERS)
 	cp $(HEADERS) $(CDIR)/include/
-	touch headers
+	# touch headers
```
NK-Nikunj, line 21: Remove this comment if it is not required.

src/libs/ck-libs/io/ckio.C
```diff
@@ -389,6 +435,86 @@ namespace Ck { namespace IO {
...
+		}
+
+		void clearBuffer() {
+			_buffer.clear(); // clears the buffer
+		}
+
+		void readData(){
+			std::ifstream ifs(_file -> name); // open the file
+			if(ifs.fail()){ // error handling if opening the file failed
+				std::cerr<< "There was an error on ReadSession chare " << thisIndex << " when trying to open file " << _file -> name << std::endl;
```
NK-Nikunj, line 469: This should use `CkAbort` instead of `std::cerr`

src/libs/ck-libs/io/ckio.C
```diff
@@ -398,6 +524,57 @@ namespace Ck { namespace IO {
...
+			// ckout << "RA Buffer init\n"; 
+			/*
+			for(char ch : _data_buffer)
+				ckout << ch;
+			ckout << endl;
+			std::cout << "ReadAssembler built\n";
+			*/
+		}
+
+		void shareData(size_t read_chare_offset, std::vector<char> data){
```
NK-Nikunj, line 554: Do not use `std::vector<char>` as an entry method parameter. Change this to `char* data, int size`. You can call `data.data(), data.size()` as parameters when calling the function.

tests/charm++/load_balancing/lb_test/Makefile
```diff
@@ -1,7 +1,7 @@
 -include ../../../common.mk
 # Makefile for load-balancing test program
 
-OPTS	= -O3
+OPTS	= -O3 -verbose
```
NK-Nikunj, line 4: -verbose should be removed

NK-Nikunj (changes requested):

mjacob1002:
At this point, zero-copy from the ReadSession chares(I/O chares) to the ReadAssembler's ReadInfo struct has been implemented. ReadAssembler has been changed to be a group that manages reads. Structure is now that ReadAssembler is created when startReadSession is invoked, mapping each Session object to the CProxy_ReadAssembler for that session. Then, when a read for a particular session comes in, it's look up in the manager of that PE, and then the ReadAssembler serves the read request. 

The ReadAssembler keeps an internal map for every read request, assigning a library-only tag to map to a ReadInfo struct that contains meta-information about that particular read, including the ReadCompleteMsg* which stores the actual return buffer. ReadSessions still send data to particular ReadAssembler chare, which will then return buffer to user via ReadCompleteMsg*.

PR #3676, opened 2022-12-03
Modifies converse schedeuler to prioritize so nodeGroup messages

### LINKED ISSUE #3674: Nodegroup messages are too low a priority
I noted this in a tram-replacement experiment, with histogram code from bale suite. Nodegroup messages carrying boatload of data were arriving on the node but not being processed as all PEs were busy cycling through there own locally generated work. (Although the work generation method was broken down using CThYield() every x iterations). @nilsdeppe also seconded this request on slack.
What is happening: scheduler's function for finding next message always processes non-local (i.e. messages from others but meant for my PE) and local (from my pe) messages, so nodeGroup messages, although they can be processed by any PE, get delayed. 
The request is to allow nodegroup messages to execute earlier.. somehow.

lvkale:
Modifies converse schedeuler's getNextMessage so nodeGroup messages can run with higher priority over local
closes #3674
As it is, nodeGroup messages are not checked until all local and regular Charm queue (prio Q) messages are checked, which cause issues when the applicaiton is using nodeGroup messages in the hope that *some* PE will attend to it quickly. The change makes getNextMessage check nodeGroup queue every 2^nodeGrpFreq iterations with high priority in addition to its usual check after exhasuting local queues (except task Q). This commit has not been tested at all. But pusing it to allow others to help me test/fix it.

### DIFF
--- a/src/conv-core/convcore.C
+++ b/src/conv-core/convcore.C
@@ -1656,6 +1656,7 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 #if CMK_NODE_QUEUE_AVAILABLE
 	s->nodeQ=CsvAccess(CsdNodeQueue);
 	s->nodeLock=CsvAccess(CsdNodeQueueLock);
+	s->nodeGrpFreq =2; // check node queue once in 4 (2^2) iterations. 
 #endif
 #if CMK_GRID_QUEUE_AVAILABLE
 	s->gridQ=CpvAccess(CsdGridQueue);
@@ -1664,6 +1665,7 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 	s->taskQ = CpvAccess(CsdTaskQueue);
 	s->suspendedTaskQ = CpvAccess(CmiSuspendedTaskQueue);
 #endif
+
 }
 
 
@@ -1719,7 +1721,20 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
  */
 void *CsdNextMessage(CsdSchedulerState_t *s) {
 	void *msg;
-	if((*(s->localCounter))-- >0)
+
+s->iter++;
+
+#if CMK_NODE_QUEUE_AVAILABLE
+	if (1 == (s->iter & (1 << s->nodeGrpFreq)) // since we use nodeGrpFreq == 0 to mean
+                                           // don't check NodeQ with high priority, i
+					   // value of 1 serves well as when to check it.i but we sshould avoid "%" operator
+					// note: s->nodeGrpFreq is raised to a power of  2
+	{
+	  msg = CmiGetNonLocalNodeQ();
+	  if (NULL != msg) return msg;
+	}
+#endif
+        if((*(s->localCounter))-- >0)
 	  {
               /* This avoids a race condition with migration detected by megatest*/
               msg=CdsFifo_Dequeue(s->localQ);
--- a/src/conv-core/converse.h
+++ b/src/conv-core/converse.h
@@ -1145,6 +1145,12 @@ typedef struct {
   void *localQ;
   Queue nodeQ;
   Queue schedQ;
+  unsigned short iter; // counting number of sched iterations (hopefully of for it to roll over	 
+  unsigned short nodeGrpFreq; // call nodegroup queue once every 2^nodeGrpFreq iterations with high prio 
+	// should add a function to change this from the program for advanced users. One obstacle: 
+        // it is inside a struct that is on stack, and so not accessible for standalone functions. Need to
+        // resolve this by making a schedule a c++ object, but even then we need a ptr to the currently-running scheduler
+	// 0 means do not check nodegroup queue with high prio.. will be checked with low prio after other Qs
   int *localCounter;
 #if CMK_OBJECT_QUEUE_AVAILABLE
   Queue objQ;


lvkale:
One thing to worry about is whether this change causes performance degradation by making the scheduler check the nodequeue too often (depend on whether the check is expensive, even for an empty queue, because of locking). It'd be nice if someone were to run a performance regression test.

src/conv-core/convcore.C
```diff
@@ -1656,6 +1656,7 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 #if CMK_NODE_QUEUE_AVAILABLE
 	s->nodeQ=CsvAccess(CsdNodeQueue);
 	s->nodeLock=CsvAccess(CsdNodeQueueLock);
+	s->nodeGrpFreq =2; // check node queue once in 4 (2^2) iterations.
```
ZwFink, line 1659: Should this be configurable by the user?

src/conv-core/convcore.C
```diff
@@ -1656,6 +1656,7 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 #if CMK_NODE_QUEUE_AVAILABLE
 	s->nodeQ=CsvAccess(CsdNodeQueue);
 	s->nodeLock=CsvAccess(CsdNodeQueueLock);
```
ZwFink, line 1658: Should ```s->iter``` be initialized here also?

src/conv-core/convcore.C
```diff
@@ -1719,7 +1721,21 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
  */
 void *CsdNextMessage(CsdSchedulerState_t *s) {
 	void *msg;
-	if((*(s->localCounter))-- >0)
+
+	s->iter++;
+
+#if CMK_NODE_QUEUE_AVAILABLE
+	if (1 == (s->iter & (1 << s->nodeGrpFreq))) // since we use nodeGrpFreq == 0 to mean
```
ZwFink, line 1728: Do we gain from avoiding the modulo operator this way? If the definition of ```s->nodeGrpFreq``` is compile-time constant, can we assume that the compiler will do the corresponding strength reduction, provided that ```s->nodeGrpFreq``` is a power of 2? I checked a recent version of GCC and it generates the same code if we use modulo or this trick

rbuch: Separately, this comment/implementation seems wrong, if `nodeGrpFreq = 0`, then `1 << s->nodeGrpFreq` would be 1, so that would check the node queue every other iteration, seemingly contradicting the comment saying that that would not check the node queue with high priority. Also, testing it via `1 == ` means that it would only work in the case where `nodeGrpFreq = 0`, otherwise it would only ever produce 0 or a value greater than 1.

jcphill: I think you want `if (0 == (s->iter & ((1 << (s->nodeGrpFreq-1))-1)))`
nodeGrpFreq of 0 -> 0 == iter & -1
nodeGrpFreq of 1 -> 0 == iter & 0
nodeGrpFreq of 2 -> 0 == iter & 1
nodeGrpFreq of 3 -> 0 == iter & 3

jcphill: Not sure what 1 << -1 produces.

jcphill: Hoping it's zero.

jcphill: It's not.
`if (s->nodeGrpFreq && ! (s->iter & ((1 << (s->nodeGrpFreq-1))-1)))`

src/conv-core/convcore.C
lvkale: Looks like i forgot about this PR once I traveled to India. To answer your question: ideally, yes. And we will make it user configurable when we do registration-based/table-driven scheduler for managing queues. But for now, 4 seems like a reasonable ad-hoc number. Adding another build time option for a feature that will be replaced in near future seems like an overkill.

src/conv-core/convcore.C
lvkale: Looks like I let this slide.. @ZwFink will you complete it ? You can simplify the code (that I had to avoid the % operator) because I am convinced by your argument that the compiler will optimize it.

lvkale:
I thought this was already merged. @ZwFink @ericjbohm .. will you please take a look?

jbakosi:
> @lvkale this is the code we discussed during the meeting today

Sorry for the spam, but where is the meeting usually announced?

ericjbohm:
Should I consider all the older unresolved comments here as acceptably resolved?

ericjbohm:
In order for this to be mergeable it should be modified to no longer be a draft and the reviewer comments should be addressed and resolved.  @lvkale

ZwFink (approved):

ericjbohm:
> One thing to worry about is whether this change causes performance degradation by making the scheduler check the nodequeue too often (depend on whether the check is expensive, even for an empty queue, because of locking). It'd be nice if someone were to run a performance regression test.

Was that performance analysis done?  If so, what were the results?

src/conv-core/convcore.C
stwhite91: Still needs to be addressed...

src/conv-core/convcore.C
```diff
@@ -1665,6 +1666,7 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 	s->taskQ = CpvAccess(CsdTaskQueue);
 	s->suspendedTaskQ = CpvAccess(CmiSuspendedTaskQueue);
 #endif
+
```
stwhite91, line 1669: Remove extraneous whitespace change

src/conv-core/converse.h
```diff
@@ -1129,6 +1129,12 @@ typedef struct {
   void *localQ;
   Queue nodeQ;
   Queue schedQ;
+  unsigned short iter; // counting number of sched iterations (hopefully of for it to roll over
```
stwhite91, line 1132: Remove text after the parenthetical

ZwFink:
> > One thing to worry about is whether this change causes performance degradation by making the scheduler check the nodequeue too often (depend on whether the check is expensive, even for an empty queue, because of locking). It'd be nice if someone were to run a performance regression test.
> 
> Was that performance analysis done? If so, what were the results?

Results of a benchmark that has one chare sending 10 messages to itself 500,000 times (SMP mode, higher is better). The entry method just increments a counter to track whether it should send another 10 messages to itself, so it should stress the queueing system. This result shows an overhead of $0.44$% averaged over 50 runs. This translates to the overhead of a few $ns$ per message. Any realistic, small tasks are on the order of a few $us$ per message. I think we should pay more attention to the fact that the scheduler can only churn through ~1.2million local entry methods per second than the addition of a few nanoseconds of overhead.
boxplot_comparison

nilsdeppe:
Another possibly interesting point is that we run with 6 virtual nodes per physical node on our 192-cores-per-node machine because of better communication. This seems counterintuitive with the model I think Sanjay described that intranode communication shouldn't be affected by the comm thread. Maybe there's a lot of locking going on?

src/conv-core/convcore.C
```diff
@@ -1655,8 +1655,10 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
 	s->schedQ=CpvAccess(CsdSchedQueue);
 	s->localCounter=&(CpvAccess(CsdLocalCounter));
 #if CMK_NODE_QUEUE_AVAILABLE
-	s->nodeQ=CsvAccess(CsdNodeQueue);
-	s->nodeLock=CsvAccess(CsdNodeQueueLock);
+  s->nodeQ=CsvAccess(CsdNodeQueue);
```
stwhite91, line 1658: Indentation doesn't match surrounding code

src/conv-core/convcore.C
```diff
@@ -1720,7 +1722,19 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
  */
 void *CsdNextMessage(CsdSchedulerState_t *s) {
 	void *msg;
-	if((*(s->localCounter))-- >0)
+
+	s->iter++;
+
+#if CMK_NODE_QUEUE_AVAILABLE
+ // we use nodeGrpFreq == 0 to mean
```
stwhite91, line 1729: Indentation

src/conv-core/converse.h
```diff
@@ -1129,6 +1129,12 @@ typedef struct {
   void *localQ;
   Queue nodeQ;
   Queue schedQ;
+  unsigned short iter; // counting number of sched iterations
+  unsigned short nodeGrpFreq; // call nodegroup queue once every nodeGrpFreq iterations with high prio 
+	// should add a function to change this from the program for advanced users. One obstacle:
```
stwhite91, line 1134: Would prefer comments like this go into github issues rather than the code

src/conv-core/convcore.C
```diff
@@ -1720,7 +1722,19 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
  */
 void *CsdNextMessage(CsdSchedulerState_t *s) {
 	void *msg;
-	if((*(s->localCounter))-- >0)
```
jcphill, line 1723: The new code should be below the localCounter branch, right above the CmiGetNonLocal() call. The default CsdLocalMax is 0 so normally it won't matter, but if the user says to prioritize local messages then they should come first.

jcphill (changes requested):

src/conv-core/convcore.C
```diff
@@ -1720,7 +1722,19 @@ void CsdSchedulerState_new(CsdSchedulerState_t *s)
  */
```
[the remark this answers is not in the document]
jcphill: Even better, break up the exposition so that the explanations are adjacent to the code they are intended to describe.

Issue #3662, opened 2022-11-03
Add dynamic insertion tests


rbuch:
@kidder provided some dynamic insertion tests (at https://github.com/kidder/charm_spectre_tests) as part of the reproduction effort for #3660. These tests should be added to the main Charm repo and run with CI so that we can catch any regressions that may come up.

PR #3683, opened 2023-01-24
Fix stale location manager entry on dynamic deletions

### LINKED ISSUE #3660: issues with dynamic insertion and deletion of array chare elements
Hello,

I am developing adaptive mesh refinement for [spectre](https://github.com/sxs-collaboration/spectre) (which uses charm++) and have encountered several issues when inserting new elements and destroying old elements of an array chare.

The issues could be caused by my missing or not understanding the documentation regarding dynamic array insertion/deletion such as calls to `beginInserting` and `doneInserting`.

I have put together several charm++ only examples that illustrate two issues I have encountered, as well as a complete 1D example of the AMR algorithm that I am trying to implement.

The first issue is reproduced in `dynamic_array_expansion`.  The issue is that some newly created elements execute all entry methods called from the beginning of execution instead of those called after their creation.  This example fails on the `main` branch but not on `v7.0.0`.  git bisection points to commit 5e26e07023 (which I note is also blamed for #3643).

The second issue is reproduced in `dynamic_array_hang`.  The issue is that the code hangs in some instances when an element has been deleted at some point and then recreated later on a different PE.  This fails on v7.0.0, but the first issue prevents this example from being tested on the main branch.

Please let me know if there is any additional information you need.

adityapb:
Adding tests from the spectre group #3662 

Resolves #3660 - The issue was caused by the location manager entry not being deleted correctly on dynamic deletions

### DIFF
--- a/src/ck-core/cklocation.C
+++ b/src/ck-core/cklocation.C
@@ -2771,6 +2771,10 @@ void CkLocMgr::removeFromTable(const CmiUInt8 id)
     CkAbort("CkLocMgr::removeFromTable called on invalid index!");
 #endif
   hash.erase(id);
+  // Don't erase this during migration because the entry will be updated to reflect
+  // the new location by calling recordEmigration
+  if (!duringMigration)
+    cache->erase(id);
 #if CMK_ERROR_CHECKING
   // Make sure it's really gone
   if (NULL != elementNrec(id))
--- a/tests/charm++/Makefile
+++ b/tests/charm++/Makefile
@@ -28,6 +28,11 @@ DIRS = \
   periodic \
   bombard \
   varTRAM \
+  dynamic_insertion \
+  dynamic_insertion_deletion \
+  amr_1d_random \
+  amr_1d_simple \
+  anytime_migration \
 
 FTDIRS = \
   jacobi3d \
@@ -41,6 +46,7 @@ NONSCALEDIRS = \
   sdag \
   sparse \
   charmxi_parsing \
+  anytime_migration \
 
 TESTPDIRS = $(filter-out $(NONSCALEDIRS),$(TESTDIRS))
 
--- a/tests/charm++/amr_1d_random/Makefile
+++ b/tests/charm++/amr_1d_random/Makefile
@@ -1,0 +1,27 @@
+-include ../../common.mk
+CHARMC=../../../bin/charmc $(OPTS)
+
+all: amr_1d_random
+
+amr_1d_random: random_amr.decl.h amr_1d_random.C
+	$(CHARMC) amr_1d_random.C -o amr_1d_random 
+
+random_amr.decl.h: amr_1d_random.ci
+	$(CHARMC) amr_1d_random.ci
+
+test: amr_1d_random
+	$(call run, +p1 ./amr_1d_random)
+	$(call run, +p2 ./amr_1d_random)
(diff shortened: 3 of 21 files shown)


ericjbohm (approved):

tests/charm++/dynamic_array_hang/dynamic_array_hang.C
```diff
@@ -0,0 +1,273 @@
...
+//  - Test hangs on more than one PE for charm v7.0.0
+//
+//  - Test fails on latest because of the dynamic_insertion error before getting
+//    to the point v7.0.0 hangs
+//
+//  - It appears that when one of the initial elements is destroyed and then
+//    recreated in a later iteration on a PE other than PE 0, it never receives
+//    the ping.  (Could this be related to the load balancing issue (3605) that
+//    was addressed by charm++ PR 3614?)
+//
```
rbuch, line 54: Same as earlier, should be removed if fixed in this PR.

tests/charm++/dynamic_array_expansion/dynamical_insertion.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
rbuch, line 1: `#pragma once` is non-standard, it should ideally be replaced with a more portable solution,

tests/charm++/random_1d_amr_example/random_amr.C
```diff
@@ -0,0 +1,619 @@
...
+//      - If it neither want to join nor split, the element updates its data
+//        (i.e. determines its new neighbors).
+//
+// Current behaviour (Oct 25, 2022):
+//
+//  - hangs (or randomly fails) on more than two PEs for charm v7.0.0
+//    because of dynamic_array_hang
+//
+//  - fails on latest because of the dynamic_insertion error before getting
+//    to the point v7.0.0 hangs
```
rbuch, line 78: Same as other blocks.

tests/charm++/dynamic_array_expansion/dynamical_insertion.ci
```diff
@@ -0,0 +1,31 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
```
rbuch, line 2: Presumably we want these tests to fall under the normal Charm++ license, so we should probably remove these blocks from all the files. The original authors of the tests have already implicitly agreed to this, but it's perhaps worth double checking with them that they're willing to contribute this under our usual license.

rbuch (dismissed):
Mostly looks good, just a few minor quibbles. Overall, I think I'd prefer if this was rebased into two commits, one for the  change to `cklocation.C` and one that adds the tests.

adityapb:
@rbuch - I think I have addressed all the comments. Let me know if you'd suggest any other changes

stwhite91 (commented):
Can you put the issue # in the commit msg header, ie preface it with "Bug #3660: "?

Also I think we can name the tests more descriptively and consistently with the rest of the tests. Here are my suggestions:

dynamic_array_expansion  ->  dynamic insertion
dynamic_array_hang  ->  dynamic_insertion_deletion
random_1d_amr_example  -> amr_1d_random
simple_1d_amr  ->  amr_1d_simple

I would also suggest renaming the source files for each test similarly. Otherwise this all looks good to me

stwhite91 (approved):

PR #3685, opened 2023-01-24
Deprecate "atomic" SDAG keyword

rbuch:
From @stwhite91's suggestion here: https://github.com/UIUC-PPL/charm/discussions/3682#discussioncomment-4768759. This is the only `.ci` file my grep found that used `atomic`. Are there other examples that use it, @stwhite91?

### DIFF
--- a/doc/charm++/manual.rst
+++ b/doc/charm++/manual.rst
@@ -1659,17 +1659,14 @@ and code blocks that they define. These definitions appear in the
 ``.ci`` file definition of the enclosing chare class as a ‘body’ of an
 entry method following its signature.
 
-The most basic construct in SDAG is the ``serial`` (aka the ``atomic``)
-block. Serial blocks contain sequential C++ code. They’re also called
-atomic because the code within them executes without returning control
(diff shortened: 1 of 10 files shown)


stwhite91:
Some of the SDAG examples contain .sdag files with atomic in them. Also the wording in the commit msg is backwards, should be 'use serial instead of atomic'.

rbuch:
> Some of the SDAG examples (in src/langs/sdag/examples/) contain .sdag files with atomic in them. Also the wording in the commit msg is backwards, should be 'use serial instead of atomic'.

Aha, gotcha, I've added another commit to replace `atomic` in those `.sdag` files. And good catch on the wording, I've corrected that, too.

stwhite91:
I think we should remove atomic from the Charm++ manual section 2.2.4 as well: https://charm.readthedocs.io/en/latest/charm%2B%2B/manual.html#structured-control-flow-structured-dagger

Could maybe leave one note there that atomic is deprecated.

stwhite91 (approved):

PR #3690, opened 2023-02-21
Always use expedited or immediate

ZwFink:


ZwFink:
@rbuch Can you double-check that this does what it says on the tin? I changed all uses of CK_MSG_EXPEDITE and CK_MSG_SKIP_OR_IMM.

src/ck-core/ck.C
```diff
@@ -1735,7 +1735,7 @@ void CkSendMsg(int entryIdx, void *msg,const CkChareID *pCid, int opts)
   _TRACE_CREATION_1(env);
   if (destPE!=-1) {
     CpvAccess(_qd)->create();
-    if (opts & CK_MSG_SKIP_OR_IMM)
+    if (opts | CK_MSG_SKIP_OR_IMM)
```
rbuch, line 1738: This (and all the other changes here) will always evaluate to true, so they should probably just be replaced with an `if (true)` if we decide to make this change wholesale.

However, I think we should not completely disable priorities, but instead put them behind a runtime flag or similar, in which case these conditions would look something like `if (disable_priorities || opts & CK_MSG_SKIP_OR_IMM)`

ericjbohm: I agree, the change should be controllable at least at compile time, if not run time.

ericjbohm: Also, immediate is a different feature.  It executes the method within the communication thread on the receiving process.  That has a variety of constraints, and therefore correctness implications, that expedited does not.

ZwFink: Makes sense! I should have noted that I wanted to make the minimal set of changes so we can test performance. It's not a proposal for the final implementation, if we decide to go ahead with the idea.

Issue #387, opened 2013-12-30
Pack variable-envelope structures to reduce overhead bytes


PhilMiller:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/387*

---
Currently, the various variable-envelope structures (`s_array, s_objid, s_group`, etc) are declared such that the compiler will pad them at the end so that they meet their individual alignment restrictions when allocated as part of an array. These objects will never be part of an array, though, and so don't need that padding. Right now, that padding is included in every message we send on the wire, since it's included in the value of `sizeof(s_foo)` used in message-size calculations. If we enable `#pragma pack` for these structures, we can trim some bytes from just about every message we send.

PhilMiller:
*Original date: 2013-12-30 23:29:06*

---
The primary barrier to this may well be compiler support. It's documented for at least GCC, Clang, MSVC++, XLC. We'll definitely need to test ICC, PGCC, and Cray. Even without compiler support, the code should still work fine with such a declaration, just without the optimization benefit. The thing to be cautious of is that some compilers have been known to have bugs in their implementation of `#pragma pack`, though usually involving the sorts of vectorization of array access that's irrelevant here.

PhilMiller:
*Original date: 2013-12-31 00:01:07*

---
ICC 13.1 won't pack the objid type defined in #170, complaining (in a warning on each compilation unit) that it's non-POD. ICC 14.0 handles it correctly, and tests run successfully.

pplimport:
*Original author: Lukasz Wesolowski*
*Original date: 2014-01-21 22:57:09*

---
We currently rely on a constant-sized envelope in several places. For example, how would pointer manipulations such as in UsrToEnv work with this approach?

PhilMiller:
*Original date: 2014-01-21 23:00:53*

---
The packing pragma is a static compile-time change in the size of the objects. For every declared instance of affected types, it determines whether there will be suitable end-of-object padding to align a subsequent instance of the same type in an array. So, the envelope would be a constant size everywhere, but that size might end up being smaller.

stwhite91:
*Original date: 2016-08-30 15:43:30*

---
Should we revisit compiler support for this now that 64bit ID is just about ready to be merged?

PhilMiller:
*Original date: 2016-08-30 16:13:38*

---
[...]

PR #3692, opened 2023-03-03
Converse: Fix and cleanup padding

rbuch:
Use alignas instead of error-prone explicit padding with char[].

### DIFF
--- a/src/ck-core/envelope.h
+++ b/src/ck-core/envelope.h
@@ -200,13 +200,8 @@ namespace ck {
   UShort epIdx;        /* Entry point to call */                               \
   ck::impl::s_attribs attribs;
 
-class envelope {
+class alignas(ALIGN_BYTES) envelope {
 private:
-
-    class envelopeSizeHelper {
-      CMK_ENVELOPE_FIELDS
-      CMK_ENVELOPE_FT_FIELDS
-    };
 
     #ifdef __clang__
     #pragma GCC diagnostic push
@@ -220,19 +215,6 @@ private:
 public:
 
     CMK_ENVELOPE_FT_FIELDS
-
-    #if defined(__GNUC__) || defined(__clang__)
-    #pragma GCC diagnostic push
-    #pragma GCC diagnostic ignored "-Wpedantic"
-    #if defined(__clang__)
-    #pragma GCC diagnostic ignored "-Wunused-private-field"
-    #endif
-    #endif
-    // padding to ensure ALIGN_BYTES alignment
-    UChar align[CkMsgAlignOffset(sizeof(envelopeSizeHelper))];
-    #if defined(__GNUC__) || defined(__clang__)
-    #pragma GCC diagnostic pop
-    #endif
 
     void pup(PUP::er &p);
 #if CMK_REPLAYSYSTEM || CMK_TRACE_ENABLED
--- a/src/ck-core/envelope.h
+++ b/src/ck-core/envelope.h
@@ -200,6 +200,9 @@ namespace ck {
   UShort epIdx;        /* Entry point to call */                               \
   ck::impl::s_attribs attribs;
 
+// alignas is used for padding here, rather than for alignment of the envelope
+// itself, to ensure that the message following the envelope is aligned relative
+// to the start of the envelope.
 class alignas(ALIGN_BYTES) envelope {
 private:
 
--- a/src/conv-core/cmishmem.h
+++ b/src/conv-core/cmishmem.h
@@ -33,28 +33,18 @@ constexpr auto defaultTimeout = 4;
 }  // namespace ipc
 }  // namespace cmi
 
-// TODO ( find better names than src/dst? )
-#define CMK_IPC_BLOCK_FIELDS \
-  int src;                   \
-  std::uintptr_t orig;       \
-  int dst;                   \
-  std::uintptr_t next;       \
-  std::size_t size;
-
-struct CmiIpcBlock {
+struct alignas(ALIGN_BYTES) CmiIpcBlock {
+  // TODO ( find better names than src/dst? )
+public:
   // "home" rank of the block
- private:
-  class blockSizeHelper_ {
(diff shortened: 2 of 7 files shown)


src/conv-core/converse.h
```diff
@@ -780,32 +780,18 @@ CpvExtern(int, _curRestartPhase);      /* number of restarts */
 /** This header goes before each chunk of memory allocated with CmiAlloc. 
     See the comment in convcore.C for details on the fields.
 */
-struct CmiChunkHeader {
+struct
+#if !(CMK_USE_IBVERBS | CMK_USE_IBUD)
+  alignas(ALIGN_BYTES) // If in verbs mode, align in infiCmiChunkHeader instead
+#endif
+CmiChunkHeader {
   int size;
```
ericjbohm, line 788: Isn't leaving the CmiChunkHeader size field an int going to result in some buggy issues when you try to allocate large messages?

rbuch: Yes, it will, but this PR isn't meant to fix that. Rather, these are some underlying issues that are getting in the way of a larger change to use the right types to fix large messages.

I'm working on another PR to support large messages properly, this just came up as a problem while I was doing that.

ericjbohm (approved):

rbuch:
Converting back to draft as there are some interactions with the machine layer that I overlooked at first that still need to be addressed.

src/conv-core/cmishmem.h
```diff
@@ -33,28 +33,18 @@ constexpr auto defaultTimeout = 4;
...
+  int src;
+  std::uintptr_t orig;
+  int dst;
+  std::uintptr_t next;
+  std::size_t size;
 
   CmiIpcBlock(std::size_t size_, std::uintptr_t orig_)
       : orig(orig_), next(cmi::ipc::nil), size(size_) {}
-
-  char padding[(sizeof(blockSizeHelper_) % ALIGN_BYTES)];
```
jszaday, line 57: Last I checked, CMI-SHMEM wasn't regularly tested. Did you ensure it still works after this change?

rbuch: No, I did not, I'll test it before I un-draft it.

But I think the original implementation is wrong in the general case, e.g. if `sizeof(blockSizeHelper_) = 10` and `ALIGN_BYTES = 8`,  `sizeof(blockSizeHelper_) % ALIGN_BYTES = 2`, yielding a final size of 12 after padding rather than 16 (separately, using `alignas` also ensures that the alignment requirement (i.e. result of `alignof`) is set correctly). In any case, I think the way it's written right now will never actually cause a problem in practice, as every existing field will end up with a padded size of 8 on every reasonable architecture, but this cleans it up and makes the struct resilient to future modifications of the fields or their order.

rbuch: Also, while you're here, can you explain the `// "home" rank of the block` comment? It seems not to relate to anything immediately around it, so I was confused as to what it was supposed to refer to.

jszaday: Ah, I see. Yeah, that conclusion seems reasonable.

The comment was likely intended to be on `src` but not moved when the rest of the fields were placed in the struct.

jszaday: By the way, another instanceof this pattern is present in [`envelope.h`](https://github.com/UIUC-PPL/charm/blob/331c13afaa2a97851757c7a522c0e863a7cedfef/src/ck-core/envelope.h#L232). That's loosely where I copied the logic for this from, but I should've just used its macros or `CMIPADDING` instead.

rbuch: > By the way, another instanceof this pattern is present in [`envelope.h`](https://github.com/UIUC-PPL/charm/blob/331c13afaa2a97851757c7a522c0e863a7cedfef/src/ck-core/envelope.h#L232). That's loosely where I copied the logic for this from, but I should've just used its macros or `CMIPADDING` instead sweat_smile

Aha, my grep of variables called `padding` didn't catch that one, thanks for pointing it out, I'll update that here as well.

src/ck-core/envelope.h
```diff
@@ -200,14 +200,9 @@ namespace ck {
   UShort epIdx;        /* Entry point to call */                               \
   ck::impl::s_attribs attribs;
 
-class envelope {
+class alignas(ALIGN_BYTES) envelope {
```
stwhite91, line 203: I think this doesn't achieve the same effective alignment as before: 'alignas' here  in this PR will ensure that the envelope structure is laid out in memory beginning at a certain alignment, while the previous padding scheme was there to ensure that whatever structure follows this one (the user msg buffer) would be laid out in memory beginning at a certain alignment after the envelope. In other words, the padding was effectively there to ensure alignment not of the envelope itself but of what follows it.

The comments on this issue contain an illustration of the whole msg structure: https://github.com/UIUC-PPL/charm/issues/387

rbuch: I believe this should achieve the same effective alignment as before, `alignas` should both ensure that the envelope structure begins at a certain alignment _and_ that the size of the envelope structure is a multiple of the requested alignment, ergo the user message buffer will also start at the same alignment.

The way we use it, I think the property of beginning at a certain alignment is probably moot since we use things like `posix_memalign` to allocate memory without reading the alignment of the type.

stwhite91: That's right, but I think there should be a comment added by the alignas statement clarifying that its purpose is really to align the user buffer that follows the envelope.

Separate from this PR, I think ALIGN_BYTES should receive a CMI_ prefix.

rbuch: > That's right, but I think there should be a comment added by the alignas statement clarifying that its purpose is really to align the user buffer that follows the envelope.
> 
> Separate from this PR, I think ALIGN_BYTES should receive a CMI_ prefix.

Ah, yes, I agree with both of those, good idea. I'll update this PR and create another to update `ALIGN_BYTES`.

stwhite91 (approved):

PR #3705, opened 2023-04-15
Darwin: bump Mac OS X version minimum to 10.14

stwhite91:


evan-charmworks:
What benefits does this provide? We should document it in the commit message

evan-charmworks:
I believe this version requirement refers to the minimum supported Xcode / toolchain to build with this mode. However, the built binaries should be runnable on earlier versions.

That said, 10.7 is still quite old, so generating binaries compatible that far back may have impacts on the resulting performance. I don't know the answer and it needs research.

PR #3709, opened 2023-04-17
Bashify conditionals in charmc, ampicc, charmrun, buildold

evan-charmworks:
This should improve build performance by eliminating many process launches from each invocation of these scripts.

### DIFF
--- a/src/arch/pami-linux-ppc64le/charmrun
+++ b/src/arch/pami-linux-ppc64le/charmrun
@@ -23,7 +23,7 @@ runCmd()
   "$@"
 }
 
-while [ $# -gt 0 ]
+while [[ $# -ne 0 ]]
 do
 	case "$1" in
 	+n|++n|++np)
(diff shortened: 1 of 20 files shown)


src/arch/pami-linux-ppc64le/charmrun
```diff
@@ -242,11 +242,11 @@ then
 		ncpus="#PBS -l ncpus=$nodes"
 		;;
 	esac
-	if test $USE_LSF -eq 0
+	if [[ "$USE_LSF" = '0' ]]
```
matthiasdiener, line 245: Why not keep the `-eq` here?

matthiasdiener:
#3201 was a previous PR with a similar goal

src/arch/pami-linux-ppc64le/charmrun
evan-charmworks: Partly for consistency. Also, my thinking is that a string comparison is less work than parsing the variable to an integer before comparing.

matthiasdiener (approved):

stwhite91 (approved):

PR #3720, opened 2023-04-21
Fix CsdPeriodic Missing CCS Messages

ZwFink:
This fixes an issue introduced in #3537. In that PR, `CsdPeriodic` is only checked if the periodic heap is empty. This change means that conditions that use `CcdCallOnConditionKeep` with `CcdPERIODIC` would not be checked because `CcdCallOnConditionKeep` doesn't add periodic conditions to the periodic heap.  An example of this is CCS built on non-net layers. A socket is opened on PE 0 and checked using `CcdCallOnConditionKeep(CcdPERIODIC,(CcdCondFn)CcsServerCheck,NULL);`. 

As a first pass, this reverts the change to `CsdPeriodic`. However, it's not clear to me that `CsdPeriodic` is the problem. Should `CcdCallOnConditionKeep` add the condition to the periodic heap, rather than appending to the list of conditionals?

### DIFF
--- a/src/conv-core/converse.h
+++ b/src/conv-core/converse.h
@@ -1103,7 +1103,7 @@ extern void  CcdCallBacks(void);
 #else
 CpvExtern(int, _ccd_numchecks);
 CpvExtern(int, _ccd_heaplen);
-#define CsdPeriodic() do{ if (CpvAccess(_ccd_heaplen) > 0 && CpvAccess(_ccd_numchecks)-- <= 0) CcdCallBacks(); } while(0)
+#define CsdPeriodic() do{ if (CpvAccess(_ccd_numchecks)-- <= 0) CcdCallBacks(); } while(0)
 #define CsdResetPeriodic()    CpvAccess(_ccd_numchecks) = 0
 #endif
 


stwhite91:
It's unfortunate that we don't have any tests to catch this issue. It would also be unfortunate to undo some of that PR's optimization work.

In general, CcdCallbacks are trying to support two different use cases IMO: periodic and conditional callbacks. To me, conditional callbacks fit much more naturally than periodic callbacks into the Charm programming model of asynchronous message-driven execution with non-preemptive scheduling. But, periodic callbacks exist and have some use cases. My PR attempted to separate out the storage of two kinds of callbacks more explicitly, and to optimize performance for conditional callbacks since they are more common. I do not remember that users could call `CcdCallOnCondition{Keep}` with periodic callbacks. Instead that they should call the `CcdCallFnAfter` routines, but it looks like there is no "Keep" variety of the latter.

My proposal would be to:
A) Add tests for the correctness of periodic callbacks.
B) Add new APIs for "keeping" periodic callbacks such as `CcdCallFnEvery`,`CcdCallFnEveryOnPe`, and `CcdCancelPeriodic` which would properly add these to the heap and manage them.
C) Add assertions to the new APIs that the condnum is between CcdPERIODIC and the max periodic value, and then either add the opposite assertions to the existing APIs (this would be a breaking API change), or check the condnum there and conditionally add it to the heap (no breaking change).

This all assumes that they *should* be added to the heap, which I think is correct without checking.

stwhite91:
I looked into this more, and the solution is not as simple as adding all "periodic" callbacks to the heap. The way that I would classify ccd callbacks as either conditional vs. periodic would be to base the definition upon whether they are timer-based or not. The current codebase does *not* use that same definition. Instead it defines them as:

1) Conditional callbacks: These are all callbacks registered through CcdCallOnCondition{Keep}{OnPE}. They can optionally be persistent or "kept", and they can be targeted at a specific PE. The keep variety can also be cancelled via their condnum and index, so we store them in an array-based list for random access. Interestingly, this category also includes timer-based callbacks that use the predefined time intervals such as CcdPERIODIC_{10ms,1s,10s,etc}. These are stored alongside true condition-based (non-timer-based, such as BEGIN_IDLE or END_IDLE) callbacks and have the same semantics with respect being cancellable or persistent.

2) Periodic callbacks: These are all callbacks registered through CcdCallFnAfter{OnPE}. These do not have any "keep" or cancellation semantics, so they are one-time-use only. Users can also specify arbitrary time intervals for these, compared to conditional time-based callbacks which must use the predefined CcdPERIODIC_{10ms,100ms,1s,10s,etc} time intervals. We store those in a heap sorted by the time they will next need to be triggered.

So, not all periodic callbacks are supposed to be in the heap. The quickest solution I could think of was to track not only the number of periodic callbacks in the heap but also the number of timer-based callbacks in the cond callbacks list. I've done this in PR #3723. Can you test it with CCS?

Longer term, we should analyze the use cases of conditional (non-timer-based) callbacks, timer-based persistent/cancellable callbacks, and timer-based non-persistent callbacks, and consider changes to the API and implementation.

PR #3721, opened 2023-04-22
Fix issues found while experimenting with `-flto`

evan-charmworks:


evan-charmworks:
With this PR, adding `-flto` to the end of a `./build all-test` command completes successfully with no warnings on GCC, and with Clang while Fortran is disabled (since clang and gfortran don't understand each other's intermediate representation).

All tests pass except for `tlsglobals-f90` which crashes, but I have reproduced the bug with a modified test case on conventional MPI without any Charm/AMPI code or CmiTLS functionality and will document it.

Building `all-test` with LTO takes a very long time since it optimizes the entire runtime for each binary created, and for a case like integration testing it would be better to test LTO with `--build-shared -charm-shared`, which also coincidentally avoids the `tlsglobals-f90` crash. Though, splitting the code into shared objects reduces the extent of how LTO can be applied, and that is compounded by how many small shared objects we generate.

evan-charmworks:
I think I've scorched the earth of compiler warnings enough for one PR. Ready for review.

src/ck-ldb/ckgraph.C
```diff
@@ -62,8 +62,8 @@ ObjGraph::ObjGraph(BaseLB::LDStats *stats) {
       from = stats->getHash(commData.sender);
       to = stats->getHash(commData.receiver.get_destObj());
 
-      vertices[from].sendToList.push_back(Edge(to, commData.messages, commData.bytes));
-      vertices[to].recvFromList.push_back(Edge(from, commData.messages, commData.bytes));
+      vertices[from].sendToList.push_back(CkEdge(to, commData.messages, commData.bytes));
```
stwhite91, line 65: emplace_back?

src/ck-core/ckrdma.C
```diff
@@ -2064,11 +2064,15 @@ void CkPostBufferInternal(void *destBuffer, size_t destSize, int tag) {
 void CkPostNodeBufferInternal(void *destBuffer, size_t destSize, int tag) {
 
   // check if tag exists in posted req node table
+  CmiLock(CksvAccess(_nodeZCPostReqLock));
```
stwhite91, line 2067: Just curious, what compiler / options pointed this out with LTO?

src/ck-ldb/ckgraph.C
evan-charmworks: Nice, done.

src/ck-core/ckrdma.C
evan-charmworks: I was building with GCC on x86_64 Linux, but the issue manifested as a crash in one of the ZC tests or examples and not a diagnostic message. The crash happened in the implementation of `std::unordered_map` and traced back to these `find()` calls.

examples/charm++/load_balancing/stencil3d/stencil3d.C
```diff
@@ -324,7 +324,7 @@ class Stencil: public CBase_Stencil {
         work = 10.0;
 
 #ifndef _MSC_VER
```
stwhite91, line 326: I would be in favor of just removing these, rather than commenting them out

evan-charmworks: Done

src/ck-core/ckmulticast.C
```diff
@@ -46,7 +46,7 @@ typedef unsigned char byte;
  * An instance of this class is stored in every mCastEntry object making it possible
  * to track redn operations on a per section basis all along the spanning tree.
  */
-class reductionInfo {
+class multicastRedInfo {
```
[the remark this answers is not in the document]
evan-charmworks: `sectionRedInfo` makes sense.

In the Charm repository, `redInfo` and `redMsg` outnumber `rednInfo` and `rednMsg`, while `rednMgr` is more frequent than `redMgr`.

stwhite91 (approved):

rbuch (approved):

PR #3722, opened 2023-04-23
CI: Add OFI

evan-charmworks:


evan-charmworks:
It looks like this fails due to a hang on exit. I have not yet reproduced it locally.

PR #3723, opened 2023-04-25
CcdCallbacks: manage timer-based cond callbacks properly

stwhite91:
Fix support for timer-based conditional callbacks that was broken by PR #3537,
while still minimizing timer calls when possible.

Also add a test for timer-based conditional callbacks.

### DIFF
--- a/src/conv-core/converse.h
+++ b/src/conv-core/converse.h
@@ -1096,14 +1096,21 @@ typedef void (*CmiStartFn)(int argc, char **argv);
   @addtogroup ConverseScheduler
   @{
 */
-extern void  CcdCallBacks(void);
 #if CSD_NO_PERIODIC
 #define CsdPeriodic()
 #define CsdResetPeriodic()
 #else
+extern void  CcdCallBacks(void);
+extern int CcdNumTimerCBs(void);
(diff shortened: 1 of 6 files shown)


evan-charmworks (approved):

src/conv-core/converse.h
```diff
@@ -1785,34 +1792,36 @@ typedef void (*CcdCondFn)(void *userParam);
 typedef void (*CcdVoidFn)(void *userParam,double curWallTime);
 
 /*CPU conditions*/
-#define CcdPROCESSOR_BEGIN_BUSY 0
-#define CcdPROCESSOR_END_IDLE 0 /*Synonym*/
-#define CcdPROCESSOR_BEGIN_IDLE 1
-#define CcdPROCESSOR_END_BUSY 1 /*Synonym*/
-#define CcdPROCESSOR_STILL_IDLE 2
-#define CcdPROCESSOR_LONG_IDLE 3
+#define CcdSCHEDLOOP            0
```
ericjbohm, line 1795: Ideally this whitespace reformatting would be in a different commit, but it's unlikely to create problems here.

ericjbohm (approved):

PR #3733, opened 2023-05-10
Avoid padding in converse message header due to alignment issues

adityapb:
Reordering members of converse message header to fix alignment padding


src/arch/netlrts/conv-common.h
```diff
@@ -23,7 +23,7 @@
 */
 #define CMK_MSG_HEADER_BASIC   CMK_MSG_HEADER_EXT
 
-#define CMK_MSG_HEADER_EXT_    CmiUInt2 d0,d1,d2,d3,hdl,type,xhdl,info,redID,rank; CmiInt4 root, size; CmiUInt1 zcMsgType:4, cmaMsgType:2, nokeep:1;
+#define CMK_MSG_HEADER_EXT_     CmiInt4 root, size; CmiUInt2 d0,d1,d2,d3,hdl,type,xhdl,info,redID,rank; CmiUInt1 zcMsgType:4, cmaMsgType:2, nokeep:1;
```
stwhite91, line 26: As pointed out in the comment above CMK_MSG_HEADER_BASIC machine-dgram.C must be kept consistent with the ordering of members here, so that code will need to be updated as well.

adityapb: I changed the order back to the original since there anyway isn't any alignment issue here

src/arch/netlrts/conv-common.h
```diff
@@ -23,7 +23,7 @@
 */
 #define CMK_MSG_HEADER_BASIC   CMK_MSG_HEADER_EXT
 
-#define CMK_MSG_HEADER_EXT_    CmiUInt2 d0,d1,d2,d3,hdl,type,xhdl,info,redID,rank; CmiInt4 root, size; CmiUInt1 zcMsgType:4, cmaMsgType:2, nokeep:1;
+#define CMK_MSG_HEADER_EXT_     CmiUInt2 d0,d1,d2,d3,hdl,type,xhdl,info,redID,rank; CmiInt4 root, size; CmiUInt1 zcMsgType:4, cmaMsgType:2, nokeep:1;
```
stwhite91, line 26: Can you remove the whitespace change? Otherwise LGTM

stwhite91 (approved):

evan-charmworks:
If you run `git grep CMK_MSG_HEADER`, there are more files that could benefit from this change.

stwhite91 (approved):

evan-charmworks (approved):

PR #3734, opened 2023-05-10
Converse: print at startup if build is using non-default features

stwhite91:


evan-charmworks (approved):

stwhite91:
This needs another review

ZwFink (approved):

PR #3740, opened 2023-06-21
release-update: upgrade hwloc to 2.8.0

ericjbohm:
Strip out references in hwloc/config/hwloc.m4 to a file that does not exist

 utils/hwloc/test-build-custom-topology.sh.in
In order to update to latest stable hwloc 2.8.0


ericjbohm:
Not hard to reproduce the build error, but the output is so cryptic I can't tell what the actual issue is.

ericjbohm:
Looks like -Werror=vla is getting passed down into the compiler used by the configure tests inside hwloc.  This causes a warning about a variable-length array used in the code in their C99 tests to be treated as a failure.  Which means hwloc thinks we aren't using a C99 compiler and so it bails out.

matthiasdiener:
> Looks like -Werror=vla is getting passed down into the compiler used by the configure tests inside hwloc. This causes a warning about a variable-length array used in the code in their C99 tests to be treated as a failure. Which means hwloc thinks we aren't using a C99 compiler and so it bails out.

I can reproduce this locally. Can we just drop the `-Werror=vla`?

evan-charmworks:
> Strip out references in hwloc/config/hwloc.m4 to a file that does not exist
> 
> utils/hwloc/test-build-custom-topology.sh.in In order to update to latest stable hwloc 2.8.0

The `update-hwloc.sh` script handles these. When a new error appears during an update, you can add it to one of the lists at the top of the file.

> Looks like -Werror=vla is getting passed down into the compiler used by the configure tests inside hwloc. This causes a warning about a variable-length array used in the code in their C99 tests to be treated as a failure. Which means hwloc thinks we aren't using a C99 compiler and so it bails out.

In the past, we've patched hwloc to disable this configure test. The instructions at the end of `update-hwloc.sh` say to review the git history of `hwloc/` and squash any such patch commits into the update. The commit in question no longer applies cleanly so I kept an updated one separate this time.

ericjbohm:
> > Strip out references in hwloc/config/hwloc.m4 to a file that does not exist
> > utils/hwloc/test-build-custom-topology.sh.in In order to update to latest stable hwloc 2.8.0
> 
> The `update-hwloc.sh` script handles these. When a new error appears during an update, you can add it to one of the lists at the top of the file.
> 
> > Looks like -Werror=vla is getting passed down into the compiler used by the configure tests inside hwloc. This causes a warning about a variable-length array used in the code in their C99 tests to be treated as a failure. Which means hwloc thinks we aren't using a C99 compiler and so it bails out.
> 
> In the past, we've patched hwloc to disable this configure test. The instructions at the end of `update-hwloc.sh` say to review the git history of `hwloc/` and squash any such patch commits into the update. The commit in question no longer applies cleanly so I kept an updated one separate this time.

All good points, I failed to understand all the elements of the script and the followup commits.

matthiasdiener (approved):

stwhite91 (approved):

PR #3741, opened 2023-06-22
CkIO Reading Capabilities

mjacob1002:
# High Level
This PR adds input capabilities to CkIO. Users would be able to specify a read session, which specifies the number of bytes and the offset from which the user will be reading from. CkIO will then read that data into memory using designated BufferChares that go to disk and fetch the data. Then, every time a user calls Ck::IO::read, the buffer chares will serve that data back to the user.

## Notable Design Choices/Features

1.  Reads are asynchronous by spawning OS-level threads to do the I/O(can be seen in the BufferChare constructor)
2. If the disk operation on a particular buffer chare isn't complete, CkIO is able to let Charm++ do other work by calling CthYield()

## Capabilities

1. Users can specify the specific bytes they plan on querying
4. Users can customize the number of BufferChares they'd like through the Options::num_readers field
5. Users are able pipeline their read requests naturally

## Potential drawbacks

1.  We bring all the data into memory. While this can be useful in cases where the user queries multiple times from CkIO, for large files, it can potentially put a constraint on how much data they want to read. We may want to explore a more MPI I/O approach, where we coalesce read requests, and then every couple of timesteps we launch a coalesced read from designated reader chares, and then serve the data to the chares that requested data


evan-charmworks:
- To fix the non-SMP build failures, find the line in `src/scripts/charmc` that says `if [[ "$M" == 'CkLoop' && "$CMK_SMP" != '1' && "$CMK_WINDOWS" != '1' ]]; then`, and try making it `if [[ ( "$M" == 'CkLoop' || "$M" == 'CkIO' ) && "$CMK_SMP" != '1' && "$CMK_WINDOWS" != '1' ]]; then` so that programs using CkIO link with `-pthread` iff the platform needs it.
- To fix the SMP build failures, try running `git rm tests/charm++/io_read/iotest` to delete a built binary of this test that was accidentally checked in.

mjacob1002:
Thanks @evan-charmworks! After applying those changes, now it's saying there are "Too many reconnection attempts" when trying to run my tests. Do you know why this is happening/how to fix this?

stwhite91:
You need to add $(TESTOPTS) to the end of all of the run commands in the Makefile, as in:

`$(call run, ./iotest +p4 4  $(TESTOPTS) )`

mjacob1002:
Thanks Sam! I pushed those changes, but it seems that I'm still getting the reconnection errors.

evan-charmworks:
Try moving the commented `dd` line and the `$(call` line from the iotest rule, which should only handle building the program, to the test rule, which is for running it.

mjacob1002:
wow, that was a huge help! Now I'm only failing the circleci smp, which is way better than every test XD

tests/charm++/io_read/Makefile
```diff
@@ -0,0 +1,21 @@
...
+iotest: iotest.ci iotest.C
+	$(CHARMC) iotest.ci
+	$(CHARMC) iotest.C -o $@ -module CkIO
+
+testp: iotest
+	$(call run, ./iotest +p$(P) $(P) $(TESTOPTS))
+
+smptest: iotest
+	$(call run, ./iotest 4 +p2 ++ppn 2 $(TESTOPTS))
+	$(call run, ./iotest 4 +p4 ++ppn 2 $(TESTOPTS) )
```
evan-charmworks, line 17:
```suggestion
iotest: iotest.ci iotest.C
	$(CHARMC) iotest.ci
	$(CHARMC) iotest.C -o $@ -module CkIO

test: iotest
	# dd if=/dev/urandom of=large_test.txt bs=64M count=16 iflag=fullblock
	$(call run, ./iotest +p2 4)
	$(call run, ./iotest +p4 4)

testp: iotest
	$(call run, ./iotest +p$(P) $(P))
	$(call run, ./iotest +p$(P) $$(( $(P) * 2 )))

smptest: iotest
	$(call run, ./iotest +p2 ++ppn 2 4)
	$(call run, ./iotest +p4 ++ppn 2 4)
```

evan-charmworks (commented):
- `$(TESTOPTS)` is actually inserted by the `run` macro so you don't need to include it here, sorry for the confusion. The earlier error was caused by including test commands in the build target.
- Don't remove the test command lines entirely, move them to the test target.

My suggested change should fix all of the above, plus improve some formatting and the test and testp rules.

mjacob1002:
Awesome, thank you so much Evan!

src/libs/ck-libs/io/ckio.C
```diff
@@ -98,14 +136,24 @@ namespace Ck { namespace IO {
             opts.basePE = 0;
           if (-1 == opts.skipPEs)
             opts.skipPEs = CkMyNodeSize();
-
+	  if(opts.num_readers == 0){
+		// CkPrintf("DEBUG: the num readers in opts struct is 0\n");
+		opts.num_readers = std::min(32, CkNumPes());
```
ZwFink, line 141: Why use 32 here? Why not 2? Why not 1 million? What does this number mean and why do you use it?

src/libs/ck-libs/io/ckio.C
```diff
@@ -389,6 +638,135 @@ namespace Ck { namespace IO {
...
+			#endif
+			double disk_read_before = CkWallTimer(); // get the before disk_read
+			std::future<char*> temp_buffer = std::async(std::launch::async, &BufferChares::readData, this);
+			_buffer = temp_buffer.share();
+			double disk_read_after = CkWallTimer(); // end disk time
+			double total_time = disk_read_after - disk_read_before;
+			CkCallback cb(CkReductionTarget(BufferChares, printTime), thisProxy[0]);
+			contribute(sizeof(double), &total_time, CkReduction::max_double, cb);
+		}
+		// deprecated function; not used any longer
```
ZwFink, line 685: If it's deprecated, we should remove it.

src/libs/ck-libs/io/ckio.C
```diff
@@ -389,6 +638,135 @@ namespace Ck { namespace IO {
...
+		void clearBuffer() {
+			char* buffer = _buffer.get();
+			delete[] buffer;
+		}
+
+		void printTime(double time_taken){
```
ZwFink, line 691: Should this be removed?

src/libs/ck-libs/io/ckio.ci
```diff
@@ -26,6 +29,53 @@ module CkIO_impl {
           entry void openFile(std::string name, CkCallback opened, Options opts);
           entry [reductiontarget] void fileOpened(FileToken file);
           entry [reductiontarget] void sessionComplete(FileToken file);
+	  // facilitates all of the read calls that will happen
+	  
+//	  entry void read(Session session, size_t bytes, size_t offset, CkCallback after_read);
```
ZwFink, line 34: Remove entry methods that are commented out

src/libs/ck-libs/io/ckio.h
```diff
@@ -1,18 +1,25 @@
...
+
+// bool operator==(const Ck::IO::Session& s1, const Ck::IO::Session& s2);
+
 namespace Ck { namespace IO {
   /// Note: The values in options are not currently a stable or working interface.
   /// Users should not set anything in them.
   struct Options {
     Options()
-      : peStripe(0), writeStripe(0), activePEs(-1), basePE(-1), skipPEs(-1)
+      : peStripe(0), writeStripe(0), activePEs(-1), basePE(-1), skipPEs(-1), read_stripe(0), num_readers(0)
```
ZwFink, line 22: We are mixing camel and snake cases here.

src/libs/ck-libs/io/ckio.h
```diff
@@ -70,11 +82,49 @@ namespace Ck { namespace IO {
...
   void close(File file, CkCallback closed);
+  
+  /**
+   * Prepare to read data from @arg file section specified by @arg bytes and @arg offset.
+   * This method will proceed to eagerly read all of the data in that window into memory
+   * for future read calls. After all the data is read in, the ready callback will be invoked.
+   * The ready callback will take in a SessionReadyMessage* that will contain the offset, the amount of bytes
+   * , and the buffer in the form of a vector<char>.
+   */
+  void startReadSession(File file, size_t bytes, size_t offset, CkCallback ready);
```
ZwFink, line 93: The publci API should be added to the documentation

tests/charm++/io_read/iotest.ci
```diff
@@ -0,0 +1,88 @@
...
+			}
+
+		}
+
+		// method that verifies the read is correct
+		entry void testRead(Ck::IO::ReadCompleteMsg* msg) {
+			serial {
+//				FILE* fp = fopen("parallel_test.out", "w+");
+//				fwrite(msg -> data, 1, msg -> bytes, fp);	
+//				fclose(fp);
```
ZwFink, line 37: Remove these comments

ZwFink (changes requested):
An initial pass.

src/libs/ck-libs/io/ckio.h
mjacob1002: Yeah the docs will be added after this gets merged presumably.

src/libs/ck-libs/io/ckio.C
mjacob1002: It's useful for debugging purposes - I'll just wrap it in a ifdef block for when we set DEBUG macro

src/libs/ck-libs/io/ckio.C
mjacob1002: done!

src/libs/ck-libs/io/ckio.ci
mjacob1002: Done!

tests/charm++/io_read/iotest.ci
mjacob1002: Done!

src/libs/ck-libs/io/ckio.h
mjacob1002: Fixed it to all camel case within the Options struct.

src/libs/ck-libs/io/ckio.C
mjacob1002: This is more of a guess like thing, where we don't want to create too many BufferChares. It's very much a number that should be determined after tuning, so I just set 32 as a placeholder. In reality, the user probably shouldn't even let this be set.

evan-charmworks:
It looks like the netlrts-win-x86_64 build has found an error in the code.

```
make -C io_read test OPTS='-optimize -production' TESTOPTS='++local'
make[2]: Entering directory '/d/a/charm/charm/netlrts-win-x86_64/tests/charm++/io_read'
# dd if=/dev/urandom of=large_test.txt bs=64M count=16 iflag=fullblock
../../../bin/testrun   ./iotest +p2 4 ++local
Charmrun> started all node programs in 0.043 seconds.
Converse/Charm++ Commit ID: 73082b2
Isomalloc> Synchronized global address space.
Charm++> scheduler running in netpoll mode.
Charm++> Running on 1 hosts (1 sockets x 2 cores x 1 PUs = 2-way SMP)
Charm++> cpu topology info is gathered in 0.000 seconds.
CharmLB> Load balancing instrumentation for communication is off.
Starting the reading
My offset for reader 0 is 0
My offset for reader 1 is 10
My offset for reader 2 is 20
My offset for reader 3 is 30
My offset for reader 4 is 40
My offset for reader 5 is 50
My offset for reader 6 is 60
My offset for reader 7 is 70
My offset for reader 8 is 80
My offset for reader 9 is 90
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: [0] Assertion "seq_data[i] == msg -> data[i]" failed in file iotest.ci line 30.

Fatal error on PE 0> [0] Assertion "seq_data[i] == msg -> data[i]" failed in file iotest.ci line 30.

real	0m0.125s
user	0m0.000s
sys	0m0.031s
make[2]: Leaving directory '/d/a/charm/charm/netlrts-win-x86_64/tests/charm++/io_read'
```

mathewjacob1002:
interesting - is there a reason would work on Linux but not on Windows?

> It looks like the netlrts-win-x86_64 build has found an error in the code.
> 
> ```
> make -C io_read test OPTS='-optimize -production' TESTOPTS='++local'
> make[2]: Entering directory '/d/a/charm/charm/netlrts-win-x86_64/tests/charm++/io_read'
> # dd if=/dev/urandom of=large_test.txt bs=64M count=16 iflag=fullblock
> ../../../bin/testrun   ./iotest +p2 4 ++local
> Charmrun> started all node programs in 0.043 seconds.
> Converse/Charm++ Commit ID: 73082b2
> Isomalloc> Synchronized global address space.
> Charm++> scheduler running in netpoll mode.
> CharmLB> Load balancer assumes all CPUs are same.
> Charm++> Running on 1 hosts (1 sockets x 2 cores x 1 PUs = 2-way SMP)
> Charm++> cpu topology info is gathered in 0.000 seconds.
> CharmLB> Load balancing instrumentation for communication is off.
> Starting the reading
> My offset for reader 0 is 0
> My offset for reader 1 is 10
> My offset for reader 2 is 20
> My offset for reader 3 is 30
> My offset for reader 4 is 40
> My offset for reader 5 is 50
> My offset for reader 6 is 60
> My offset for reader 7 is 70
> My offset for reader 8 is 80
> My offset for reader 9 is 90
> ------------- Processor 0 Exiting: Called CmiAbort ------------
> Reason: [0] Assertion "seq_data[i] == msg -> data[i]" failed in file iotest.ci line 30.
> 
> Fatal error on PE 0> [0] Assertion "seq_data[i] == msg -> data[i]" failed in file iotest.ci line 30.
> 
> real	0m0.125s
> user	0m0.000s
> sys	0m0.031s
> make[2]: Leaving directory '/d/a/charm/charm/netlrts-win-x86_64/tests/charm++/io_read'
> ```

src/libs/ck-libs/io/ckio.C
stwhite91: I would guess that a reasonable default would be something related to CmiNumNodes() or CmiNumPhysicalNodes(). Maybe different depending on whether SMP mode is being used or not.

evan-charmworks:
My first guess in this situation, where file I/O works everywhere but Windows, is that you need to specify the binary flag when opening the file so that the system doesn't mangle line ending bytes, which Windows does but others do not. Add a second argument to any `std::ifstream` constructors: `std::ios::in | std::ios::binary`

src/libs/ck-libs/io/ckio.h
mjacob1002: ok, I'll get on that.

src/libs/ck-libs/io/ckio.C
```diff
@@ -138,7 +138,7 @@ namespace Ck { namespace IO {
             opts.skipPEs = CkMyNodeSize();
 	  if(opts.numReaders == 0){
 		// CkPrintf("DEBUG: the num readers in opts struct is 0\n");
-		opts.numReaders = std::min(32, CkNumPes());
+		opts.numReaders = std::min(CmiNumNodes(), CkNumPes());
```
evan-charmworks: I don't know if it's possible for the number of PEs to be less than the number of nodes. You might be able to make this just `CkNumNodes()`.

evan-charmworks (commented):
Congratulations on passing the CI!

Some notes on formatting:

- We don't keep spaces around `->` when accessing a member of a struct from a pointer. When surrounded by spaces it reads more like the `=>` operator from C# lambdas.
- We prefer to use spaces in C/C++ files instead of tabs.

mjacob1002:
3rd time's the charm: https://github.com/UIUC-PPL/charm/pull/3788

PR #3751, opened 2023-07-19
buildcmake: add --disable-fortran option, restructure

matthiasdiener:


ericjbohm (changes requested):
Functions as advertised, but has an odd effect in that if you build AMPI --disable-fortran it will make the usual target wrappers like ampif90, but they won't actually work.

ericjbohm:
> Functions as advertised, but has an odd effect in that if you build AMPI --disable-fortran it will make the usual target wrappers like ampif90, but they won't actually work.

Fixed in 682c3d1

Issue #3755, opened 2023-07-28
HAPI block mapping gets "Fatal CUDA Error [101] invalid device ordinal"


trquinn:
This will happen with `+gpumap block` if the number of PEs is not an integer multiple of the number of GPUs.  E.g., if there are 3 GPUs and 127 PEs, PE 126 will attempt call cudaSetDevice(3),

PR #3756, opened 2023-07-29
hapiMapping(): handle PEs not a multiple of GPUs.

trquinn:
Fixes issue #3755


evan-charmworks (approved):

stwhite91 (approved):

minitu:
This should work but for better load balance we should split PEs more equally - e.g., with 8 PEs and 3 GPUs, this PR will do a 2/2/4 split but something like 3/3/2 would be more ideal.

PR #3758, opened 2023-08-01
CMake updates for CUDA

dutta-alankar:
I have encountered a few problems while trying to build Charm using `CMake`. I'm entirely new to this. I did a few changes to the `CMakeLists.txt` to be able to compile Charm. Also Charm++ was able to compile with CUDA 11.3 but not CUDA 12.0 because of some missing `_v2` functions. 
I have just pushed the changes I made that enabled me to compile Charm++ with CUDA 11.3. Since I'm just starting out, I'm not sure f developers would find this useful. Starting a PR nevertheless.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 # - the buildcmake script
 # - files in cmake/
 
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.20)
 
 # Print warning
 message("##################################################################")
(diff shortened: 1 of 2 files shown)


evan-charmworks:
Thanks for the PR. Can you explain why you raised the minimum version of CMake? I don't think this should be necessary.

dutta-alankar:
> Thanks for the PR. Can you explain why you raised the minimum version of CMake? I don't think this should be necessary.

@evan-charmworks I was getting a few deprecation warnings with my version of `CMake` also there were a few errors, as I mentioned, because of which I was unable to compile Charm with CUDA. I went to `CMake` documentation and saw that they are encouraging usage of new keywords. So I got rid off the old ones. Those removed the warnings. However, the inclusion of a CUDA dependent include dir was needed in ROMIO `CmakeLists.txt` otherwise the code was unable to use ampicc to compile and failed in the Configure stage of ROMIO.

evan-charmworks:
Thanks for bringing these issues to our attention. I've addressed them in #3759.

Issue #3762, opened 2023-08-25
Bad memory accesses in springCleaning()


trquinn:
Running ChaNGa under valgrind reports that CkArray::springCleaning() is accessing freed memory.
This is with ChaNGa version 3.5 commit v3.5-11-gc7ba57c0 and charm version v7.1.0-devel-321-g606459e74
This is built on an AMD/infiniband machine with mpi-linux-x86_64-smp with gcc v11.2.0 and mvapchi2 2.3.6

Soon after writing an output (using CkIO) valgrind reports errors like:
`
==3563697== Invalid read of size 8
==3563697==    at 0x7DC2DB: CkArray::staticSpringCleaning(void*) (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x88A663: CcdRaiseCondition (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x88AD15: CcdCallBacks (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x88EE0B: CsdScheduleForever (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x88F234: CsdScheduler (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x8DDD46: ConverseRunPE(int) (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x8DDD8A: call_startfn(void*) (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x50A91CE: start_thread (in /usr/lib64/libpthread-2.28.so)
==3563697==    by 0x6CDBE72: clone (in /usr/lib64/libc-2.28.so)
==3563697==  Address 0x848f768 is 1,016 bytes inside a block of size 1,024 free'd
==3563697==    at 0x4C4AB30: free (in /apps/spack/anvil/apps/valgrind/3.15.0-gcc-11.2.0-u7tvx2t/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==3563697==    by 0x7DF84B: CkIndex_CkArray::_call_ckDestroy_void(void*, void*) (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
==3563697==    by 0x7D4ABF: CkDeliverMessageFree (in /anvil/scratch/x-trq/testdata/ChaNGa.smp)
[...]

PR #3765, opened 2023-09-22
ckio: set AnytimeMigration to false for WriteSession

### LINKED ISSUE #3678: Reductions in CkIO::WriteSession not completing
This is with Charm version 7.0.0 and ChaNGa.
I am running on frontera with the mpi build of charm on 40 nodes.

I am occasionally getting hangs during the use of Ck::IO.  Sticking in CkPrintf()s narrows down the hanging to two different different places:
1) sessionReady(), which is set as an InitCallback, will sometimes not get called after the CProxy_WriteSession::ckNew().
2) (More commonly) the reduction triggered by contribute at the end of WriteSession::forwardData() never completes, and WriteSession::syncData() is never called.

At the moment I'm reproducing this from a checkpoint.  I'm checking now if I can reproduce it from a non-checkpoint restart.

If someone has an account on frontera, I can point them at the simulation.

trquinn:
This prevents springCleaning on the WriteSession array, and works around possible problems after the array gets deleted.

This works around the springCleaning problem reported in #3762.
It may also fix #3678, but testing this is still in progress.

### DIFF
--- a/src/libs/ck-libs/io/ckio.C
+++ b/src/libs/ck-libs/io/ckio.C
@@ -125,6 +125,7 @@ namespace Ck { namespace IO {
 
           CkArrayOptions sessionOpts(numStripes);
           sessionOpts.setStaticInsertion(true);
+          sessionOpts.setAnytimeMigration(false);
 
           CkCallback sessionInitDone(CkIndex_Director::sessionReady(NULL), thisProxy);
           sessionInitDone.setRefnum(sessionID);


trquinn:
Another ChaNGa user reports that this fixes #3762

ericjbohm (approved):

mjacob1002:
LGTM!

mjacob1002 (approved):
LGTM!

PR #3764, opened 2023-09-04
add autobuild on delta

matthiasdiener:

PR #3768, opened 2023-10-10
add a gitaction for nightly build

ericjbohm:
Simple first cut of a gitaction to trigger nightly build on delta


matthiasdiener:
In general, this looks fine; 
- Why have two separate files for nightly builds?
- What's the difference to #3764 ?

PR #3769, opened 2023-10-10
Fix CUDA hello example and compilation with HAPI_TRACE

minitu:
Still needs a way to pass `HAPI_TRACE` as defined to the header and source files in `src/arch/cuda/hybridAPI`.
I think this can be done by modifying the `CMakeLists.txt` (and maybe a couple of other places), but would need help from the Charm++ folks.


matthiasdiener:
Something like this in the main CMakeLists.txt might work, adding after line https://github.com/UIUC-PPL/charm/blob/520585628155479ae7f50df671e21fc084274a46/CMakeLists.txt#L705:

```cmake
if(TRACING)
  target_compile_definitions(cudahybridapi PRIVATE HAPI_TRACE)
endif()
```

minitu:
@matthiasdiener Confirmed that does the right thing. Thanks!

matthiasdiener (approved):

trquinn (approved):
Tested with changa w/ projections: it runs now.

PR #3780, opened 2024-02-14
Doc: Use ++n instead of +p in charmrun examples 

### LINKED ISSUE #2947: charmrun examples should use ++n instead of +p
The example charmrun command lines at https://charm.readthedocs.io/en/latest/charm++/manual.html#running-charm-programs still use the old +p style even though ++n is now recommended. Given that most users scan for the examples rather than reading through all of the text, the charmrun examples (especially the smp examples) should all use ++n.

Also, the final paragraph in the SMP Options section discusses the +p option without mentioning the ++n option. The ++n option should be prominently and frequently discussed in the SMP Options section, in particular reminding people about it when the ++ppn option is introduced.

hizv:
Resolves #2947.

- Keeps +p in some examples (e.g. ./hello) to show that it's fine to use +p for testing.
- Adds an example to illustrate how +p relates to ++n in the SMP options section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ executable named `nqueen`.
 
 Following the previous example, to run the program on two processors, type
 
-     $ ./charmrun +p2 ./nqueen 12 6
+     $ ./charmrun ++n 2 ./nqueen 12 6
 
 This should run for a few seconds, and print out:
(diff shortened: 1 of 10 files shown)


hizv:
Should AMPI examples retain +p? The introduction section of AMPI uses P=4 in a diagram, so it might be easier to relate the +p option to that.

PR #3788, opened 2024-03-13
CkIO Reading Capabilities

mjacob1002:
This PR will add high performance reading capabilities to CkIO. These performance gains have been verified through experiments on Bridges2. It is a session-based capability, where a user opens a read session, in which data is then read in and the user can then ask for such data through a Ck::IO::read call, where they specify both an offset and amount of data to read. 

Will add a well-documented example, as well as readthedocs, later.


mtaylo12:
Commit [] will make small implementation adjustment to better support concurrent background work:

The previous implementation monitors the status of a buffer chare's read in `sendData` (which is called every time a client requests data from a buffer chare). To avoid spinning on the future's value and hogging resources, `sendData` is a threaded method and our implementation yields the waiting thread. If `# clients >> # buffer chares`, this implementation does not work as intended, because an excessive number of threads are created and suspended for a single read future.


### Projections Analysis ###
The following projections screenshots show this issue in more detail. These results were produced on Bridges2, on 1 node with 2 cores, using non-SMP mode. I used 1 buffer chare and 2048 clients reading 1 GB for these preliminary results. The green denotes dummy background work and the light red denotes time spent in sendData. The old implementation shows that handling the large number of threads occupies an entire core, blocking any background work that could use that resource:

Screenshot 2024-03-18 at 9 41 52 AM

src/libs/ck-libs/io/Makefile
```diff
@@ -18,7 +18,6 @@ $(LIB): $(LIBOBJ)
 
 headers: $(HEADERS)
 	cp $(HEADERS) $(CDIR)/include/
-	touch headers
```
matthiasdiener, line 21: What's the reason for this change?

mjacob1002: It didn't appear it was actually used for anything useful when we looked - we can add it back if needed?

matthiasdiener: It probably creates the `headers` file so that the target "headers" two lines above isn't executed multiple times. I'd suggest adding it back, just in case.

mjacob1002: Resolved!

mtaylo12 (approved):
Looks good, thanks @mjacob1002

src/libs/ck-libs/io/ckio.C
```diff
@@ -1,436 +1,995 @@
...
-        {
+  int procNum(int arrayHdl, const CkArrayIndex& element)
+  {
 #if 0
           int peIndex = stripeIndex % opts.activePEs;
           int pe = opts.basePE + peIndex * opts.skipPEs;
 #endif
-          return 0;
-        }
+    return 0;
```
ericjbohm, line 562: Not clear from context why we're mapping all elements to 0.

ericjbohm (approved):

src/libs/ck-libs/io/ckio.C
mjacob1002: this is code from the output part written years ago - the git diff is being weird

trquinn:
This commit undid a fix (work around, actually) committed in PR #3765.  ChaNGa now crashes after using CkIO.

PR #3791, opened 2024-03-25
feature: support CXI in OFI for Slingshot-11

ericjbohm:
This supports the CXI interface for Cassini (AKA Slingshot-11) within the OFI machine layer.  No application level changes are required, but a variety of command line options are provided to configure the memory pool and the selection of cxi interfaces.

Note: this does require the use of the memory pool in order to efficiently support the FI_MR_ENDPOINT mode of memory registration. As the time cost of registering individual messages would be otherwise be entirely too high.

Charmrun has been configured to wrap srun and currently assumes PMI2 with cray extensions for launching.

The build system has been set up to autodetect CXI and enable support for it accordingly.  For compatibility purposes, it also supports the use of cxi on the build line, but that should not be necessary on most HPE systems with proper LMOD environments.


src/arch/ofi/machine-onesided.C
```diff
@@ -142,8 +150,9 @@ void process_onesided_reg_and_put(struct fi_cq_tagged_entry *e, OFIRequest *req)
                        ncpyOpInfo->srcSize);
 
   ncpyOpInfo->isSrcRegistered = 1; // Set isSrcRegistered to 1 after registration
-
-  const char *rbuf  = (FI_MR_SCALABLE == context.mr_mode) ? 0 : (const char*)(ncpyOpInfo->destPtr);
+  // if FI_MR_VIRT_ADDR is not enabled, then we use offset from the
+  // registered address which would be a 0 offset in this case.
+  const char *rbuf  = ((FI_MR_SCALABLE == context.mr_mode) || ( FI_MR_VIRT_ADDR & context.mr_mode)==0) ? 0 : (const char*)(ncpyOpInfo->destPtr);
```
stwhite91, line 155: This logic is repeated multiple times below- should pull it out into a static function or macro

src/arch/ofi/machine.C
```diff
@@ -195,17 +315,19 @@ static inline int process_completion_queue();
...
  *  - mr: Address of memory region; Sent back as part of OFIRmaAck
+ *  - orig_msg: actual address of source message; Sent back as part of OFIRmaAck
  */
 typedef struct OFIRmaHeader {
     uint64_t src_msg;
     uint64_t len;
     uint64_t key;
     uint64_t mr;
     int      nodeNo;
+    uint64_t orig_msg;
```
stwhite91, line 330: Could swap nodeNo and orig_msg declaration order and potentially save 32 bits on the size of OFIRmaHeader from padding/alignment

src/arch/ofi/machine.C
```diff
@@ -87,37 +160,84 @@
...
+#elif CMK_USE_PMI || CMK_USE_SIMPLEPMI
 #include "runtime-pmi.C"
 #elif CMK_USE_PMI2
 #include "runtime-pmi2.C"
 #elif CMK_USE_PMIX
 #include "runtime-pmix.C"
 #endif
 
+/* this should be set for us during the build, but cmake confounds again */
+#define CMK_CXI 1
```
stwhite91, line 184: Should probably be fixed

src/arch/ofi/machine.C
ericjbohm: good idea. swapped in latest commit

src/arch/ofi/machine.C
ericjbohm: Latest push has changes to build and buildold to address this.

src/arch/ofi/machine-onesided.C
ericjbohm: True.  This was a pattern inherited from the initial OFI implementation, so I left it alone as it is slightly ugly, but doesn't seem to have performance implications.

stwhite91: I was thinking for readability/maintainability rather than performance, but up to you

ericjbohm: implemented DETERMINE_OFFSET macro for this

src/arch/ofi/machine.C
```diff
@@ -1104,19 +1792,28 @@ void process_long_send_ack(struct fi_cq_tagged_entry *e, OFIRequest *req)
      * An OFIRmaAck was received; Close memory region and free original msg.
```
jcphill, line 1792: The memory region is no longer closed (commented out in code below). Either code or comment is wrong.

src/arch/ofi/machine.C
```diff
@@ -1290,14 +2011,19 @@ void LrtsPreCommonInit(int everReturn)
 {
     MACHSTATE(2, "OFI::LrtsPreCommonInit {");
 
-    PRINT_THREAD_INFO();
+    //    PRINT_THREAD_INFO();
```
jcphill, line 2014: Why is this commented out?

jcphill (commented):
Looks good but obviously too big to be sure it's correct. Couple of small comments.

CMakeLists.txt
lvkale, line 1: Add a comment after line 157 saying why the option needs to be OFF by default.  ?

src/arch/ofi/machine-onesided.C
```diff
@@ -232,18 +243,29 @@ void LrtsIssueRget(NcpyOperationInfo *ncpyOpInfo) {
     req->size     = ncpyOpInfo->ncpyOpInfoSize;
     req->callback = send_short_callback;
     req->data.short_msg = ncpyOpInfo;
-
-    ofi_send(ncpyOpInfo,
+    // in CXI we cannot just send this unregistered thing
+    //
+#if CMK_OFI_CXI
```
adityapb, line 248: Where is `CMK_OFI_CXI` defined?

ericjbohm: That was replaced by CMK_CXI, due to it being impractical to get CMK_OFI_CXI to play nice with the cmake build process.   Though clearly I missed the two calls to it outside of machine.C.  !fixup posted to address this.

CMakeLists.txt
ericjbohm: Comment added.

src/arch/ofi/machine.C
ericjbohm: Restored in cleanup commit.

src/arch/ofi/machine.C
ericjbohm: Comment was out of date, the close is not connect there.  The free will either return it to the pool, or truly free and close if the buffer is out of pool.

src/arch/ofi/machine.C
```diff
@@ -1079,22 +1761,28 @@ void process_long_recv(struct fi_cq_tagged_entry *e, OFIRequest *req)
...
-        OFI_RETRY(fi_read(context.ep,
+
+	OFI_RETRY(fi_read(context.ep,
                           lbuf,
                           chunk_size,
-                          (mr) ? fi_mr_desc(mr) : NULL,
+			    fi_mr_desc(long_msg->mr),
                           nodeNo,
-                          rbuf,
+                          rmsg,
```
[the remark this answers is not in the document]
ericjbohm: Probably just a stray debugging experiment.  I don't think it matters in the CXI case, but would be an issue in the non-CXI case.  Made it consistent in the !fixup.

src/arch/ofi/conv-common.h
```diff
@@ -91,11 +91,11 @@
...
-
+#define CMK_USE_CRAYPMI                                 0
 /*
  * Use Simple client-side implementation of PMI.
  * Valid only for CMK_USE_PMI.
  * Optional in an SLURM environment.
  * See src/arch/util/proc_management/simple_pmi/
  */
-#define CMK_USE_SIMPLEPMI                               1
+#define CMK_USE_SIMPLEPMI                               0
```
stwhite91, line 101: Does this get set somewhere else now in the default OFI case?

stwhite91 (commented):
Any documentation updates needed?

Has non-CXI OFI been tested and benchmarked for performance with the changes to use the LRTS mempool and the ofi request cache?

ericjbohm:
The last machine that I had access to with regular OFI was turned off last
year.  I could try it over TCP/IP somewhere, but not sure how useful the
resulting information would be.

adityapb (approved):
LGTM

src/arch/ofi/conv-common.h
ericjbohm: I'm not sure if that was a good to have.  It seems to me that it may be better to have the builder choose their pmi on the build line.  Simplepmi will often compile just fine, but may not run launch the way the user expects.  I've seen it result in a non parallel launch (as in a bunch of independent processes), which is arguably worse.

ritvikrao (approved):
finally got a chance to look at this. everything looks all set to merge

cmake/detect-features-c.cmake
```diff
@@ -367,9 +367,43 @@ int main() {
...
-if(${CMK_BUILD_OFI} EQUAL 1)
+if(${NETWORK} STREQUAL "ofi" OR ${NETWORK} STREQUAL "ofi-crayshasta" OR ${NETWORK} STREQUAL "ofi-linux")
+# assume HPC installation
+include(CMakePrintHelpers)
+  find_package(EnvModules REQUIRED)
+  find_package(PkgConfig REQUIRED)
+  if(EnvModules_FOUND)
+  #at least get libfabric loaded if it isn't already
+	env_module(load libfabric)
+  endif()
```
RMeli, line 378: I think this assumption is too restrictive. It does not seem to work when building Charm++ with Spack, since `libfabric` is provided as a Spack package and not through modules.

ericjbohm: It makes sense for the standard case of building on a DOE machine, but could present a problem for Spack.  Is there is a simple way to test for the Spack case and then extract the necessary information to accomplish the same smooth build this code accomplishes with LMOD and PkgConfig?

RMeli: I think the simplest way is to setup Charm++ as a develop build and introduce changes to the build system and Spack package until it works. I started working on it but got stuck and then other stuff took priority.

Issue #3795, opened 2024-03-27
Build failure for charm4py build on mac/darwin


ritvikrao:
When attempting to build the charm4py-compatible version of charm++ on mac, the build fails. The failure message is the following:

```
clang: error: no such file or directory: ' '
Fatal Error by charmc in directory /Users/ritvik/charm4py/charm_src/charm/netlrts-darwin-x86_64
   Command clang++ -framework Foundation -framework IOKit -multiply_defined suppress -stdlib=libc++ -D_DARWIN_C_SOURCE -mmacosx-version-min=10.7 -ftls-model=initial-exec -rdynamic -fpic -shared CMakeFiles/charm.dir/empty.cpp.o  -Wl,-all_load -Wl,-search_paths_first -Wl,-headerpad_max_install_names lib/libck.a lib/libconverse.a lib/libmemory-default.a lib/libthreads-default.a lib/libldb-rand.a -Llib/   -lz -lm -lstdc++ -o lib/libcharm.dylib returned error code 1
charmc exiting...
make[2]: *** [lib/libcharm.dylib] Error 1
make[1]: *** [CMakeFiles/charm.dir/all] Error 2
```

The error happens during "Linking CXX shared library lib/libcharm.dylib".

To reproduce:
1. Use any Mac machine (x86 or apple silicon)
2. Run the following:
```
$ git clone https://github.com/UIUC-PPL/charm4py.git && cd charm4py
$ mkdir charm_src && cd charm_src
$ git clone https://github.com/UIUC-PPL/charm.git && cd charm
$ ./build charm4py netlrts-darwin-<relevant architecture> -j8 --with-production
```

The error will appear at ~86% of the build or so. I suspect this issue is related to the way the libcharm library file is built based on cmakelists.txt:969:

```
if(${TARGET} STREQUAL "charm4py")
  # Create libcharm.so for Charm4py
  # add_compile_options(-build-shared)
  file(WRITE empty.cpp "// This file left intentionally blank. It is used for the charm4py build.")
  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
  add_library(charm SHARED empty.cpp)
  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib_so)
  target_link_libraries(charm ck converse memory-default threads-default ldb-rand "-Llib/ -standalone -whole-archive -c++stl -shared")

[...]

PR #3796, opened 2024-03-27
bugfix: libcharm build for charm4py should use CMK_SHARED_SUF

ericjbohm:
The makefile hardcoded the filename to libcharm.so, should be libcharm.$CMK_SHARED_SUF to support darwin or other architectures with targets that are not .so.


matthiasdiener:
This PR contains unrelated changes. [95b3371](https://github.com/UIUC-PPL/charm/pull/3796/commits/95b3371c6e48d92e8821be70cb56a828d0b9c5e4) looks good, but I think #3795 is an unrelated issue (#3795 shows an issue with the CMake build system, which does not use `src/scripts/Makefile`).

matthiasdiener (changes requested):
See my comment

ericjbohm:
> See my comment ([#3796 (comment)](https://github.com/UIUC-PPL/charm/pull/3796#issuecomment-2024075902))

Well, someone made the same mistake over in the cmake, but I don't  know where that is.  In any case, this should give people a workaround via buildold.

ericjbohm:
> This PR contains unrelated changes. [95b3371](https://github.com/UIUC-PPL/charm/pull/3796/commits/95b3371c6e48d92e8821be70cb56a828d0b9c5e4) looks good, but I think #3795 is an unrelated issue (#3795 shows an issue with the CMake build system, which does not use `src/scripts/Makefile`).

The unrelated changes are due to my apparently having done this on the wrong branch. No idea why I even had the ofi cxi branch checked out on this laptop.

ericjbohm:
change was made to the wrong branch

matthiasdiener:
> > See my comment ([#3796 (comment)](https://github.com/UIUC-PPL/charm/pull/3796#issuecomment-2024075902))
> 
> Well, someone made the same mistake over in the cmake, 

It's a different issue, in #3795 there is already the correct .dylib extension.

PR #3797, opened 2024-03-27
bugfix: use CMK_SHARED_SUF instead of a hardcoded .so

ericjbohm:
Address buildold issue for libcharm as used by charm4py on Darwin

### DIFF
--- a/src/scripts/Makefile
+++ b/src/scripts/Makefile
@@ -82,7 +82,7 @@ AMPIC AMPIF: AMPI
 f90charm: charm++
 
 libcharm: charm++
-	cd $(L) ; $(CHARMC) -standalone -whole-archive -c++stl -shared -o libcharm.so $(LIBCHARM_LIBS)
+	cd $(L) ; $(CHARMC) -standalone -whole-archive -c++stl -shared -o libcharm.$(CMK_SHARED_SUF) $(LIBCHARM_LIBS)
 
 charm4py: libcharm
 


ritvikrao (commented):
This workaround only allows charm++ to build if I comment out a couple of lines in setup/scripts/configure.ac that blocks charm4py build. I added that commit. But even then, now I get an error when running setup.py in charm4py, saying that there's no CmiCommitId in the library file.

src/scripts/configure.ac
```diff
@@ -874,11 +874,11 @@ EOF
...
-  if test -z $got_opt_whole_archive
-  then
-    echo "Don't know how to build libcharm.so for Charm4py"
-    test_finish 1
-  fi
+  # if test -z $got_opt_whole_archive
+  # then
+    # echo "Don't know how to build libcharm.so for Charm4py"
+    # test_finish 1
+  # fi
```
matthiasdiener, line 881: Is commenting this out necessary? This while-archive options seems important.

ritvikrao: This blocks the build on my machine if it isn't commented out. I haven't determined if this is the source of the setup.py issue where the build doesn't have a commit id.

ritvikrao:
On macs, the flag for linking whole archives is different than with linux, so I had to change the flag in CMakeLists.

src/scripts/configure.ac
ritvikrao: just looked, the whole archive is needed for the build to work, it's fixed (for the cmake build) in the new commit

ritvikrao:
sending this message to get more eyes on it; it's a trivial flag change for darwin

matthiasdiener (approved):
LGTM, thanks! (besides the (unrelated?) CI errors)

ZwFink (approved):

PR #3802, opened 2024-04-10
bugfix: change test to +p 2 +vp 8, because +p 2 +vp 1 is nonsense

ericjbohm:
Fix an old typo in the makefile for the megaampi test.

### DIFF
--- a/tests/ampi/megampi/Makefile
+++ b/tests/ampi/megampi/Makefile
@@ -20,7 +20,7 @@ test: megampi
 	$(call run, ./megampi +p1 +vp4 +balancer RandCentLB)
 	$(call run, ./megampi +p2 +vp2 +balancer RandCentLB)
 	$(call run, ./megampi +p2 +vp4 +balancer RandCentLB)
-	$(call run, ./megampi +p2 +vp1 +balancer RandCentLB)
+	$(call run, ./megampi +p2 +vp8 +balancer RandCentLB)
 
 testp: megampi
 	$(call run, ./megampi +p$(P) +vp$(P) )


stwhite91:
+p2 +vp1 is often the most simple way to reproduce a migration/PUP bug in AMPI. I don't have a problem with removing it from megampi so long as we still have it in tests/ampi/migration though. Also, +p2 +vp8 shouldn't add anything meaningful to the testing that +p2 +vp4 isn't already covering so I'd rather just remove the +p2 +vp1 and not add anything new.

ericjbohm:
> +p2 +vp1 is often the most simple way to reproduce a migration/PUP bug in AMPI. I don't have a problem with removing it from megampi so long as we still have it in tests/ampi/migration though. Also, +p2 +vp8 shouldn't add anything meaningful to the testing that +p2 +vp4 isn't already covering so I'd rather just remove the +p2 +vp1 and not add anything new.

I finally got back to looking at this.  There seems to be an issue with this independent from anything going on in CXI.  A multicore build running on the headnode of frontier has a similar crash in the pieglobals-f90 +p1 +vp2 +balancer RandCentLB.  

So, I think there is a more fundamental issue with how that interacts with newer toolchains.  For reference, that bug was with PrgEnv-gnu/8.3.3 loaded.  AMPI is of course totally unstable with the cray compiler, but even GNU is having issues with this test.

stwhite91:
> pieglobals-f90 +p1 +vp2 +balancer RandCentLB

I assume you mean `+p2 +vp1`

ericjbohm:
I had initially thought it to be the same problem, but it seems to be more
general than +p2 +vp1, or there may be more than one problem with AMPI
virtualization, as this is +p1 +vp2.

This bug is:
munmap_chunk(): invalid pointer

ericjbohm:
closing this, will open an issue regarding some oddities with AMPI virtualiation

Issue #1521, opened 2017-04-21
CkIO file read support


stwhite91:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/1521*

---
CkIO currently only supports writing files.

Branch rohan/ckioread: https://charm.cs.illinois.edu/gerrit/gitweb?p=charm.git;a=shortlog;h=refs/heads/rohan/ckioread

stwhite91:
*Original date: 2017-04-21 02:41:13*

---
Semi-related stackoverflow question: http://stackoverflow.com/questions/43272225/file-handling-support-in-charm

ericjbohm:
Add documentation for readthedocs describing how to use the new CkIO features.

PR #3805, opened 2024-04-23
CkIO documentation and cleanup

mayantaylor:
Adding CkIO Input documentation to the Charm++ readthedocs and updating relevant comments to address #1521 . This should be basically complete for the current status of CkIO in main. Mat implemented a FileReader streaming abstraction as well (not merged yet) - should that be merged in for this release? In this case, we should add these docs.

### DIFF
--- a/doc/libraries/manual.rst
+++ b/doc/libraries/manual.rst
@@ -826,7 +826,7 @@ using the library are different for input and output, but follow the same basic 
 
 #. Open a file via ``Ck::IO::open``.
 #. Create a session for writing to the file via ``Ck::IO::startSession`` or create a session for reading from a file via ``Ck::IO::startReadSession``.
-#. Write or read via ``Ck::IO::write`` or ``Ck::IO::write``. Note that these function take a
+#. Write or read via ``Ck::IO::write`` or ``Ck::IO::read``. Note that these function take a
    session token that is passed into the callback, which should refer to the current session.
 #. In the case of a read session, the session must be closed manually when the read is complete via ``Ck::IO::closeReadSession``
 #. When the specified amount of data for the session has been written or a read session has been closed, a
@@ -967,8 +967,6 @@ The following functions comprise the interface to the library for parallel file 
 
 - Reading data:
 
-  Note there are three variants of the ``read`` call.
-
   .. code-block:: c++
 
     void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read);
@@ -978,13 +976,6 @@ The following functions comprise the interface to the library for parallel file 
   after_read callback is invoked taking a ReadCompleteMsg* which points to a vector<char> buffer, the offset,
   and the number of bytes of the read.
 
-  .. code-block:: c++
-
-    void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read, size_t tag);
-
-  .. code-block:: c++
-
-    void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read, size_t tag, char* user_buffer);
 
 - Closing a file:
 
@@ -996,6 +987,7 @@ The following functions comprise the interface to the library for parallel file 
   been closed. Note that ``file`` is provided as a member of
   the ``FileReadyMsg`` sent to the ``opened`` callback after a file has been
   opened. This method should only be called from a single PE, once per file.
+
 
 Examples
 --------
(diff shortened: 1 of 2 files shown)


doc/libraries/manual.rst
```diff
@@ -809,22 +813,30 @@ entire stripe is actually written to the filesystem all in one fell swoop. The
...
-in an earlier step):
+using the library are different for input and output, but follow the same basic structure:
 
 #. Open a file via ``Ck::IO::open``.
-#. Create a session for writing to the file via ``Ck::IO::startSession``.
-#. Write to the file via ``Ck::IO::write``. Note that this function takes a
-   session token that is passed into the callback.
-#. When the specified amount of data for the session has been written, a
+#. Create a session for writing to the file via ``Ck::IO::startSession`` or create a session for reading from a file via ``Ck::IO::startReadSession``.
+#. Write or read via ``Ck::IO::write`` or ``Ck::IO::write``. Note that these function take a
```
mayantaylor, line 829: typo, should be write or read

doc/libraries/manual.rst
```diff
@@ -909,8 +921,85 @@ The following functions comprise the interface to the library:
...
+    void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read);
+
+  This method is invoked to read data from the read session. If the data is not yet available, this request
+  will be buffered until the Buffer Chares can respond with the requested data. After the read finishes, the 
+  after_read callback is invoked taking a ReadCompleteMsg* which points to a vector<char> buffer, the offset,
+  and the number of bytes of the read.
+
+  .. code-block:: c++
+
+    void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read, size_t tag);
```
mayantaylor, line 983: remove

mayantaylor:
Decided not to add FileReader impl/docs in this PR. Will add future PR for this work.

doc/libraries/manual.rst
```diff
@@ -909,8 +921,77 @@ The following functions comprise the interface to the library:
...
+  This can be useful when the user has a specific decomposition in mind for the read.
+
+- Reading data:
+
+  .. code-block:: c++
+
+    void read(Session session, size_t bytes, size_t offset, char* data, CkCallback after_read);
+
+  This method is invoked to read data from the read session. If the data is not yet available, this request
+  will be buffered until the Buffer Chares can respond with the requested data. After the read finishes, the
```
ericjbohm, line 975: I think there is a slight ambiguity here in that the user should rely on the callback regardless.  The function returns to the user when the request is submitted,  regardless of whether the data is already available, unlike a typical synchronous read call.

ericjbohm (changes requested):
Mostly good, one minor comment posted to clarify expected behavior.

doc/libraries/manual.rst
mayantaylor: I've updated this description. Let me know if it needs further clarification.

ericjbohm (approved):

doc/libraries/manual.rst
ericjbohm: looks good

doc/libraries/manual.rst
```diff
@@ -809,22 +813,30 @@ entire stripe is actually written to the filesystem all in one fell swoop. The
 size and layout of stripes and the number and organization of aggregating PEs
 are available as options for the user to customize.
 
+CkIO Input
+----------
+
+The CkIO input library similarly aggregates read requests for a single file via an intermediate layer of chares, called "Buffer Chares." The number of Buffer Chares should be chosen to read from the file system with optimal granularity. Currently, the choice of the number of Buffer Bhares must be made by the user (via the Options parameter, discussed below), considering factors such as file size, number of PEs, and number of nodes.
```
stwhite91, line 819: typo: "Buffer Bhares" -> "Buffer Chares"

stwhite91 (commented):
Commit msg should be updated to say CkIO documentation and cleanup

mjacob1002 (commented):
We should also mention that one requirement to use CkIO input is that the file must be seekable. This is how the buffer chares find where they should read the data from and thus it won't work if the file is not seekable. Just my two cents though

mayantaylor:
> We should also mention that one requirement to use CkIO input is that the file must be seekable. This is how the buffer chares find where they should read the data from and thus it won't work if the file is not seekable. Just my two cents though

Added a sentence for this disclaimer under the CkIO Overview.

mjacob1002 (approved):
LGTM

PR #3808, opened 2024-05-01
fix Mac CI

matthiasdiener:
`macos-latest` is now referring to Mac M1, not x86_64, so we need to use an older version to build for x86_64.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -148,26 +148,27 @@ jobs:
         files=$(find . -name *.sts)
         for f in $files; do echo $f; pushd .; cd $(dirname $f); $proj --exit $(basename $f); popd; done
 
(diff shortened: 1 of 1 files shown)


ericjbohm (approved):
LGTM

ritvikrao:
it says it's still waiting for the spack_macos-latest check. maybe that needs to be commented out?

matthiasdiener:
> it says it's still waiting for the spack_macos-latest check. maybe that needs to be commented out?

Should be fixed now (the test was renamed to spack_macos-12)

ritvikrao (approved):
everything looks good

ritvikrao:
looks like the parameter server test needs to be shortened

matthiasdiener:
> looks like the parameter server test needs to be shortened

Sure, can you do this in a separate PR?

PR #3810, opened 2024-05-08
bugfix: use larger mempool parameters to avoid running out of pinnables

ericjbohm:
Very wide (i.e., 128 pes per node) nonsmp runs of message intensive applications ran into message allocation errors with the prior memory pool settings.  These appear as failures to fi_enable with an error message that maps to "address already in use".  The latter appears to be a misleading error message as the actual bug appears to be an exhaustion of the number of pinned segments that can be enabled onto the fabric. 

This can be avoided by increasing the expand size we use in the memory pool and related quantities for the overall sizing of the pool.  Thereby reducing the total quantity of base addresses being pinned, bound, and enabled.


ritvikrao (approved):
I tested this pr with the ofi-linux-x86 build on delta, and it fixed all my bugs, so it looks good to me

matthiasdiener (approved):

PR #3811, opened 2024-05-09
bugfix: craycc now uses -v instead of -V

ericjbohm:
Cray (HPE) has switched over to clang/LLVM, so the command line interface for their compiler is different.

This just replaces -V with -v and a test for the string Version with version.  Nonetheless, it will allow a simple multicore-linux-x86_64 craycc build to compile properly on frontier with the cray compilers. This is desirable for NAMD GPU Resident targets to use newer versions of cce, rocm, and hip modules.


ritvikrao (approved):
looks good to me

ZwFink (approved):

mayantaylor (approved):

PR #3814, opened 2024-05-15
Charm 7.0.1

ericjbohm:
Change License to Apache 2.0 with LLVM exceptions.
Otherwise unchanged from 7.0.0

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,20 @@
 This file describes the most significant changes. For more detail, use
 'git log' on a clone of the charm repository.
 
+===============================================================================
+What's new in Charm++ 7.0.1
+================================================================================
+
+This is a change license only release, with the following change:
+
+Highlights:
+
+- The License has changed to Apache 2.0 with LLVM exception.  This
+  change to a popular Open Source license is intended to simplify use,
+  collaboration, and greater community involvement in the development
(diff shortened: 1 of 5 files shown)


CHANGES
```diff
@@ -1,6 +1,20 @@
 This file describes the most significant changes. For more detail, use
 'git log' on a clone of the charm repository.
 
+===============================================================================
+What's new in Charm++ 7.0.1
+================================================================================
+
+This is a change license only release, with the following change:
```
matthiasdiener, line 8: maybe:
```suggestion
This is a license-only release, with the following change:
```

ericjbohm:
> Not sure how to set this up as a PR, merging onto main looks like a lot of changes, but this looks like we're trying to make a change to 7.0.

For that matter, not sure if needs to be a PR, since it already exists as a branch in the repo, so anyone can just check it out and use it as normal.

matthiasdiener:
> Not sure how to set this up as a PR, merging onto main looks like a lot of changes, but this looks like we're trying to make a change to 7.0.

Maybe make the version number changes in this PR on top of main, then create a new branch `charm-7.0.1` off of `charm-7.0.0` and just cherry pick the version number commit into charm-7.0.1`.

ericjbohm:
That is how I created the 7.0.1 branch.

ericjbohm:
no PR required.

Issue #3834, opened 2024-08-14
Molecular2D example doesn't compile


ritvikrao:
In examples/charm++/Molecular2D, which includes entry methods which take CkVec objects as parameters, when attempting to compile, the following error is thrown:

```
In file included from Patch.C:403:
./Patch.def.h:765:26: error: non-const lvalue reference to type 'CkVec<...>' cannot bind to a temporary of type 'CkVec<...>'
  impl_obj->updateForces(std::move(particles.t));
```

mayantaylor:
PR #2886 might be relevant

stwhite91:
How was this unnoticed for so long? Is Molecular2d not compiled by default as part of "make test"? If not, it should probably be added to that...

ritvikrao:
Make test only compiles the tests, right? This is in the examples folder. Should examples also be added to the list of tests?

ericjbohm:
The examples/charm++/Makefile only operates on a specific enumerated list
of directories and Molecular2d is not in that list.

PR #3835, opened 2024-08-14
Molecular2D example switch from CkVec to std::vector

ritvikrao:
Uses std::vector in place of CkVec in the Molecular2D example, fixes the bug (#3834) in the compilation of Molecular2D by using const references in the method parameters of updateParticles and updateForces, and updates the documentation to clarify how pup_stl.h should be used.

### DIFF
--- a/doc/charisma/manual.rst
+++ b/doc/charisma/manual.rst
@@ -661,7 +661,7 @@ for the 6D array.
 
 .. code-block:: c++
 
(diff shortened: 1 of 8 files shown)


ericjbohm (approved):

sanjaykale (approved):
Looks good.

lvkale (approved):
good.

PR #3858, opened 2024-12-05
allow craype for any targets (i.e., multicore)

ericjbohm:
On newer machines with extensive layers of compilation (i.e,. delta-ai's grace/hopper supernodes) it can be difficult to build a proper binary using individual compiler targets.  Our craype target delegates those issues to whatever the user has chosen within their craype module selection, but has previously been restricted to cray specific interconnect builds. There really isn't a good reason to do that.  This change frees our craype wrapper for use with other targets, such as the multicore, or netlrts, builds.


kavithachandrasekar:
Would it be helpful to remove duplicate CMK_BUILD_CRAY=1 and CMK_CRAY_NOGNI=1 from files like arch/ofi-crayshasta/conv-mach.sh, arch/mpi-crayshasta/conv-mach.sh, arch/mpi-crayxc/conv-mach.sh and arch/mpi-crayxe/conv-mach.sh as part of this PR?

PR #3859, opened 2024-12-05
feature: support ofi-linux-arm8

ericjbohm:
Support OFI with Cassini extensions on the arm8 architecture.

Basically a copy what we did on x86 with the minimum changes necessary for arm.


green-br:
Is there anything holding this PR up?  It would be great to see this committed.

PR #3864, opened 2025-01-20
bugfix: due to bugs in the ubuntu-latest mpich install, use 22.04

ericjbohm:
Work around the fact that the MPICH install in ubuntu-latest has some wacky misconfiguration that makes it launch an application like a bunch of independent processes all at rank 0.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,7 +101,7 @@ jobs:
 
   mpi-linux-x86_64:
     timeout-minutes: 90
-    runs-on: ubuntu-24.04  #  ubuntu-latest does not work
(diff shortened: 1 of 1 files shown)


kavithachandrasekar (commented):
Would it be better to update it to use ubuntu-22.04 instead? One of the recently passing mpi builds (in PR 3858) seems to use 22.04. I believe ubuntu-latest now points to 24.04. Otherwise, looks good to me.

matthiasdiener:
Would it make sense to switch to OpenMPI instead?

kavithachandrasekar (approved):

matthiasdiener (approved):

ericjbohm:
bundled into #3871

PR #3865, opened 2025-01-23
bugfix: avoid spack race condition

ericjbohm:
set -j2 to avoid race condition between ROMIO config and AMPI


ericjbohm:
not a race condition, so this is not going to help anything

PR #3866, opened 2025-01-23
update AMPI and ROMIO to be compatible gcc14 for spack

ericjbohm:
Fix a gcc 14 "bug" in the ROMIO configure test by explicit return type of int for main function.

Reconcile incompatible function prototypes to the one in the MPI standard.

### DIFF
--- a/src/libs/ck-libs/ampi/romio/configure
+++ b/src/libs/ck-libs/ampi/romio/configure
@@ -23719,7 +23719,7 @@ $as_echo_n "checking if a simple MPI program compiles and links... " >&6; }
   rm -f mpitest.c
   cat > mpitest.c <<EOF
 #include "mpi.h"
-     main(int argc, char **argv)
+     int main(int argc, char **argv)
      {
          MPI_Init(&argc,&argv);
          MPI_Finalize();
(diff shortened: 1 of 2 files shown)


ericjbohm:
I managed to resolve a few of the GCC 14 issues, but I'm not sure what to do about the multiple definitions of MPI_Info.

stwhite91:
Can you provide more info on the multiple definition of MPI_Info issue?

stwhite91:
It looks like the offending code in ROMIO's header is protected by a macro HAVE_MPI_INFO which should be defined to 1 since AMPI defines all of the MPI_Info interface. So my guess is that ROMIO's configure script is not testing/setting HAVE_MPI_INFO correctly

ericjbohm:
> It looks like the offending code in ROMIO's header is protected by a macro HAVE_MPI_INFO which should be defined to 1 since AMPI defines all of the MPI_Info interface. So my guess is that ROMIO's configure script is not testing/setting HAVE_MPI_INFO correctly

That is part of the problem,  but is basically just the tip of the iceberg.  Once you get past that there is a cascade of bugs all stemming from having multiple types (MPI_Info, MPI_Request, MPIO_Request, MPI_Status) all being basically aliases of int.  Along with some slightly sketchy ptr to int conversions.  In prior versions of gcc it didn't care about you assigning differently named types back and forth without explicit casts.  In GCC 14 it cares a lot, arguably too much as they're all int under the hood.

I tried liberally sprinkling some explicit casts where necessary and could get past a lot of it, but it feels like a bottomless well of debugging that should be solvable by a better type definition scheme.

stwhite91:
I see, yeah this is a problem because ROMIO is tailored to MPICH, and MPICH's MPI types are pointers. OpenMPI's MPI types are ints akin to AMPI's. It would take some effort, but if there is a desire to improve AMPI's IO functionality I think looking into OpenMPI's ompio library or its version of ROMIO v3.4.1 might be a good starting point: https://github.com/open-mpi/ompi/commit/bb7396248655cecfd0542ddb7b5904064bc13677

PR #3867, opened 2025-02-03
Initial Stream Setup Working?

mjacob1002:
TODO: 
- [ ] Clean up the debugging code and make sure the variable names _actually correspond_ with what the variable is supposed to represent.
- [ ] Update the doc with the protocol and make it detailed for sends, receives, and closing of streams. Specifically make sure we take care of _all code paths_, mainly what happens when the stream is closed and there are buffered get requests and alike.


zhuangy47 (approved):
LGTM

PR #3870, opened 2025-02-07
bugfix: avert ci failures by disabling ROMIO by default

ericjbohm:
GCC 14's strict adherence to types is not compatible with our ROMIO implementation's use (AKA abuse) of the old tolerance for casting everything to int.

ROMIO is only necessary for AMPI applications relying on MPI IO, which appears to be the null set at this time.

This PR disables ROMIO by default in both build and buildold.  

A better solution would be a re-implementation of ROMIO, but given the lack of user demand, that is not a current priority.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -597,8 +597,9 @@ if(NOT CMK_AMPI_ONLY)
 endif()
 
 #disable ROMIO due to GCC 14 cascade of failures
-set(CMK_AMPI_WITH_ROMIO 0)
-
+if(NOT DEFINED CMK_AMPI_WITH_ROMIO)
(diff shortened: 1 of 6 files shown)


matthiasdiener (approved):

stwhite91 (commented):
It would be nice to document the need to build with --with-ampi-romio (or whatever the option is for that) for MPI I/O support in the README.ampi file and in the AMPI manual

ericjbohm:
> It would be nice to document the need to build with --with-ampi-romio (or whatever the option is for that) for MPI I/O support in the README.ampi file and in the AMPI manual

Good point.  see 9366727 and 221514c

ericjbohm:
bundled into #3871

PR #3873, opened 2025-02-12
cleanup: re-enable ChaNGa test after checkpoint fix

ericjbohm:


matthiasdiener (approved):

trquinn (approved):
I also checked that this test is "valgrind clean".  Looks good!

PR #3882, opened 2025-04-02
cmake min to 3.12

ritvikrao:
CMake 4.0 now throws an error if the minimum specified version is below 3.5, so this updates the version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,19 +334,10 @@ set(CMK_HAS_PYTHON 0)
 set(CMK_PYTHON_VERSION "")
 set(CMK_BUILD_PYTHON "")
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 # - the buildcmake script
 # - files in cmake/
 
-cmake_minimum_required(VERSION 3.4...3.5)
+cmake_minimum_required(VERSION 3.10)
```
matthiasdiener, line 13: If we set the minimum version to 3.12, we could simplify a lot of the version-dependent code below.

ritvikrao: done

Sh0g0-1758 (approved):
LGTM

PR #3886, opened 2025-04-20
feat: add allGather library

Sh0g0-1758:
cc: @lvkale @mayantaylor @ericjbohm


src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,47 @@
+# collectiveSim
+
+A library for implementing collectives commonly used in machine learning tasks including allGather and allReduce.
+
+## allGather
+
+allGather lets you gather data distributed accross different chare array elements. The library provides 3 algorithms for doing the allGather operations, namely ring, hypercube and flooding.
```
mayantaylor, line 7: can we give more details on these three strategies? Just a couple notes on how each is implemented

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### How to use
+
+You can build the library using 
+```bash
+make allGather
+```
+
+After that you need to declare allGather as an extern module in your `.ci` file and include the `allGather.hh` header file in your `cc/hh` file. Then create an AllGather group object.
```
mayantaylor, line 16: do we want allGather to be built by default on a charm++ build? @lvkale

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```C++
+enum allGatherType {
+  ALL_GATHER_RING,
+  ALL_GATHER_HYPERCUBE,
+  ALL_GATHER_FLOODING
+};
+```
+
+##### [NOTE]: `ALL_GATHER_HYPERCUBE` switches to recursive doubling when n is not a power of 2 as described in [this paper](https://ieeexplore.ieee.org/abstract/document/342126?casa_token=vuF8Rhhm2f4AAAAA:TBigoTv8ge_lz8Bqt7wF0jWnyVrEXfPBL7cQGsWgnsXVZqEx3pFgtputZ8lvNma9pHjKAnR_pck5).
```
mayantaylor, line 32: this could go in a deeper explanation of each implementation, as I commented above

mayantaylor:
Can we do the following things:
- [ ] example in `examples/charm++`
- [ ] make sure the charm++ build also builds this library correctly (this might be happening already, but might require some cmake tweaks? not sure). This is also related to my comment about build. Do we want this to build by default?
- [ ] address the rest of my comments
- [ ] maybe have @ericjbohm look at the usage of `ncpy` functions

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+You must also declare a callback to a function which the library can return to after its done and it must only take a pointer to `allGatherMsg` as its argument. To start operations, each chare array element must make a local pointer to the library chare group element on the same PE as it.
+
+```C++
+AllGather *libptr = AllGatherGroup.ckLocalBranch();
+libptr->init(result, data, thisIndex, cb);
+```
+Here, result is a pointer to where the user wants the result of allGather operation to be stored(note that it must be of size n * k), data refers to per chare array element contributed data and cb refers to the callback. ALso, we must pass the `thisIndex` to `init` to place the gathered data in the right order.
```
stwhite91, line 40: Does the implementation guarantee that the result buffer is ordered in any way? The documentation should specify that either way here

src/libs/ck-libs/allGather/README.md
Sh0g0-1758: Done.

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+You can build the library using 
+```bash
+make allGather
+```
+
+After that you need to declare allGather as an extern module in your `.ci` file and include the `allGather.hh` header file in your `cc/hh` file. Then create an AllGather group object.
+
+```C++
+AllGather = CProxy_AllGather::ckNew(k, n, (int)allGatherType::ALL_GATHER_RING);
```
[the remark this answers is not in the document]
Sh0g0-1758: Removed n. Thanks for the catch!

src/libs/ck-libs/allGather/README.md
Sh0g0-1758: Done.

src/libs/ck-libs/allGather/README.md
Sh0g0-1758: The implementation does not constrain gather to a specific order. It depends on the index that the user has specified while calling init. So suppose if the 8th chare array element called init with 4 (in place of thisIndex) and the 4th chare array element called init with 8 (in place of thisIndex) then in the gathered data, the data at 4th and 8th place would be interchanged.

src/libs/ck-libs/allGather/README.md
lvkale: Yes, i think. If we want to start considering these operations as a part of Charm++, which I do, we should do that, and also document in readthedocs. Any counterpoint to that?

ericjbohm (changes requested):
Seems functional, but UNREG is unlikely to perform well in latency sensitive contexts.

examples/charm++/allGather/user.cc
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void simBox::begin(CProxy_AllGather AllGatherGroup)
+{
+  CkCallback cb(CkIndex_simBox::done(NULL), CkArrayIndex1D(thisIndex), thisProxy);
+  AllGather* libptr = AllGatherGroup.ckLocalBranch();
+  libptr->init(result, data, thisIndex, cb);
+}
+
+void simBox::done(allGatherMsg* msg)
```
stwhite91, line 71: msg is leaked. I think the best solution would be to mark the entry method's declaration in the .ci file with the [nokeep] attribute.

examples/charm++/allGather/user.cc
```diff
@@ -0,0 +1,111 @@
...
+  {
+    ckout << "[STATUS] Completed the AllGather Simulation" << endl;
+    CkExit();
+  }
+}
+
+simBox::simBox(CProxy_start startProxy, int k, int n, int x, int y)
+    : startProxy(startProxy), k(k), n(n), x(x), y(y)
+{
+  result = (long int*)malloc(k * n * sizeof(long int));
```
stwhite91, line 49: result and data are leaked, would be better to make them std::vector<long int>.

src/libs/ck-libs/allGather/allGather.hh
```diff
@@ -0,0 +1,65 @@
...
+  ALL_GATHER_FLOODING
+};
+
+class AllGather : public CBase_AllGather
+{
+private:
+  int k{};
+  int n{};
+  int idx{};
+  long int* store;
```
stwhite91, line 29: Can this and 'data' below be std::vector's for better memory management?

src/libs/ck-libs/allGather/allGather.hh
```diff
@@ -0,0 +1,65 @@
...
+  int numHypercubeIter{};
+  bool HypercubeRecursiveDoubling{};
+  int iter;
+  int HypercubeToSend;
+  std::vector<std::vector<int>> graph{};
+  std::map<int, bool> recvFloodMsg{};
+  int randCounter{};
+  std::vector<int> hyperCubeIndx{};
+  std::vector<CkNcpyBuffer> hyperCubeStore{};
+  allGatherMsg* msg = new allGatherMsg;
```
stwhite91, line 42: I think we generally prefer dynamic memory allocation be done in the constructor rather than here

src/libs/ck-libs/allGather/allGather.cc
```diff
@@ -0,0 +1,124 @@
...
+void AllGather::init(long int* result, long int* data, int idx, CkCallback cb)
+{
+  this->lib_done_callback = cb;
+  this->idx = idx;
+  zero_copy_callback =
+      CkCallback(CkIndex_AllGather::local_buff_done(NULL), thisProxy[CkMyPe()]);
+  dum_dum = CkCallback(CkCallback::ignore);
+  this->store = result;
+  this->data = data;
+  thisProxy[0].initdone();
```
stwhite91, line 68: I think this should be a no-op reduction rather than a point-to-point entry method invocation

examples/charm++/allGather/user.cc
Sh0g0-1758: no result and data cannot be vectors as the library expects a pointer to where the initial data is and a pointer to where it should store the data when done.

src/libs/ck-libs/allGather/allGather.cc
Sh0g0-1758: thanks for the suggestion. Helped me do away with a function :)

src/libs/ck-libs/allGather/allGather.ci
```diff
@@ -0,0 +1,60 @@
...
+                                 std::vector<int> dataIndx)
+        {
+          serial
+          {
+            for (int m = 0; m < data.size(); m++)
+            {
+              hyperCubeStore.emplace_back(data[m]);
+              hyperCubeIndx.emplace_back(dataIndx[m]);
+              CkNcpyBuffer dst(store + dataIndx[m] * k, k * sizeof(long int),
+                               zero_copy_callback, CK_BUFFER_UNREG);
```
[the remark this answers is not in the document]
anant37289: Changed it to use the PREREG API with the user now having to declare both buffers using `CkRdmaAlloc`

mayantaylor (approved):

.github/workflows/ci.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
         ./build charm4py netlrts-linux-x86_64 tcp -g -j4 --with-production
     - name: build-charm4py
       run: |
-        pip3 install setuptools cython cffi greenlet numpy torch torchvision
+        pip3 install setuptools cython cffi greenlet numpy torch torchvision matplotlib
```
stwhite91, line 54: Remove

examples/charm++/allGather/Makefile
```diff
@@ -0,0 +1,20 @@
+CHARMC = ../../../bin/charmc $(OPTS)
+FLAGS = -g -Wall -Wno-unused-variable -Wno-sign-compare -Wno-class-memaccess -O2
```
stwhite91, line 2: Can you make this Makefile consistent with others in Charm where the compiler flags can be passed into it rather than hardcoded here?

examples/charm++/allGather/Makefile
```diff
@@ -0,0 +1,20 @@
...
+TYPE ?= RING
+
+all: build
+
+build:
+	$(CHARMC) $(FLAGS) ./user.ci
+	$(CHARMC) $(FLAGS) -o allGather -D$(TYPE) ./user.cc -lmoduleallGather
+
+test: build
+	$(call run, +p19 ./allGather 19 29 7 )
```
stwhite91, line 12: +p19 is too large for default testing. I think we usually keep it to +p6 or +p4 by default

examples/charm++/allGather/user.cc
```diff
@@ -0,0 +1,108 @@
...
+    for (int i = 0; i < n * k; i++)
+    {
+      ckout << result[i] << " ";
+    }
+    ckout << endl;
+  }
+  CkRdmaFree(result);
+  CkRdmaFree(data);
+  CkCallback cbfini(CkReductionTarget(start, fini), startProxy);
+  contribute(0, 0, CkReduction::nop, cbfini);
```
stwhite91, line 105: This can be simplified further to `contribute(cbfini);`

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+In `ALL_GATHER_RING` the data for all the groups, starting from the originating group elements, gets forwarded in a ring(each element getting the data from lower `PE` group element and passing it to higher `PE` group element).
+
+In `ALL_GATHER_HYPERCUBE`, when `n` is a power of 2, the group elements assume a hypercube connectivity and use the standard [hypercube communication pattern](https://en.wikipedia.org/wiki/Hypercube_(communication_pattern)) .`ALL_GATHER_HYPERCUBE` switches to recursive doubling when n is not a power of 2 as described in [this paper](https://ieeexplore.ieee.org/abstract/document/342126?casa_token=vuF8Rhhm2f4AAAAA:TBigoTv8ge_lz8Bqt7wF0jWnyVrEXfPBL7cQGsWgnsXVZqEx3pFgtputZ8lvNma9pHjKAnR_pck5).
+
+In `ALL_GATHER_FLOODING`, we make a sparse graph over the group elements to specify the connectivity. The communication is done by each group element getting data from it's neighbours, keeping it and forwarding/flooding it to all it's neighbours () in case it has not already seen it and discard it otherwise.
+
+### Notes
+- Currently only gathering equal sized data is supported.
+- The number of PEs needs to be the same as `n`.
```
stwhite91, line 55: Since n has moved from the API to an internal call to CkNumPes() I don't think the user needs to be aware of this anymore

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,56 @@
...
+- The program still has an unresolved bug, where very rarely(2 in a 100 runs), the program reports a segfault after all the data has been correctly gathered.
```
stwhite91, line 56: Is this specific to one of the algorithms?

src/libs/ck-libs/allGather/allGather.cc
```diff
@@ -0,0 +1,118 @@
...
+{
+  this->lib_done_callback = cb;
+  this->idx = idx;
+  zero_copy_callback =
+      CkCallback(CkIndex_AllGather::local_buff_done(NULL), thisProxy[CkMyPe()]);
+  dum_dum = CkCallback(CkCallback::ignore);
+  this->store = result;
+  this->data = data;
+  CkCallback cbInitDone(CkReductionTarget(AllGather, startGather), thisProxy);
+  contribute(0,0, CkReduction::nop, cbInitDone);
```
stwhite91, line 62: `contribute(cbInitDone);`

src/libs/ck-libs/allGather/allGather.hh
```diff
@@ -0,0 +1,63 @@
...
+  bool HypercubeRecursiveDoubling{};
+  int iter;
+  int HypercubeToSend;
+  std::vector<std::vector<int>> graph{};
+  std::map<int, bool> recvFloodMsg{};
+  int randCounter;
+  std::vector<int> hyperCubeIndx{};
+  std::vector<CkNcpyBuffer> hyperCubeStore{};
+  allGatherMsg* msg;
+  long int* data;
```
stwhite91, line 43: We typically use `char *` to represent an array of bytes inside the runtime.  The user shouldn't need to cast their data to any particular type like `long int` when passing data into the library. Same for `store` above.

stwhite91 (commented):
Do we want to require that users use `CkRdmaAlloc()` to allocate their buffers or should we use the default `CK_BUFFER_REG` mode? There's a trade-off in usability vs performance. I believe `CK_BUFFER_REG` may check if the buffer is already pinned and do the right thing in the case that the user has used `CkRdmaAlloc` already... That way CkRdmaAlloc is a performance optimization, not a requirement.

.github/workflows/ci.yaml
ritvikrao: no, the charm4py tests will fail. unless you suggest removing the whole charm4py test?

src/libs/ck-libs/allGather/allGather.cc
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void AllGather::local_buff_done(CkDataMsg* m)
```
stwhite91, line 65: `m` is leaked

src/libs/ck-libs/allGather/README.md
anant37289: Here n is the number of participants in all gather, we can't have more or less than n participants. This is kept as a reminder that we should not be doing this.

src/libs/ck-libs/allGather/README.md
Sh0g0-1758: The error is specific to the flooding algorithm. 
image

stwhite91: Interesting, have you tried reproducing this on a multicore build of Charm++ with `-g -O0` compiler flags? That may be easier to debug if it reproduces there. In any case I would suggest removing this from the documentation and instead opening a github issue for it

src/libs/ck-libs/allGather/allGather.ci
```diff
@@ -0,0 +1,59 @@
...
+          if (HypercubeRecursiveDoubling)
+            HypercubeToSend = (n + CkMyPe() - (int)pow(2, iter)) % n;
+          else
+            HypercubeToSend = CkMyPe() ^ ((int)pow(2, iter));
+        }
+        serial
+        {
+          if (HypercubeRecursiveDoubling && iter == numHypercubeIter - 1)
+          {
+            std::vector<CkNcpyBuffer> hyperCubeStoreCopy(
```
stwhite91, line 28: If you pull these two vectors out into member variables of the AllGather group you could reuse their memory buffers between iterations (call .clear() and .resize() each iteration as needed).

src/libs/ck-libs/allGather/allGather.ci
```diff
@@ -0,0 +1,59 @@
...
+                hyperCubeStore.begin(), hyperCubeStore.begin() + (n - (int)pow(2, iter)));
+            std::vector<int> hyperCubeIndxCopy(
+                hyperCubeIndx.begin(), hyperCubeIndx.begin() + (n - (int)pow(2, iter)));
+            thisProxy[HypercubeToSend].recvHypercube(iter, hyperCubeStoreCopy,
+                                                     hyperCubeIndxCopy);
+          }
+          else
+            thisProxy[HypercubeToSend].recvHypercube(iter, hyperCubeStore, hyperCubeIndx);
+        }
+        when recvHypercube[iter](int ref, std::vector<CkNcpyBuffer> data,
```
stwhite91, line 38: It would be more efficient to send these vector's as C-style arrays to avoid copying from the Charm++ msg to a std::vector on the receive side. On the sender-side you can keep them as std::vector's and pass the .size() and .data() separately.

stwhite91 (commented):
Can you change all .hh files to .h and .cc to .C for consistency with the rest of Charm++? Otherwise it looks good to me!

src/libs/ck-libs/allGather/README.md
Sh0g0-1758: > have you tried reproducing this on a multicore build of Charm++ with -g -O0 compiler flags
No, not yet. 

>  would suggest removing this from the documentation and instead opening a github issue for it
sounds good. Tracking the issue in #3892 and marking this as resolved.

examples/charm++/allGather/user.cc
```diff
@@ -100,7 +100,6 @@ void simBox::done(allGatherMsg* msg)
     ckout << endl;
   }
   CkRdmaFree(result);
-  CkRdmaFree(data);
```
stwhite91: I think you may need to use an actual callback instead of the `dum_dum` / ignore callback you are currently using in some zero copy transfers to properly track completion of the transfer on both src & dst sides

Sh0g0-1758: Ah! Thanks, this would have been a nasty bug. Made the changes.

src/libs/ck-libs/allGather/allGather.cc
```diff
@@ -0,0 +1,114 @@
...
+  this->store = (char*)result;
+  this->data = (char*)data;
+  CkCallback cbInitDone(CkReductionTarget(AllGather, startGather), thisProxy);
+  contribute(cbInitDone);
+}
+
+void AllGather::local_buff_done(CkDataMsg* m)
+{
+  numRecvMsg++;
+  if (numRecvMsg == 2 * (n - 1))
```
stwhite91, line 66: Should this check be `if (numRecvMsg == 2*n)`? Also make sure numRecvMsg (and any other relevant state) is being reset to 0 in init() in case users perform multiple allGathers.

Sh0g0-1758: no each chare needs to send n - 1 messages and receive n - 1 messages. Made the changes to zero out state vars.

src/libs/ck-libs/allGather/allGather.ci
```diff
@@ -0,0 +1,59 @@
...
+            HypercubeToSend = (n + CkMyPe() - (int)pow(2, iter)) % n;
+          else
+            HypercubeToSend = CkMyPe() ^ ((int)pow(2, iter));
+        }
+        serial
+        {
+          if (HypercubeRecursiveDoubling && iter == numHypercubeIter - 1)
+          {
+            int size = n - (int)pow(2, iter);
+            if(size > hyperCubeIndxCopy.size()){
```
stwhite91, line 29: I don't think you need hyperCubeIndxCopy and hyperCubeStoreCopy at all actually, can't you just use hyperCubeIndx and hyperCubeStore in the call to recvHypercube()?

src/libs/ck-libs/allGather/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+A library for implementing collectives commonly used in machine learning tasks including allGather and allReduce.
+
+## allGather
+
+allGather lets you gather data distributed accross different chare array/group elements. The library provides 3 algorithms for doing the allGather operations, namely ring, hypercube and flooding.
+
+### How to use
+
+The library is available in a default charm++ build and to use allGather, you simply have to include the header `allGather.hh` and link against the library using the flag `-lmoduleallGather`.
```
stwhite91, line 11: allGather.hh -> allGather.h

PR #3887, opened 2025-04-23
[CrayPMI] Missing headers

etiennemlb:
When checking for the appropriate pmi functionalities, we use the module defined variables to get the library path, but we don't do that for the headers.

In the (common) situation where we load cray modules but do not use the wrapper, this breaks. By explicitly adding the include path, we wont get issues.

### DIFF
--- a/cmake/detect-features-c.cmake
+++ b/cmake/detect-features-c.cmake
@@ -290,7 +290,8 @@ int main()
 }
 " CMK_HAS_NUMACTRL)
 
-set(tmp ${CMAKE_REQUIRED_LIBRARIES})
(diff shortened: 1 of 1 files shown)


ritvikrao (approved):

matthiasdiener (changes requested):
Can you try resetting the value of CMAKE_REQUIRED_INCLUDES after this test, similar to how CMAKE_REQUIRED_LIBRARIES is reset after the test?

ericjbohm (approved):

matthiasdiener (approved):

PR #3890, opened 2025-04-29
Revert "Fix Charm4py HAPI"

mayantaylor:
Reverts charmplusplus/charm#3889
Seems to not actually be needed? Charm4py builds and runs examples without error after adding a condition around the hapi call.


ritvikrao (approved):

mayantaylor:
Decided this error checking is still important - users could use a charm build without cuda but turn on the hapi library in charm4py, which should still produce an error.

PR #3891, opened 2025-04-30
Add "send to self" converse benchmark

ZwFink:
Times how long it takes for a PE to send some number of messages of a certain size to itself


ritvikrao (approved):

ritvikrao:
It's important to merge this so that we can compare old Converse and reconverse on this benchmark.

matthiasdiener (approved):

Issue #3893, opened 2025-05-16
HAPI callbacks never trigger


spencerw:
A recent commit to the main branch seems to have broken the HAPI callbacks.

Building charm with the cuda flag enabled and running the `vecadd` example in`  examples/charm++/cuda` causes a hang.

I believe the issue is caused by [this change](https://github.com/charmplusplus/charm/commit/ca862af7e52eafa316aee6be4d06e8c6042fd37f#diff-39832dfff65801e44edf5b7237bdcb183fab3e31371f3fffa2fc727263a260a3R1370-R1380). Moving `queue.pop()` back to its original spot seems to fix the issue for me.

PR #3894, opened 2025-05-16
HAPI polling bugfix

mayantaylor:
Undoing a change from latest release (#3852). The HAPI event should not be popped from the queue unless it is successful. I tested this on Delta with our basic Charm4py program successfully.


mayantaylor:
Addressing issue #3893

ritvikrao (approved):

Issue #3898, opened 2025-07-02
Restart from checkpoint can have _numGroups inconsistent with _groupTable


trquinn:
If an application deletes chare arrays with ckDestroy(), then writes a checkpoint with CkStartCheckpoint(),  the application restarted from the checkpoint will have _numGroups smaller than the last ready entry in _groupTable.  In the restarted application, such entries will be overwritten with subsequent ckNew()s.  One possible outcome is a hang when a message sent to a new group arrives before the group is created and gets delivered to the group object that will be overwritten at group creation.

I believe this is the underlying issue causing issue #3678

PR #3899, opened 2025-07-02
CkPupPerPlaceData(): set numObjects correctly.

trquinn:
In CkPupPerPlaceData(), numObjects will set _numGroups and _numNodeGroups, and these need to be greater than the maximum index of a valid entry in _groupTable and _nodeGroupTable, respectively.

This is a proposed fix for #3898 and #3678.
I will do some testing.

### DIFF
--- a/src/ck-core/ckcheckpoint.C
+++ b/src/ck-core/ckcheckpoint.C
@@ -607,9 +607,9 @@ static void CkPupPerPlaceData(PUP::er &p, GroupIDTable *idTable, GroupTable *obj
   }
 
   if (p.isUnpacking()) {
-    if(CkMyPe()==0)  
-      numObjects = maxGroup+1; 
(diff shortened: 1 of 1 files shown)


trquinn:
I tested this on a run that hung after 5 to 8 restarts.  This fix kept it going for at least 15 restarts.

rbuch (approved):
Looks good to me, but, after looking at the other code a bit, it's unclear to me why the original code wasn't working. Do you have any insight on when `gID.idx` would be greater than `numGroups`?

ericjbohm (approved):
Looks like it should address the problem.

trquinn:
@rbuch If you delete a chare array and then create a new one, the ckNew() does not reuse groupIDs.  In this case the largest groupID.idx will be greater than numGroups.

PR #3903, opened 2025-08-28
improve cmake support for charm on top of reconverse 

JiakunYan:
Changes in this PR include
- Enable cmake autofetch for reconverse: no need to download the reconverse source code.
- Build reconverse as a shared library: no need to specify all its dependencies inside `charmc`.
~~- Comment out `CmiCheckAffinity` for now.~~ (https://github.com/charmplusplus/reconverse/pull/93 fixes it.)
- Add more arguments to the `build` script, including
  - `--with-fetch-reconverse-tag`: git tag/hash to fetch reconverse (default option: `main`)
  - `--wtih-fetch-reconverse-dir`: path to local reconverse source code (default option: none). If specified, charm will not download reconverse from GitHub but use the local copy.
  - `--with-cmake-args`: additional cmake variables to pass.
  - `--without-reconverse-fetch-lci`: if you don't want reconverse to fetch lci automatically (e.g., if you only run a single process).

Build on a local Linux laptop:
```
./build charm++ multicore-linux-x86_64
```
(We still use the multicore build for now.)

Build on Delta.
```
./build charm++ multicore-linux-x86_64 --with-cmake-args="-DLCI_NETWORK_BACKENDS=ofi"
```
(Because Delta weirdly has a non-functional libibverbs installed.)

### DIFF
--- a/buildcmake
+++ b/buildcmake
@@ -131,7 +131,10 @@ opt_shmem=0
 opt_xpmem=0
 opt_qlogic=0
 opt_randomized_msgq=0
-opt_reconverse=0
+opt_reconverse=1 # fake argument for now
+opt_fetch_reconverse_tag=""
+opt_fetch_reconverse_dir=""
+opt_reconverse_fetch_lci=1
 opt_refnum_type="unsigned short"
 opt_replay=0
 opt_shrinkexpand=0
@@ -144,6 +147,7 @@ opt_tcp=0
 opt_tracing= #undef
 opt_tracing_commthread=0
 opt_zlib=1
+declare -a opt_cmake_extra_args=()
 
 # default to not building ROMIO on AMPI due to GCC 14 cascade failures
 case "$actual_triplet" in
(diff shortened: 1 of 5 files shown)


buildcmake
```diff
@@ -131,7 +131,10 @@ opt_shmem=0
 opt_xpmem=0
 opt_qlogic=0
 opt_randomized_msgq=0
-opt_reconverse=0
+opt_reconverse=1 # fake argument for now
```
matthiasdiener, line 134: Does it make sense to add this as an actual option (like `--reconverse` or so)

cmake/fetch_reconverse/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# This directory is solely for fetching reconverse.
```
matthiasdiener, line 1: Out of curiosity, does this file need its own directory, or could it be (e.g.) `cmake/fetch_reconverse.cmake`?

buildcmake
JiakunYan: I think we already have the `--reconverse` option. Since this branch only works for the reconverse build now, I just set it to 1 by default to simplify things. In the long run, I would recommend adding a new NETWORK type called `reconverse` and making it an alias of `multicore` plus enabling this option.

JiakunYan:
> > Would it make sense to add a CI test to this PR?
> 
> Good idea. Working on it.

But then I found that this PR does not have the permission to run the github workflows.

ritvikrao:
This looks good to me and works on my mac, so I will merge it

PR #2530, opened 2019-09-25
[lbMerge - 4] Create distinct LBDatabase and LBManager, remove batsyncer, various cleanups

rbuch:

PR #3904, opened 2025-09-05
Compile charm with CMK_LBDB_ON=0

ritvikrao:
This allows Charm++ to compile with CMK_LBDB_ON set to off.

### DIFF
--- a/src/ck-core/ck.C
+++ b/src/ck-core/ck.C
@@ -640,16 +640,17 @@ void CkCallstackPop(Chare *obj) {
   _ckStartTiming();       // resume timing of the previous obj
 }
 
-#if CMK_LBDB_ON
+
 CkLocRec *CkActiveLocRec(void) {
+  #if CMK_LBDB_ON
   auto *obj = CkActiveObj();
   if (obj && obj->ckInitialized) {
     return obj->getCkLocRec();
   } else {
     return nullptr;
   }
+  #endif
 }
-#endif
 
 /******************** Basic support *****************/
 void CkDeliverMessageFree(int epIdx,void *msg,void *obj)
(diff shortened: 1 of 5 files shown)


lvkale:
It will be worthwhile to run some fine grained benchmarks with and without this being set to 1. (make sure you are timing inside the program.. not the startup.) . For pingpong with chare arrays, use a long number of iterations. You also may need a program with UsesAtSync (?) and isMigratable turned on, which is nevertheless fine grained. Of course without any actual calls to AtSync().

lvkale (commented):
I will approve it soon, since we need it merged but: 
(a) we should write the performance test I indicated. Although admittedly is tedious to run the test since w need two different bulds of charm (without and without CMK_LBDB_ON). 
(b) is this meant to shut of only object load measurement or object migration itself? If only the former, we still need location record (right?) . This is when (for exmple) you may use migrteMe() calls, or balance based on estimated loads, or migrate for other objectives than load balancing (such as shrink expand or fault tolerance).

ritvikrao:
> I will approve it soon, since we need it merged but: (a) we should write the performance test I indicated. Although admittedly is tedious to run the test since w need two different bulds of charm (without and without CMK_LBDB_ON). (b) is this meant to shut of only object load measurement or object migration itself? If only the former, we still need location record (right?) . This is when (for exmple) you may use migrteMe() calls, or balance based on estimated loads, or migrate for other objectives than load balancing (such as shrink expand or fault tolerance).

a ) I will work on it. But what's interesting is that you can just turn the instrumentation off with `+LBOff` at runtime. The CMK_LBDB_ON option seems to be a legacy of when LBDatabase was also responsible for load balancing management. But PR #2530 from 2020 created a separate LBManager that is always on, with object statistics able to be turned off at runtime (and then optionally turned on manually in the code with LBTurnInstrumentOn().
b) that's a good catch, I changed the build to allow the location record to be retrieved at all times. CMK_LBDB_ON just enables/disables the load balancing database (the stats collection and timing)

PR #3906, opened 2025-09-12
add the pingack benchmark to charm++

ritvikrao:
Adds ping-ack (with timing) as a converse benchmark


ericjbohm (approved):

ritvikrao:
> Please add a short describing what the benchmark does.
> 
> Also.. is any purpose served by having the filename be ping rather than pingack?

I changed the file name and added a readme

ericjbohm (approved):

lvkale (commented):
The readme text needs to be elaborated. (Also, you mean e.g. instead of "i.e." for the result data format. )
I suggest adding: Other than control messages all message sends happen from one process and all receives happen on the other process. This asymmetry is intentional so send related overheads can be separated easily from receive related overheads. This benchmark also exposes any bottlenecks in the message path, such as communication threads, serializing locks, etc. It has also been used to see how wide a processes should be in finegrained apllications (i.e. how many pes per processes.. or alternatively how many processes should one have on one physical node, for good performance).

ritvikrao:
> The readme text needs to be elaborated. (Also, you mean e.g. instead of "i.e." for the result data format. ) I suggest adding: Other than control messages all message sends happen from one process and all receives happen on the other process. This asymmetry is intentional so send related overheads can be separated easily from receive related overheads. This benchmark also exposes any bottlenecks in the message path, such as communication threads, serializing locks, etc. It has also been used to see how wide a processes should be in finegrained apllications (i.e. how many pes per processes.. or alternatively how many processes should one have on one physical node, for good performance).

I changed the readme to add this information

lvkale (approved):

PR #3912, opened 2025-10-17
fix windows test

ritvikrao:
This bumps the VS Enterprise version to the latest one used in Github Actions.


ericjbohm (approved):
Necessary hardcode due to vswhere being MIA.

lvkale (approved):

PR #3916, opened 2025-12-24
Fix for redundant thread contention in CsdNextMessage

vladimir-polin:
Fix for redundant thread contention in CsdNextMessage (CmiTryLock/pthread_mutex_trylock)  for node-level threading parallelism with high number of threads.
Instead of simple try_lock call this patch checks lock-free first that the queue is not empty and if it is not empty then there is try_lock and the queue is checked again inside the critical section to avoid a potential data race.

### DIFF
--- a/src/conv-core/convcore.C
+++ b/src/conv-core/convcore.C
@@ -1766,14 +1766,18 @@ void *CsdNextMessage(CsdSchedulerState_t *s) {
 	/*#warning "CsdNextMessage: CMK_NODE_QUEUE_AVAILABLE" */
 	if (NULL!=(msg=CmiGetNonLocalNodeQ())) return msg;
 #if !CMK_NO_MSG_PRIOS
-	if(CmiTryLock(s->nodeLock) == 0) {
-	  if (!CqsEmpty(s->nodeQ)
-	   && CqsPrioGT(CqsGetPriority(s->schedQ),
-		         CqsGetPriority(s->nodeQ))) {
-	    CqsDequeue(s->nodeQ,(void **)&msg);
+	/* Check the node queue lock-free if we have anything to work with. */
+	/* This significantly reduces redundant thread contention in CmiTryLock(). */
+	if (!CqsEmpty(s->nodeQ)) {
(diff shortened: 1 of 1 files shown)


trquinn:
I had also noticed a lot of time spend in locking the Node Queue.  I implemented a similar fix:
```diff --git a/src/conv-core/convcore.C b/src/conv-core/convcore.C
index 9581f234f..718bb9cc0 100644
--- a/src/conv-core/convcore.C
+++ b/src/conv-core/convcore.C
@@ -1766,7 +1766,7 @@ void *CsdNextMessage(CsdSchedulerState_t *s) {
        /*#warning "CsdNextMessage: CMK_NODE_QUEUE_AVAILABLE" */
        if (NULL!=(msg=CmiGetNonLocalNodeQ())) return msg;
 #if !CMK_NO_MSG_PRIOS
-       if(CmiTryLock(s->nodeLock) == 0) {
+       if(!CqsEmpty(s->nodeQ) && (CmiTryLock(s->nodeLock) == 0)) {
          if (!CqsEmpty(s->nodeQ)
           && CqsPrioGT(CqsGetPriority(s->schedQ),
                         CqsGetPriority(s->nodeQ))) {
```
but I never got around to benchmarking the change.  Neverless I think this is a good idea for machines with a high core count.

vladimir-polin:
> I had also noticed a lot of time spend in locking the Node Queue. I implemented a similar fix:
> 

yeah, I had also started with one-line fix but then decided to add one more basic block to make the commentary more understandable.

ericjbohm (approved):

ritvikrao (approved):

vladimir-polin:
Thank you for approving changes! But I do not see "merge" button, is this because of CI stall on darwin? Or it will be manually merged by reviewers?

ericjbohm:
We're working through a fix to the CI issue with darwin.  When we have that
in place, I'll rebase your change to get it merged.

vladimir-polin:
So nothing is needed from my side yet, thanks for letting me know.

PR #3919, opened 2026-03-21
changes to CMK_GLOBAL_LOCATION_UPDATE so that it uses new loc-cache API

anant37289:
@lvkale changes for global location update(tested with MetisLB on examples/charm++/load_balancing/stencil3d)

### DIFF
--- a/doc/charm++/manual.rst
+++ b/doc/charm++/manual.rst
@@ -3304,6 +3304,24 @@ charm/lib, named as *libcldb-foo.a*, where *foo* is the strategy name
 used above. Now one can use **-balance foo** as compile time option to
 **charmc** to link with the *foo* seed load balancer.
 
+.. _globalupdate
+
+Global Location Update
+~~~~~~~~~~~~~~~~~~~~~~
+
+When Charm++ is built with the ``CMK_GLOBAL_LOCATION_UPDATE`` option(``-DCMK_GLOBAL_LOCATION_UPDATE=1``), the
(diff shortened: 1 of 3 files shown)


lvkale (changes requested):
Looks good. See 2 action items below (readthedocs and assert)

I am wondering what happens to anytime migration. (i.e. if individual chares are migrated in-between load balancing steps.. say using migrateme). Either we should disallow it (in which case, we need to add an assert somewhere, maybe in the hopcount code) or permit it (it will just increment epoch for that one chare, but inconsistently.. in only the aquaintences that get the updateLocation because of hopcount). 
I would rather go withthe former for now. And if we are doing that, the epoch counts are pointless. They only change at load balancing step, where everyones count goes up by one, and there is never a hopcount exception leading to updateLocation message. We could rest all hopcounts to 0 at load balancing. But its harmless to leave them as they are, I think. 

So, for now, I suggest we update (or add) documentation in readthedocs in the load balancing section (at the end?)  saying "when you use the build option (there is a build option, right) for global update, you cannot use anytime migration such as via MigrateMe() --- all migrations must happen at global load balancing, and you must use the AtSync() method for triggering load balancing from the application. "

Also add a CkAssert(..) protected by the flag "CMK_GLOBAL_LOCATION_UPDATE" in the code that sends an individual updateLocation message back based on hopcount. This should never happen if that flag is on, unless there is a bug.

lvkale (approved):

ritvikrao (approved):

PR #3924, opened 2026-05-11
Ericjbohm/update hwloc 2.13.0

ericjbohm:
Update HWLOC to 2.13.0 in order to support Blackwell and other new architectures


ericjbohm:
Trying to figure out how we're hitting this "configure: WARNING: hwloc requires a C99 compiler" bug on compilers that should exceed that requirement.

evan-charmworks:
@ericjbohm Double-check the instructions that update-hwloc.sh prints. You'll need to cherry-pick commit 6802ac46b829427fcc6f6f2a8feeebff9c2caa3e to restore the bypass for the C99 configure check that we maintain in our vendored copy of hwloc.

PR #3928, opened 2026-07-06
perf: old dequeue is scheme faster for ucx-arm

ericjbohm:
The dequeue optimization in #3916 creates a
performance regression (loss of 1ns/day) for NAMD on Vista at large scale. This adds CMK_NODE_QUEUE_LOCK_BEFORE_CHECK to switch back to the prior scheme and enables it in the
ucx-linux-arm8 conv-mach.h

### DIFF
--- a/src/arch/ucx-linux-arm8/conv-mach.h
+++ b/src/arch/ucx-linux-arm8/conv-mach.h
@@ -82,5 +82,6 @@ In fault tolerant architectures, CK_MEM_CHECKPOINT can be set. In this case the
 extended header must contain also another field called "pn" (phase number).
 
 */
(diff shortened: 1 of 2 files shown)


ritvikrao (approved):

lvkale:
1ns/day out of what? (as in what is the percent increase in time). 
Also.. it will be nice to create an issue for the performance loss that Tom mentioned "a lot of time" sounds bad. We probably should have a small benchmark to illustrate it.

ericjbohm:
STMV_8M runs at better than 13ns/day with this enabled, vs about
11.5ns/day.   So it is around 10%.

lvkale (approved):
Seems straightforward.

PR #3930, opened 2026-07-13
reconverse-darwin-arm8: compile with C++17 - Apple clang default C++98

lvkale:
Apple clang still defaults to gnu++98; reconverse headers require C++17. Append -std=gnu++17 to CMK_CXX_FLAGS and also to CMK_NATIVE_CXX_FLAGS / CMK_SEQ_CXX_FLAGS, because conv-mach-darwin.sh snapshots those from CMK_CXX_FLAGS before this arch script runs (charmxi builds with NATIVE).

Problem

A fresh ./build charm++ reconverse-darwin-arm8 ... on macOS fails compiling charmxi with ~20 errors per file (alignas rejected, constexpr unknown, >> in template args requiring > >). Reconverse headers (included via xi-util.h → converse.h) require C++17, but Apple clang (tested: 16.0.0) still defaults to C++98 when no -std= flag is given — verified with clang++ -x c++ -dM -E /dev/null | grep __cplusplus → 199711L. The Linux arch files don't hit this because modern g++ defaults to gnu++17.

Fix (7 lines, mac arch file only)

Append -std=gnu++17 in src/arch/reconverse-darwin-arm8/conv-mach.sh to three flag sets:

- CMK_CXX_FLAGS — the general C++ flags;
- CMK_NATIVE_CXX_FLAGS and CMK_SEQ_CXX_FLAGS — these are snapshotted from CMK_CXX_FLAGS inside conv-mach-darwin.sh (line 31), i.e. before the arch-specific script gets to append anything, and charmxi builds with the NATIVE set (charmc:1324–1329). Appending only to CMK_CXX_FLAGS fixes nothing for the tool build — this ordering subtlety is documented in the comment.

Testing

On macOS 15 / Apple clang 16 / arm64:
- ./build charm++ reconverse-darwin-arm8 --with-fetch-reconverse-dir=... --with-production -j8 completes ("Build successful", 100%).
- ckhello runs correctly under lcrun (2 procs × 2 PEs).
- Converse pingpong and a chare-based fib app run correctly (1×8 and 2×4 configs).

No effect on any non-darwin arch.

🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/src/arch/common/cc-gcc.sh
+++ b/src/arch/common/cc-gcc.sh
@@ -43,7 +43,7 @@ if [ "$CMK_MACOSX" ]; then
 
   # keep in sync with mpi-darwin-x86_64/conv-mach.sh
   CMK_CC_FLAGS="-fPIC"
-  CMK_CXX_FLAGS="-fPIC -Wno-deprecated"
+  CMK_CXX_FLAGS="-fPIC -Wno-deprecated -std=gnu++17"
   CMK_LD_FLAGS="-fPIC"
   CMK_LDXX_FLAGS="-fPIC -multiply_defined suppress"
 fi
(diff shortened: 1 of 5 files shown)


lvkale:
an alternative fix is putting -std=gnu++17 in cc-clang.sh or conv-mach-darwin.sh so all darwin targets get it . @ritvikrao which is better?

ritvikrao:
> an alternative fix is putting -std=gnu++17 in cc-clang.sh or conv-mach-darwin.sh so all darwin targets get it . @ritvikrao which is better?

I think this makes sense, I think C++17 should work for netlrts and multicore builds too (the other darwin targets we need)

lvkale:
Can you make that change?

PR #3932, opened 2026-07-24
ci: skip the build/test matrix for documentation-only pull requests

lvkale:
A pull request touching only doc/ currently runs the entire CI matrix (17 build-and-test jobs, up to 90 minutes each). That wastes compute and, more importantly, discourages documentation changes — a small manual fix should not carry the cost of a full acceptance run.

Mechanism:
- A new first job (`changes`) diffs the PR against its merge base and outputs whether anything outside doc/ changed.
- Every build/test job gains `needs: changes` and an `if:` on that output. For documentation-only PRs they conclude as **skipped**, which satisfies required status checks (unlike workflow-level `paths-ignore`, which leaves required checks unreported and blocks merging).
- The `doc` job (sphinx build) always runs — it is the check that actually validates a documentation change.
- `merge_group` events always report `non_doc=true`, so the merge queue still runs the full matrix as the final gate before anything lands. Code PRs are unaffected apart from the one small detection job.

Example of the class of PR this helps: #3931 (a 47-line manual change currently running the full matrix).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,37 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Detect whether the pull request changes anything outside doc/.
+  # Documentation-only PRs skip the build/test matrix below: each gated job
+  # reports "skipped", which satisfies required status checks, and only the
+  # doc job does real work. merge_group events (the merge queue) always run
+  # the full matrix, so the final gate before landing is unchanged.
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      non_doc: ${{ steps.filter.outputs.non_doc }}
+    steps:
+    - uses: actions/checkout@v4
+      if: ${{ github.event_name == 'pull_request' }}
+      with:
+        fetch-depth: 0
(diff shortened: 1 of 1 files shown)


ericjbohm (approved):

evan-charmworks:
The YAML syntax for GitHub Actions has mechanisms to specify which jobs should run depending on which files are changed and which are not. Using these would be preferable to setting up a multi-job chain to determine the same information. I've linked some examples below. You can see the `'on':` key right at the top of each file.

- https://github.com/NVIDIA/NVTX/blob/dev/.github/workflows/core.yml
- https://github.com/NVIDIA/NVTX/blob/dev/.github/workflows/python.yml
- https://github.com/NVIDIA/NVTX/blob/dev/.github/workflows/rust.yml
- https://github.com/NVIDIA/NVTX/blob/dev/.github/workflows/validate.yml

lvkale:
Noted, Evan. It is somehow reassuring to see all the tests Green, and also our CI has become faster. But you're basic point is valued and we will discuss it in a group/converse meeting sometime.

adityapb (approved):

PR #3955, opened 2026-08-31
9.2: device-to-device communication with reconverse

### LINKED ISSUE #3954: CUDA examples: no -arch flag and no launch-error checks — real kernels silently no-op on newer GPUs
Found during stage-9.1 GPU validation (PR #3951 evidence trail, Anvil A100).

The CUDA example Makefiles pass no `-arch` to nvcc, so objects carry only nvcc's default fatbin (currently sm_75 cubin+ptx — verified with cuobjdump on `helloCUDA.o`). On an sm_80 A100, a kernel from such an object can end up never executing, **with no launch error reported**: observed directly — a fill kernel left its buffers untouched (32768/32768 elements wrong), and adding `-arch=sm_80` alone took it to 0 wrong. `examples/charm++/cuda/hello` cannot detect this because its kernel is empty and it verifies nothing.

Two-part fix, pattern already in-tree in `tests/charm++/cuda/hapitest` (from #3951):
1. `CUDA_ARCH ?= sm_80` (overridable) and `-arch=$(CUDA_ARCH)` in each CUDA example Makefile;
2. a `cudaGetLastError()` check after kernel launches in examples that claim to demonstrate real GPU work — and ideally make the examples verify their outputs, since a silent no-op kernel otherwise looks like success.

ritvikrao:
Implements GPUDirect device-to-device communication with Reconverse. Tested on Frontier and on NVIDIA GPUS (Delta). This pr also includes changes to address issue #3954

### DIFF
--- a/.github/workflows/reconverse-cuda.yaml
+++ b/.github/workflows/reconverse-cuda.yaml
@@ -26,6 +26,10 @@ on:
       - 'src/arch/common/special.sh'
       - 'src/ck-core/ckrdmadevice.*'
       - 'src/conv-core/conv-rdmadevice.*'
+      # ckrdma.C routes device (D2D) completions, and xi-Parameter.C emits the
+      # sender-side device call, so both can break the device path on their own.
+      - 'src/ck-core/ckrdma.*'
+      - 'src/xlat-i/xi-Parameter.C'
       - 'CMakeLists.txt'
       - 'cmake/**'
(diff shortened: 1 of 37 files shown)


lvkale (approved):
Looks good now. 3954 resolved in this branch,  and 3957 created for future to eliminate a performance issue (blocking call). Approving.

PR #3961, opened 2026-09-02
Device zerocopy: release registrations on completion, register-once API, piggybacked acks (#3960)

### LINKED ISSUE #3960: Device zerocopy (`nocopydevice` / `CkDeviceBuffer`) registers the GPU buffer per message on both ends and never deregisters; long GPU-direct runs die with ENOSPC
Branch: `reviewed-with-reconverse` (observed at 9b7a8c2d1, jacobi-overdecomp-bench worktree; the code is unchanged on the base branch), reconverse 0673ee8, LCI OFI backend, Frontier (8x MI250X per node, Slingshot/CXI, ROCm 6.2.4).

## Severity: any sufficiently long run dies

An 8-process, one-node Jacobi (`benchmarks/charm++/cuda/gpudirect/jacobi-overdecomp`, `-z`) with ghost buffers allocated once and reused every iteration dies on every rank, 4 to 7 seconds into the run, at every decomposition tried:

```
------- Processor 1 Exiting: Called CmiAbort ------
Reason: Unhandled C++ exception in user code: 1:313641:0:.../_deps/lci-src/src/network/ofi/backend_ofi.cpp:register_memory_impl:394<lci:Assert failed: false> err : No space left on device
```

That is `ENOSPC` from `fi_mr_regattr`: the NIC's memory registration table is full.

| block | chares/GPU | `-i` | outcome |
|---|---|---|---|
| 512 | 32 | 1000 | pass (4.27 ms/iter) |
| 512 | 32 | 4000 | exit 134, ~5 s in |
| 1024 | 8 | 12000 | exit 134, ~4 s in |
| 2048 | 2 | 20000 | exit 134, ~6 s in |
| 2048 | 2 | 60000 | exit 134, ~4 s in |

Runs of 100 iterations, which is what the existing benchmarks and papers use, never get near the limit, so this has been invisible.

## Cause (from source)

Two independent things combine.

**1. The same-node IPC path is opt-in and silently falls back to the network.** `findTransferModeDevice` (ckrdmadevice.C:532) classifies by topology alone: same process → MEMCPY, same physical node → IPC, else RDMA. But all four branch sites (ckrdmadevice.C:272, 438, 832, 958) test `mode == IPC && csv_gpu_manager.use_shm`, and `use_shm` is set only by the `+gpushm` command-line flag (hapi_impl.cpp:537-549). Without the flag the "inter-node" branch runs for same-node transfers: an RDMA get over the NIC. Any mode counter reports the intended mode, not the branch executed.

**2. On the RDMA-get branch, charm registers the device buffer per message on both sides and never releases either registration.**

- Sender (ckrdmadevice.C:1010): `buffers[i]->lci_ncpy_buffer = CmiNcpyBuffer(buffers[i]->ptr, buffers[i]->cnt);` per buffer, per send.
- Receiver (ckrdmadevice.C:867-868): `CmiNcpyBuffer lci_dest_ncpy_buffer(arrPtrs[i], arrSizes[i], &save_op); lci_dest_ncpy_buffer.rdmaGet(source.lci_ncpy_buffer, ...)`.
- The `CmiNcpyBuffer` constructor registers immediately (reconverse `include/conv-rdma.h:235` → `registerMem` → `CmiSetRdmaBufferInfo` → `conv-rdma.cpp:589 comm_backend::registerMemory` → `lci::register_memory_x` per LCI device → `fi_mr_regattr` with `FI_HMEM_ROCR`).
- Nothing deregisters. `CkRdmaDirectAckHandler` (ckrdma.C:312) hands device operations to `CkRdmaDeviceRecvHandler` and returns before the host-path completion handlers that call `deregisterBuffer`. `CkRdmaDeviceRecvHandler` (ckrdmadevice.C:575-627) does QD bookkeeping, the source callback, the counter and the enqueue, and contains no `CmiDeregisterMem` and sends no dereg message to the sender (contrast the host path's `remoteDeregHandler`, conv-rdma.cpp:48). The sender's `CmiNcpyBuffer` lives in the caller's temporary `CkDeviceBuffer`, and `CmiNcpyBuffer` has no releasing destructor.

Net: two NIC registrations per process-crossing message, for the life of the process.

This is not LCI's own per-message registration (rendezvous.hpp:129/212, communicate.cpp:382), which LCI releases in `~internal_context_t`. It is charm's explicit registration, one layer up.

## What is measured and what is not

- The call chain above is established by reading the source; the line numbers are from the branch named at the top.
- The per-rank registration budget is estimated from wall timestamps only: ~1000-1250 iterations at 32 chares/GPU, ~2900 at 8, ~12000-20000 at 2, which with two registrations per crossing message is of order 1e5-2e5 registrations per rank at every shape. That is consistent with a fixed table budget, but it is a ±30% estimate and wants a counter, not a claim.
- `LCI_USE_REG_CACHE` was OFF in this build (LCI's default). Turning it ON hides the crash, because the cache hands back the existing entry instead of registering again, and it removes the per-message cost for reused buffers. It does not fix this issue: the charm-side refcount still never drops. Please do not close this by flipping the build option.

## Two fixes of different size

**(a) The leak.** Deregister the destination in `CkRdmaDeviceRecvHandler`, and send the sender a dereg message following the host path's `remoteDeregHandler` pattern (conv-rdma.cpp:48). No API change. This alone stops the crash.

**(b) The optimization.** The buffers are the same addresses every iteration (ghost buffers allocated once per chare). Register a reused buffer once and keep the handle for the buffer's lifetime. This needs a device-side equivalent of `CkNcpyBuffer`'s REG/NODEREG lifetime model, which does not exist for `CkDeviceBuffer`, plus release on buffer free and on chare migration. That is a design change and could be a separate issue.

Also worth a look, separately: the `+gpushm` fallback in (1) is silent. A same-node transfer that goes over the NIC because a flag is missing should at least print once.

## Reproduce

```
export PMI_MAX_KVS_ENTRIES=1000
srun -N1 -n8 -c7 --cpu-bind=cores --network=single_node_vni --gpus-per-node=8 \
  ./jacobi_od -W 8192 -H 8192 -w 512 -h 512 -z -i 4000 +pe 8
```

Dies at roughly iteration 1000-1250. `-i 1000` passes.

lvkale:
Fixes #3960.

## What

Two commits, both in `src/ck-core/ckrdmadevice.C` / `.h` (plus two handler registrations in `init.C`), reconverse half only:

1. **Release device registrations on completion (fix a).** Both ends of every inter-node device get were registered per message and never released: the completion returns from `CkRdmaDirectAckHandler` before the host path's deregistration, and `CkRdmaDeviceRecvHandler` had none of its own. With LCI's registration cache off that exhausts the NIC's registration table (`fi_mr_regattr` ENOSPC, every rank, a few seconds in); with the cache on it pins every region at a refcount that never returns to zero. Now the receiver deregisters its destination in place when the get completes and tells the sender to release its source. PE 0 prints once that releases are in force, so a run states its own configuration.
2. **Piggybacked acks.** The sender mints an 8-byte id for every buffer that needs a release and/or a source callback and keeps the work in a per-process table; the receiver returns ids inside the first `CkDeviceBuffer` of its next device message to that PE (`CkDeviceBuffer::acks`, capacity 8, count-prefixed so the other buffers of a message cost one byte); ids that find no such message go out in a standalone `DeviceAckMsg` when the block fills or at idle. The sender resolves an id locally. The converse header and every non-device message are untouched; charmxi is unchanged.

There is deliberately no third piece: no application call, and no table in the runtime, that holds a device registration across messages. An earlier revision of this branch added one — `CkDeviceBufferRegister(ptr, cnt)` / `CkDeviceBufferDeregister(ptr, cnt)` over a per-process interval map looked up by range containment — and the commit that removes it is in the history. Keeping a registration alive for a buffer whose lifetime the runtime does not own is the network layer's registration cache doing its job (charmplusplus/reconverse#214); a second cache here only gives the two a chance to disagree about a range. Per-message registration with release on completion is now the only behaviour for an ordinary device buffer, with no API and no environment variable that suppresses the release.

## How this relates to what exists

- **It is the host path's design.** `handleEntryMethodApiCompletion` sends one reverse message per completion, and `remoteDeregHandler` on the source PE deregisters first and then invokes the source callback, gated on `srcDeregMode`. Commit 2 keeps that shape and replaces the message with an id riding on traffic already going the other way. The device path was missing the state, not the idea: it had no registration-mode concept, so it now reuses `CMK_BUFFER_DEREG` / `NODEREG` / `isSrcRegistered` rather than inventing parallel ones, and the two paths read side by side.
- **The source callback is not duplicated; it composes.** `CkDeviceBuffer` already accepts a source callback ("the data is no longer needed, safe to overwrite"); the release ack answers a different question ("the registration can go"). Same event, two actions: one id per buffer needing either, one reverse path, resolved into deregister and/or fire. For a buffer whose registration is held elsewhere — a pool arena, in the follow-up PR — the release drops out and the callback survives. In the unpatched tree `CmiDeregisterMem` appears nowhere in this file and all four callback sites do `cb->send(); delete cb;`, so the callback never touched registration.
- **The host path still sends the whole `NcpyOperationInfo` per deregistration.** The id table and the piggyback block are not device-specific; the same batching should lift to host RDMA later (separate issue).

## Measured

One Frontier node, 8 processes x 1 PE, 8x MI250X, Slingshot, `-z` without `+gpushm` (the RDMA-get path), a 2-D Jacobi stencil with 2-16 KB ghosts, both arms in one allocation, two reps, `avg_iter_us`:

| | 32 chares/GCD | 2 chares/GCD |
|---|---|---|
| shipped (leak) | 4245 / 4249 | 269.6 / 270.2 |
| fix (a), standalone message per buffer | 4354 / 4322 | 299.8 / 299.6 |
| fix (a), piggybacked — **what this PR ships** | **4239 / 4231** | **272.9 / 271.7** |
| source callback requested, standalone | 4510 / 4515 | 322.4 / 322.0 |
| source callback requested, piggybacked | 4391 / 4369 | 279.9 / 280.0 |
| register-once, since removed (see above) | 4123 / 4130 | 260.8 / 260.9 |

- The shipped code dies at about 1,000 iterations; every fixed arm ran 15,000.
- Piggybacking takes the release from ~4 us per remote message to ~0.3 us (within 1% of not releasing). A requested source callback costs 3-4% piggybacked against 6-20% standalone; what remains is the local callback delivery, not the wire.
- The last row is the arm that was dropped. Holding the registration was worth 2-3% over piggybacked release; that is the margin the LCI registration cache should recover without an application call, and it is measured against the cache, not in place of it.
- With LCI's registration cache compiled in (charmplusplus/reconverse#214) the numbers move by 2-3% and the conclusions do not; the cache masks the leak (the leaking arm survives) but does not fix it.

## Not in this PR

A registered device pool (arenas registered lazily on first network use) is a separate PR on top of this one — the pool owns its arenas' lifetime, which is what makes holding those registrations safe without allocator hooks, and is why `NODEREG` survives in this file. Sweeping idle registrations, and a migration test for pending ack ids (ids are per process; the return PE is taken from the message's `srcPe`), are open.

https://claude.ai/code/session_014B3dZrnTsqMkWjs9CXMFs4

### DIFF
--- a/src/ck-core/ckrdmadevice.C
+++ b/src/ck-core/ckrdmadevice.C
@@ -43,7 +43,6 @@
  */
 
 #ifndef _WIN32
-#include <cstdlib>
 #include <pthread.h>
 #endif
 #include "envelope.h"
@@ -1241,26 +1240,7 @@ void CkRdmaDeviceIssueRgets(envelope *env, int numops, void **arrPtrs, int *arrS
       continue;  // no stream callback: the network, not the GPU, completes this
     }
 
-    // Intra-node transfers finish on the stream, so hang the completion off it.
-    // CK_GPU_STREAMORDER prototype (2026-09-01): "stream-ordered delivery".
-    // The copy into the destination buffer is already enqueued on the stream
-    // the receiver posted, so any device work the entry method issues on that
-    // same stream is ordered after it.  With the env var set, and when the
-    // sender asked for no source callback (which would otherwise promise the
-    // source buffer is reusable), complete the operation now instead of
-    // spending an event record + pending polls per message.  The entry method
-    // must not touch the buffer from the host or from another stream without
-    // its own synchronisation -- that is the contract this mode trades on.
-    static int stream_order = -1;
-    if (stream_order < 0) {
-      stream_order = (getenv("CK_GPU_STREAMORDER") != nullptr) ? 1 : 0;
-      if (stream_order && CkMyPe() == 0)
-        CmiPrintf("CkRdmaDevice> stream-ordered delivery for intra-node device receives (CK_GPU_STREAMORDER)\n");
-    }
-    if (stream_order && save_op.src_cb == nullptr) {
-      CkRdmaDeviceRecvHandler(&save_op, nullptr);
-      continue;
-    }
+    // Intra-node transfers finish on the stream, so hang the completion off it
     hapiAddCallback(postStructs[i].hapi_stream, CkCallback(CkRdmaDeviceRecvHandler, &save_op));
   }
 }
--- a/src/ck-core/ckrdmadevice.C
+++ b/src/ck-core/ckrdmadevice.C
@@ -629,7 +629,7 @@ extern "C" {
 #include "buddy_allocator.h"
 
 struct DeviceRegion { uintptr_t end; bool arena; char layerInfo[CMK_NOCOPY_DIRECT_BYTES]; };
-struct DeviceArena { buddy::allocator* alloc; uintptr_t start, end; bool registered; };
+struct DeviceArena { buddy::allocator* alloc; uintptr_t start, end; bool registered; int device; };
 static std::map<uintptr_t, DeviceRegion> device_regions;   // start -> region
 static std::vector<DeviceArena> device_arenas;
 static std::mutex device_reg_mutex;
@@ -687,8 +687,14 @@ static size_t deviceArenaBytes() {
 }
 
 void* CkDeviceMalloc(size_t size) {
+  // Arenas are per device: a process driving several GPUs gets one arena
+  // set per device, and a request is served only from arenas on the
+  // calling PE's current device (the device hapiMalloc would allocate on).
+  int dev = 0;
+  hapiCheck(hapiGetDevice(&dev));
   std::lock_guard<std::mutex> g(device_reg_mutex);
   for (auto& ar : device_arenas) {
+    if (ar.device != dev) continue;
     void* q = ar.alloc->malloc(size, true);
     if (q) return q;
   }
@@ -697,10 +703,11 @@ void* CkDeviceMalloc(size_t size) {
   DeviceArena ar;
   ar.alloc = new buddy::allocator(bytes, bytes);
   ar.start = (uintptr_t)ar.alloc->base_ptr; ar.end = ar.start + bytes; ar.registered = false;
+  ar.device = dev;
   device_arenas.push_back(ar);
   if (CkMyPe() == 0)
-    CmiPrintf("CkRdmaDevice> device pool: arena %zu of %zu MB at %p (CK_GPU_ARENA_MB)\n",
-              device_arenas.size(), bytes >> 20, (void*)ar.start);
+    CmiPrintf("CkRdmaDevice> device pool: arena %zu of %zu MB at %p on device %d (CK_GPU_ARENA_MB)\n",
+              device_arenas.size(), bytes >> 20, (void*)ar.start, dev);
   void* q = ar.alloc->malloc(size, true);
   if (!q) CkAbort("CkDeviceMalloc: request larger than a fresh arena");
   return q;
--- a/src/ck-core/ckrdmadevice.C
+++ b/src/ck-core/ckrdmadevice.C
@@ -43,6 +43,7 @@
  */
 
 #ifndef _WIN32
+#include <cstdlib>
 #include <pthread.h>
 #endif
 #include "envelope.h"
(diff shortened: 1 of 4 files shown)


src/ck-core/ckrdmadevice.C
```diff
@@ -569,6 +570,304 @@ extern "C" {
...
+
+void CkRdmaDeviceInit() {
+  CcdCallOnConditionKeep(CcdPROCESSOR_STILL_IDLE, (CcdCondFn)deviceFlushAllAcks, NULL);
+}
+
+// The receiver owes 'id' to 'pe'.  Held for the next device message to that
+// PE; sent now if the block is full or piggybacking is disabled.
+static void deviceQueueAck(int pe, uint64_t id) {
+  if (device_ack_nopiggy < 0) {
+    device_ack_nopiggy = (getenv("CK_GPU_ACK_NOPIGGY") != nullptr) ? 1 : 0;
```
adityapb, line 852: Can we cache this env variable value at startup to avoid making this call several times

adityapb: Nevermind, I guess it is cached

adityapb (approved):

ritvikrao (approved):

Issue #222, opened 2013-06-04
Port miniFE to Charm++


PhilMiller:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/222*

---
No body.

PhilMiller:
*Original date: 2017-08-14 14:59:31*

---
This is now done, after having sat idle for a long time.

charmgit:benchmarks/mantevo/miniFE-2.0

pplimport:
*Original author: Justin Szaday*
*Original date: 2017-08-14 15:04:06*

---
Well, I think there was an unfixed issue with reproducibility with this. When running on multicore builds it seems to work perfectly but on netlrts/multinode builds it will occasionally fail (I don't remember how). I meant to look at this but ended up getting caught up with other things.

PR #3968, opened 2026-09-08
GPU-aware load balancer

adityapb:
## GPU-aware load balancing and device-state migration

Lets a GPU application be balanced on the **measured GPU cost of each object**, and lets a chare whose state lives in device memory migrate with that state intact. Three pieces that only make sense together: measure the device work, have a strategy that can act on it, and be able to move an object once the strategy decides to.

### Measurement

CUPTI activity tracing attributes each kernel to the object that launched it. The runtime already brackets every entry method, so applications need no annotation unless they launch outside one (`CUPTI_LAUNCH_WRAPPER`).

Per-kernel records are reduced, while still process-local, to one scalar per object, and only those scalars reach the central balancer. The reduction is a sweep over the device timeline that splits each interval among the kernels holding the device in proportion to the SMs each occupies, so the result is **seconds of whole-device occupancy** rather than wall time. That matters for placement: it estimates what an object would cost on a *different* device, not only what it cost sharing this one.

Measurement follows `LBTurnInstrumentOn()` / `LBTurnInstrumentOff()`, so an application that instruments a window around its own `AtSync` pays the tracing cost only inside it. Tracing is the dominant cost of the feature.

### Strategy

`GreedyRefineCentralGPULB` balances at two granularities at once. Objects are assigned across *GPU groups* — the PEs sharing a device — by GPU load, then across PEs within a group by host load. This matches the usual one-GPU-per-process arrangement: device work decides which GPU an object runs on, host work decides which PE drives it.

### Migration

Device buffers are pup'd with `PUPMode::DEVICE`. On emigration the state is copied into a staging allocation so the element can be destroyed immediately, shipped over the device zerocopy path as a `nocopydevice` entry-method argument, and rebound on arrival. The staging buffer is released by the buffer's own **source callback**, which is the zerocopy layer's existing signal that a source buffer may be reused, rather than by a hand-rolled ack message from the destination.

### Prerequisite, and a restriction worth reading

Requires `CMK_GLOBAL_LOCATION_UPDATE`. A device zerocopy send is addressed to the PE the sender believes hosts the target, so without it a send issued around a migration reaches a PE that no longer hosts the element, and the receive path aborts. The option brings its own restriction: **migrations must happen at load balancing steps**, so a GPU application migrates through `AtSync()`, never `migrateMe()`.

### Tuning: `+LBGreedyRefineTolerance`

Among candidate assignments within the migration budget, one with fewer migrations wins only if its max load is within this factor of the best. It was a hardcoded 1.003, tight enough that it rarely binds — and since PE 0 deliberately runs plain greedy with migrations ignored, the winner is usually that one, which moves nearly every object to gain a fraction of a percent.

Measured on `jacobi2d-imbalance`, 256 chares on 32 PEs, 3 LB steps, one A40 node, varying only this value:

| `+LBGreedyRefineTolerance` | moves | total time | final max/avg |
|---|---|---|---|
| 1.003 (default) | 513 | 0.184 s | 1.152 |
| 1.05 | 275 | 0.130 s | 1.232 |
| 1.20 | 7 | 0.061 s | 1.173 |

The default is left unchanged here: one small workload is not enough to re-pick it, and the knob now makes that cheap to investigate.

### Platform

CUDA only. Every CUPTI-specific declaration, definition and struct field is behind `CMK_CUDA`. On HIP there is no equivalent measurement wired up, so the balancer warns at construction that it will see zero GPU load and balance on host load alone, rather than appearing to balance on a dimension that is empty.

Note `HAPI_CUPTI_LB` is not a second spelling of `CMK_CUDA`: cmake defines it only when it finds the CUPTI library, so a CUDA build without CUPTI has one and not the other, and the inner gate and its stubs exist for that case.

### Testing

`tests/charm++/cuda/gpumigrate` covers device-state migration: two buffers per element, the second deliberately not a multiple of `DEVICE_PUP_ALIGN` so that a packer and unpacker disagreeing about where it starts cannot go undetected; destination buffers poisoned before the device pup so the test cannot pass by reading memory the allocator happened to hand back; driven through `AtSync` with `RotateLB`, since no load-based strategy would move identically loaded objects. CI compiles it, which gates the `PUPMode::DEVICE` surface and the `nocopydevice` entry method; running it needs a GPU.


lvkale (commented):
Read the load-balancer, migration and device-zerocopy parts against `reviewed-with-reconverse`. Nothing here should hold the merge; five comments, in order of how much I would like them addressed before it lands.

**1. Base update.** The branch is still on `310d470e0`, before #3974 bumped `contrib/reconverse` (reconverse #217, #222) and widened `reconverse-ci`'s tier to fourteen directories. GitHub does not rerun checks when the base moves, so the green checks describe the old reconverse and the five-directory tier. Please rebase or merge the base so they run against what this will actually land on; there are no conflicting files.

**2. The `CMK_GLOBAL_LOCATION_UPDATE` dependence, and a suggestion for the follow-up.** As I read `ckrdmadevice.C`, the reason is the shape of the device transfer, not the balancer: the metadata is an ordinary array message and gets forwarded to the element's current PE, but the payload is addressed at send time to `dest_pe` (`CmiSendDevice(dest_pe, ptr, cnt, tag)` under `CMK_GPU_COMM`, the IPC/staged path otherwise, with the explicit abort at `ckrdmadevice.C:258` when `source.dest_pe != CkMyPe()`). A migration between the sender's lookup and delivery strands the payload at the old PE; under `CMK_GPU_COMM` that is an unmatched tagged send rather than an abort, i.e. a hang. The global update narrows the window (and the option's own rule confines migrations to LB steps) but a send in flight during the step still targets the old PE. For the PR that removes the dependence, the host zerocopy entry-method API's design fits: make the payload receiver-driven, so the element's PE, wherever the metadata reaches it, issues a device get from the source PE, which has not moved. Then no location knowledge is needed at the sender at all, and `migrateMe()` comes back for GPU applications. If that is already the plan, ignore this; if not, please consider it before another mechanism grows around the current one.

**3. `conv-mach-cuda.sh`** now links `-lcupti` and defines `HAPI_CUPTI_LB` unconditionally for every legacy (`buildold`) CUDA build, while the cmake path defines `HAPI_CUPTI_LB` only when it finds the CUPTI library. A CUDA installation without `extras/CUPTI` builds with cmake and fails to link with the legacy script. Guarding the shell side the same way (test for `$CUDA_DIR/extras/CUPTI/lib64`) keeps the two builds consistent.

**4. AMD: the migration half could be portable today.** `pupDeviceCopy` in `pup_util.C` is a raw `cudaMemcpy` under `#if CMK_CUDA`, so a HIP build has no device-mode pup at all, although everything around it is already portable (`hapiMalloc` staging, and the device zerocopy transport that ran on MI250X). Routing that copy through HAPI's device-to-device memcpy, and the gpumigrate test's setup copies likewise, makes device-state migration work on AMD independently of measurement. The measurement half stays CUDA-only until someone writes the roctracer/rocprofiler collector, which the HIP warning already says well. The manual's "CUDA only" note would then distinguish the two.

**5. Manual note.** The requirement paragraph should say the dependence is temporary and why (comment 2), so users do not design the AtSync-only restriction into applications as permanent.

Two things that read as clearly right: the `CentralLB` change not updating the source PE's own cache before it emigrates, which fixes a real first-migration assert in the existing option's path independent of GPUs; and gating the pup surface through a CI-compiled test with poisoned destination buffers and a deliberately misaligned second buffer.

src/ck-ldb/GreedyRefineCentralGPULB.C
```diff
@@ -0,0 +1,830 @@
...
+    if (procs[pe].available && procs[pe].cpuLoad > maxCpuLoad)
+      maxCpuLoad = procs[pe].cpuLoad;
+
+  // --- Main assignment loop ---
+  for (int i = 0; i < (int)pobjs.size(); i++) {
+    const GreedyRefineCentralGPULB::GObj *obj = pobjs[i];
+    double obj_gpu = obj->load;
+    double obj_cpu = obj->cpuLoad;
+
+    // ----- Outer: choose a GPU group by GPU load -----
```
lvkale, line 623: **Objects with no GPU work, under a finite B.** The outer choice picks the group with the least GPU load and nothing else. An object with `obj_gpu == 0` (a host-only chare; NAMD has many) adds nothing to any group, so the lightest group stays the lightest and every such object that moves goes to the *same* group; only the inner step then spreads them over that group's PEs. With the default `B = FLT_MAX` the stay-put rule keeps them where they are, so the default is safe, but the concurrent parameter search uses finite `B` and can produce exactly that pile-up, and `max(gpu, cpu)` as the objective may still select it. Suggest: when `obj_gpu == 0`, choose the group whose busiest PE has the least host load (or the group containing the least-loaded PE), and fall back to the GPU rule otherwise.

src/util/pup_util.C
```diff
@@ -162,11 +173,59 @@ void PUP::fromMem::bytes(void *p,size_t n,size_t itemSize,dataType t)
...
+/*Device-mode PUP::er's.
+ *
+ * The sender's device region layout is never described on the wire: the
+ * receiver reconstructs it by making the same sequence of pup calls, so the
+ * sizer, the packer and the unpacker must agree exactly on where each buffer
+ * starts. DEVICE_PUP_ALIGN is that agreement.
+ *
+ * A walker with no device region (gpuOrigBuf == NULL) drops DEVICE buffers,
+ * matching the base er::bytes default. A chare with device state pupped this
+ * way therefore migrates its device data but does not checkpoint it.
```
lvkale, line 193: **Device state silently dropped by walkers without a device region.** As the comment says, a sizer/packer without `gpuOrigBuf` ignores DEVICE buffers, so a chare that pups device state this way migrates it but any checkpoint (`toDisk`, `toMem` without a device buffer, the in-memory double checkpoint) writes none of it, and restart silently hands the chare uninitialized device memory. That should be loud: either abort in `bytes_device` when the er is `IS_CHECKPOINT` (or any non-migration purpose) and the target has no device region, or copy the device bytes into the host stream for those purposes. At minimum the manual should state that device state is not checkpointed.

lvkale (commented):
Two more points from a second read (with Kale), neither blocking; see also charm #3976 for the load-metric discussion (placement dependence of gpuTime, idle time as capacity).

ritvikrao (approved):

lvkale (approved):

PR #3969, opened 2026-09-09
Reconverse as a submodule, and an RPATH so charmc-built binaries run without LD_LIBRARY_PATH

ritvikrao:
## Summary

Two changes to how reconverse is consumed, both aimed at making a reconverse build behave like an ordinary one.

### 1. Reconverse is a submodule, not a FetchContent pin

Reconverse was fetched at configure time from a SHA hardcoded in `cmake/fetch_reconverse/CMakeLists.txt`. It is now the `contrib/reconverse` submodule, with the gitlink as the pin.

The win is that trying a different reconverse is a checkout and a rebuild, with no CMake edit and no build directory to blow away:

    git -C contrib/reconverse checkout <branch-or-sha>

It also retires a real hazard: the old SHA lived in a `set(... CACHE STRING)`, which seeds on first configure and never updates, so bumping it had no effect on an existing build directory — it kept fetching the old reconverse and failed later as an unrelated-looking compile error. The file carried a guard for exactly that. A submodule needs none: the tree on disk *is* the pin.

- `cmake/fetch_reconverse` → `cmake/reconverse`. It `add_subdirectory`s the submodule into `<build>/reconverse`, so the submodule tree stays clean, and runs `git submodule update --init` itself when the submodule is empty — cloning without `--recurse-submodules` shouldn't be a first-build failure.
- `--with-fetch-reconverse-dir` → `--with-reconverse-dir` (old spelling still accepted); relative paths now resolve against the invoking directory rather than the build directory.
- `--with-fetch-reconverse-tag` is removed; passing it exits with the equivalent submodule commands.
- `reconverse-cuda`'s path filters gain `contrib/reconverse` and `.gitmodules`. A pin bump used to touch `cmake/**`; as a gitlink it would have skipped the CUDA job silently.

### 2. charmc records an RPATH for libreconverse, so binaries run bare

Pre-existing on this line and independent of the submodule work, but it's the other half of "building with reconverse should just work".

`libreconverse` is linked by install_name/SONAME (`@rpath/libreconverse.<suf>`), but charmc contributed only `-L`, which is link-time only. Every charmc-built binary died at startup unless the user set `(DY)LD_LIBRARY_PATH` by hand:

    dyld: Library not loaded: @rpath/libreconverse.dylib
          Reason: no LC_RPATH's found

cmake-built targets were fine because cmake stamps its own build RPATH, so the gap was exactly in the wrapper user programs go through. charmc now records the directory it linked from as an RPATH; one entry also covers `liblci`/`liblct`, which sit alongside it. Classic builds have no libreconverse for the guard to find, so their link lines are untouched.

This is branch commit `bf7329331` from `reconverse-specific-build`. The ledger had marked `src/scripts/charmc` **superseded**, but the reviewed charmc had no RPATH logic for that to be authoritative about — the fix was being dropped, not superseded. The row now says so.

**The env-var workarounds go with it**, in `src/scripts/testrun` and four CI lines. Leaving them would have kept the tier passing with the RPATH gone, which is how the gap survived this long.

### 3. Pin bumped to reconverse main

`2c50813` (what the fetch pin named) → `38091f1`, picking up #213 partitions, #214 LCI registration cache, #215 SPSC self-message queue.

`.gitmodules` also declares `branch = main`. That changes nothing for a clone or `--init`, which still check out the recorded commit; it only lets `git submodule update --remote` follow main in one command. Tracking main automatically on every build is deliberately not done — the recorded commit is what keeps a given Charm++ commit building against a known reconverse.

## Validation

- [x] CI green (reconverse-ci) — all 5 jobs on all three commits, including at the bumped pin. The RPATH run is the meaningful one: the tier executes with no library-path env var anywhere (`testrun(reconverse): ./megatest +pe 1`), covering ELF/Linux as well as local Mach-O.
- Local, macOS 15 / Apple Silicon, `reconverse-darwin-arm8 --with-production --enable-tracing`:
  - Clean build at the new pin; megatest, simplearrayhello, anytime_migration, anytime_bcastred and qd all pass, single-process and under `lcrun -n 2`, run with `env -u DYLD_LIBRARY_PATH -u LD_LIBRARY_PATH`. charmc-built binaries carry `LC_RPATH -> <build>/lib`.
  - Clone *without* `--recurse-submodules` self-initializes; `--with-reconverse-dir=<relative>` works; `--with-fetch-reconverse-tag=` exits 1 with instructions.
  - Classic `netlrts-darwin-arm8` builds with reconverse untouched, gets no LC_RPATH, and still runs.
- **Runtime-core changes only:** n/a — build system and link line only.

## Review notes

- Worth the hardest look: the auto-init in `cmake/reconverse/CMakeLists.txt` writes to the working tree from a configure step. It is gated on the submodule being absent *and* `.git` existing, so it won't stomp a branch you checked out there.
- **Release tarballs**: `git archive` doesn't carry submodules, so a tarball reconverse build now fails with a message pointing at `--with-reconverse-dir` where before it would have fetched. Deliberate — one source of truth for the pin. If tarball builds need to work, the fix belongs in `package-tarball.sh`, not a second pin in CMake.
- The removed `LD_LIBRARY_PATH` prefixes are worth a sanity check on a real fabric; hosted CI only covers loopback/TCP.


adityapb (approved):

lvkale (commented):
Ok, I guess. We may decide in future to just use latest version of reconverse, since we charm++ will remain a test for reconverse (probably should make a ci for that).

Issue #3967, opened 2026-09-03
Device zerocopy: user documentation and diagnostics owed


lvkale:
Collecting the user-facing work owed around the device zerocopy API, so it is assignable rather than remembered. Standing rule (Kale): user-facing documentation changes with the change. #3961 does this for registration and buffer reuse; this is what remains.

## 1. The LCI-mediated path: warn, and fail informatively

Decision (Kale, 2026-09-03): the LCI-mediated path — the application allocates its own buffers, the runtime registers per send, and LCI's registration cache deduplicates — **stays supported**, on condition that users are warned in the documentation and that failures are informative.

It is the right default for the common case: a bounded set of long-lived buffers sent many times, where the first send registers and every later send is a cache lookup. Two properties need stating plainly, because both fail silently today:

- **A freed and reallocated device buffer can be served a stale registration.** Nothing invalidates a cached device region on free: LCI asks for `UCM_EVENT_MEM_TYPE_FREE`, but the vendored UCX subset omits the `ucm/cuda` and `ucm/rocm` modules that would deliver it. Demonstrated by putting host memory in the same condition — freeing via a raw syscall so no UCM event fires — where the cache returned the registration created before the mapping was destroyed.
- **Registrations are never returned while the cache is on.** `max_regions`, `max_size` and `max_unreleased` are all unlimited, so nothing evicts; a release only drops a refcount. An application with a bounded buffer set never notices. One that registers many distinct short-lived buffers accumulates memory regions without bound.

Documentation should therefore say: with application-allocated buffers, do not free and reallocate device buffers that have been sent; if the working set of distinct sent buffers is large or unbounded, use `CkDeviceBufferRegister` or the device pool.

Diagnostics worth adding, since neither condition is currently detectable by a user:

- **A better exhaustion message.** Today the failure is `LCI_Assert(false) ... err : No space left on device` from `register_memory_impl`, which says nothing about what to do. It should name the cause (the NIC's memory-region table is full) and the remedies (register once, use the pool, enable eviction).
- **A registration-pressure warning.** The runtime can count outstanding device registrations and warn once past a threshold, well before the table fills — the difference between a diagnosable warning and an abort thousands of iterations later.

## 2. `hapiGetStream` deprecation

[...]

Issue #3963, opened 2026-09-02
Device zerocopy examples repack send buffers with nothing guaranteeing the neighbour's get has finished reading them


lvkale:
# Device zerocopy examples repack send buffers with nothing guaranteeing the neighbour's get has finished reading them

Every `.C` under `examples/charm++/cuda/gpudirect` and `benchmarks/charm++/cuda/gpudirect` constructs its send buffers as `CkDeviceBuffer(ptr, stream)` and none passes a source callback, although `CkDeviceBuffer` accepts one. `jacobi2d` (and the copies derived from it) repacks the same `d_send_*` buffers every iteration.

On the inter-node path the receiver reads the sender's buffer with an RDMA get issued when the metadata message is processed. The sender proceeds to its next iteration as soon as it has received its own ghosts, which does not depend on whether every neighbour's get of its buffer has completed, so the repack can overwrite a buffer a neighbour is still reading. Under 32 chares per GPU with lagging chares the window is real; the examples check no values, so it shows as nothing.

Two fixes, either sufficient:

- **Double-buffer the send sets**, alternating by iteration parity. A neighbour that has sent its message i+1 has finished its get of buffer i, so no acknowledgement is needed. About ten lines; measured cost nil.
- **Request the source callback** and gate the repack on it. With piggybacked acks (#3961) this costs 3-4% of a chare-iteration on a Jacobi stencil at 2-16 KB ghosts; standalone, 6-20%.

Reproduction and both fixes are in the `jacobi-overdecomp` benchmark on the `jacobi-overdecomp-bench` branch (`-a` requests callbacks with double-buffered sets).

A registered device pool (#3962) makes this more important: it removes the per-message release, so the source callback becomes the only signal a sender gets, and "the pool owns the lifetime" must not be read as "the buffer is safe to overwrite".

PR #3979, opened 2026-09-15
Manual: device zerocopy user documentation and reconverse build lines

lvkale:
# Manual: device zerocopy user documentation and reconverse build lines

Documentation only. One file changed, `doc/charm++/manual.rst`.

Issue #3967 collected the user-facing documentation owed around the device
zerocopy API. This writes it, together with the reconverse build and run
idioms the announcement needs users to have somewhere citable.

## Sections changed

**Installation chapter**

- Three rows in the build-command table (`tab:buildlist`) for
  `reconverse-linux-x86_64`, `reconverse-linux-arm8`, `reconverse-darwin-arm8`.
- New `Reconverse` subsection under *Installation for Specific Builds*
  (label `sec:reconverseinstall`): the three targets; `contrib/reconverse`
  as a submodule and the recorded gitlink as the pin; how to move it; the
  `--with-reconverse-dir=<path>` escape for an out-of-tree reconverse;
  SMP always on; and the run idioms — no `charmrun`, `./pgm +pe N` for one
  process, `lcrun -n 2 ./pgm +pe 8` for several, `lcrun` under
  `<builddir>/_deps/lci-src/`, `+pe` as the total and divisible by the
  process count, and a site launcher (needed even for a single process
  inside a Slurm allocation) with `TESTRUN_LAUNCHER` for the test harness.

**GPU chapter, Enabling GPU Support**

- The `amd` / `hip` build option and the `-amd` build-directory suffix, and
  the `reconverse-linux-x86_64 cuda` / `amd` build lines on which the
  device-to-device path is supported, with a pointer to the new install section.

**GPU chapter, new "Streams: create your own, non-blocking" section**

- User-created streams with `hapiStreamCreateWithPriority(&stream,
  hapiStreamNonBlocking, 0)`, and the note that the portable header exports
  no `...CreateWithFlags`, so this is how flags are passed.
- Why: default-flagged streams implicitly synchronise with the legacy null
  stream and cost 11–13 µs each to create at scale against a flat 1.6 µs.
  `hapiGetStream` hands out default-flagged runtime streams and is therefore
  documented as discouraged for new code.
- The hardware limits a stream count has to respect: 32 hardware queues on
  NVIDIA (`CUDA_DEVICE_MAX_CONNECTIONS`), about 8 on AMD
  (`GPU_MAX_HW_QUEUES`), at most 128 concurrent kernels, and a small fixed
  number of copy engines, so same-direction transfers overlap only up to
  that count.
- One sentence pointing at `GreedyRefineCentralGPULB` (#3968), which the
  load-balancing chapter already documents.

**GPU chapter, Direct GPU Messaging**

- The `CkDeviceBuffer` source callback is no longer described as
  "(optional)". It is the only signal the sender gets that the buffer may be
  written again; the repack race is stated in two sentences, and
  double-buffering by iteration parity is named as the callback-free
  alternative.

**GPU chapter, registration and buffer reuse**

- The two silent properties of the network layer's registration cache from
  #3967 section 1, as a two-item list with remedies: a freed and reallocated
  device buffer can be served a stale registration (allocate once, or use
  the pool), and registrations are never evicted, so an unbounded working
  set of distinct sent buffers fills the NIC's memory-region table (use the
  pool, one registration per arena).

**GPU chapter, device pool**

- Pool scope: `CkDeviceMalloc` is for buffers that will be sent or received
  into, not a general device allocator; an arena is registered whole, so
  scratch taken from the pool is pinned along with the communication buffers
  sharing its arena.

**GPU chapter, `+gpushm`**

- Reframed: opt-in, silent when absent (same-node traffic takes the NIC and
  nothing says so), and measured slower than the RDMA-get path at 2–16 KB,
  so it is something to measure at the application's message sizes rather
  than the setting to turn on. Also records that `+gpucommbuffer`,
  `+gpulbbuffer` and `+gpuipceventpool` are parsed but ignored, with a note
  on PE 0, unless `+gpushm` is given.

**GPU chapter, examples**

- The `examples/charm++/cuda/gpudirect` and
  `benchmarks/charm++/cuda/gpudirect` pointer now says these codes pass no
  source callback and repack each iteration, that they are being corrected
  (#3963), and that they should not be copied as a model for buffer reuse
  until they are.

## What was verified against the source in this tree

- `CkDeviceMalloc` / `CkDeviceFree` exist: `src/ck-core/ckrdmadevice.h:166-167`,
  implemented at `src/ck-core/ckrdmadevice.C:675,702`, arenas per device,
  `CK_GPU_ARENA_MB` for the arena size.
- `+gpushm`, `+gpucommbuffer`, `+gpulbbuffer`, `+gpuipceventpool` and the
  "ignored without `+gpushm`" print: `src/arch/cuda/hybridAPI/hapi_impl.cpp:1101-1143`.
- `hapiGetStream` is implemented at `hapi_impl.cpp:2239` and exported at
  `hapi_functions.h:37`. `hapiCreateStreams` creates its streams with plain
  `hapiStreamCreate` (`gpumanager.h:475`), which is `cudaStreamCreate` /
  `hipStreamCreate` (`hapi_portable.h:79,240`) — default flags. That is the
  basis for calling it discouraged.
- The portable stream spellings that exist: `hapiStreamCreate`,
  `hapiStreamCreateWithPriority`, `hapiStreamDefault`,
  `hapiStreamNonBlocking`, `hapiStreamPerThread` (`hapi_portable.h:79-88`,
  `240-244`). There is no `hapiStreamCreateWithFlags`; only
  `hapiEventCreateWithFlags`. The in-tree examples use
  `hapiStreamCreateWithPriority(&s, hapiStreamDefault, 0)`.
- Reconverse targets: `src/arch/reconverse-linux-x86_64`,
  `reconverse-linux-arm8`, `reconverse-darwin-arm8`.
- Build options and submodule handling: `buildcmake:366-391` for
  `--enable-reconverse`, `--with-reconverse-dir`, the removed
  `--with-fetch-reconverse-tag`, and `amd|hip` at `buildcmake:189-192` with
  `-DBUILD_HIP` at `:716`; `cmake/reconverse/CMakeLists.txt` for the
  submodule being the pin and being auto-initialized; `.gitmodules` for the
  path and URL; `CMakeLists.txt:276-305` for `CMK_SMP` forced on for
  reconverse and `:557-562` for `charmrun` not being built.
- Run idioms and the launcher variable: `src/scripts/testrun:13-66`,
  including `lcrun` at `<builddir>/_deps/lci-src/lcrun` and the reason a
  single-process run inside a Slurm allocation must go through the site
  launcher.
- `GreedyRefineCentralGPULB` is already documented at `manual.rst:2685-2691`
  with `+balancer GreedyRefineCentralGPULB`, so this change only points at it.

## Discrepancies between the issue text and the shipped code

1. **`CkDeviceBufferRegister` / `CkDeviceBufferDeregister` do not exist.**
   #3967 section 1 recommends them as the remedy for both cache hazards.
   They were added and then removed inside #3961: commit `b3e5799d0`
   contains the step "ckrdmadevice: drop the explicit device-buffer
   registration API (#3960)". Nothing in the tree references either name.
   The manual therefore offers allocate-once and the device pool as the
   remedies, and documents no register-once call. The issue text and the
   announcement briefing (which lists "register-once" among the API changes
   GPU users face) both need correcting.
2. **`hapiGetStream` carries no deprecation marker.** No attribute, no
   comment, and it sits outside the `/***** DEPRECATED *****/` blocks that
   `hapi_functions.h` uses for `hapiCreateWorkRequest`, `hapiEnqueue` and
   `hapiWorkRequestSetCallback`. Per the brief, the manual documents
   user-created non-blocking streams as the model and describes
   `hapiGetStream` as discouraged, without claiming a deprecation the code
   does not declare. The one-line code change (a deprecation marker plus the
   AMPI export note) is still owed under #3967 section 2.
3. **The pool cannot be described as invalidating registrations "on free" in
   general.** The existing text says freeing and reallocating *pool* buffers
   is safe because the arena stays registered and the pool owns it; that is
   what the new text says. The issue's framing ("use the pool") holds, but
   for that reason rather than because anything invalidates a region.
4. **`+gpushm` still has no fallback print.** #3967 asks for a one-time
   message when a same-node transfer takes the network; no such print exists
   in `hapi_impl.cpp`, so the manual states the silence as behaviour rather
   than describing a diagnostic. Same for the exhaustion message and the
   registration-pressure warning in section 1 — none are present, so none
   are documented.
5. **Minor:** the pre-existing registration section cites
   `reconverse-linux-x86_64-amd` as an example build directory; that
   spelling is correct per `buildcmake` (the `-amd` suffix), so it was left
   alone.

## Build

`sphinx-build -W -n -b html` over `doc/` succeeds with no warnings
(Sphinx 9.1.0 in a throwaway venv from `doc/requirements.txt`). `-W` makes
any warning an error, so the file is clean, not merely no worse.

`examples/charm++/cuda/gpudirect` and `benchmarks/charm++/cuda/gpudirect`
were not touched; the example corrections for #3963 are a separate change.

Addresses #3967


doc/charm++/manual.rst
```diff
@@ -9432,6 +9502,28 @@ address range is absorbed by the network layer's own registration cache;
 the runtime does not keep a second cache of its own. The one exception is
 memory the runtime itself owns, which is what the device pool below is for.
 
+Two properties of that network-layer cache have to be worked around, and
```
adityapb: I don't know if these details belong in the manual. In any case we should also add a line with our recommendation on when to use CUDA mallocs versus device mempool. Something like -

By default, device buffers allocated by CUDA/HIP malloc will be deregistered after each send is complete. For persistent buffers used for communication, we recommend the use of  `CkDeviceMalloc` to allocate the buffer from a pool of registered memory.

lvkale: Added a recommendation paragraph ahead of the two cache properties, adapted slightly for accuracy: the runtime registers per transfer and releases its registration on completion (that is what #3961 did), while the network layer's cache may retain the region; buffers used for communication throughout a run should come from CkDeviceMalloc, whose arena is registered once and to which the two properties do not apply. Kept the two properties themselves, marked as applying only to non-pool buffers, since testers on the LCI-mediated path hit them silently. Pushed as the latest commit; preview rebuilds in a couple of minutes.

adityapb (approved):

PR #3958, opened 2026-09-01
9.3: wire examples and change CI

ritvikrao:
This completes part 9.3 of the reconverse merge plan. Tested on Frontier, CUDA in progress

PR #3980, opened 2026-09-15
gpudirect/jacobi3d: double-buffer the send faces by iteration parity (#3963)

### LINKED ISSUE #3963: Device zerocopy examples repack send buffers with nothing guaranteeing the neighbour's get has finished reading them

lvkale:
Double-buffer jacobi3d's GPU-direct send and receive faces, and write down the rule.

`CkDeviceBuffer` hands the runtime an address, not a copy. On the inter-host
path the receiver reads that address with an RDMA get issued when it processes
the metadata message, and the sender is told nothing about when that read
finishes — it proceeds as soon as its own ghosts have arrived, which does not
depend on any neighbour's get. An iterative program that repacks the same send
buffer every iteration therefore needs something to order the repack after the
neighbour's read, and `jacobi3d` had nothing.

## What changed

**`examples/charm++/cuda/gpudirect/jacobi3d/jacobi3d.C`** — the fix.

- `d_send_ghosts[DIR_COUNT]` becomes `d_send_ghosts[2][DIR_COUNT]`: two full
  sets of six send faces.
- New `curSendGhosts()` returns the set for the current iteration,
  `d_send_ghosts[use_zerocopy ? (my_iter & 1) : 0]`. This is the only place
  the alternation happens, so the explanation of why parity suffices lives
  there.
- `packGhosts()` packs into `curSendGhosts()`; `sendGhosts()`'s zerocopy branch
  reads the same pointer and builds its six `CkDeviceBuffer`s from it. Both run
  inside one iteration with `my_iter` unchanged (`my_iter` advances in
  `prepNextIter()`, after `update()`), so the pack and the send always agree on
  the set.
- The second set is allocated, ghost-initialized and freed only under
  `-d` (`use_zerocopy`). The persistent path (`-s`) registers its
  `CkDevicePersistent` buffers once at init against fixed addresses, so it
  keeps using set 0 and is otherwise untouched.

- **Receive side, same shape.** `d_recv_ghosts[DIR_COUNT]` becomes
  `d_recv_ghosts[2][DIR_COUNT]`. The post entry method `recvGhostZC` runs when
  a message is delivered to the object (the generated `_call_` wrapper posts
  and issues the rgets before the SDAG buffering method sees the message), so
  a neighbour that has finished iteration *i* can have its *i+1* ghost fetched
  into our receive buffer while our iteration *i* data in that buffer is still
  waiting on a third neighbour. The post entry now selects the set by the
  message's own iteration, `d_recv_ghosts[ref & 1][dir]`, and
  `processGhostZC` unpacks from the pointer the runtime passes back instead of
  from a fixed `d_recv_ghosts[dir]` (the FIXME that was there). Two sets
  suffice by the same argument as the send side: a neighbour cannot send
  iteration *i+2* until it has our *i+1* ghost, which we send only after
  unpacking iteration *i*. The persistent path posts fixed addresses once and
  keeps set 0.
- **Unpack from our own posted buffer, never from the delivered pointer.**
  `processGhostZC` now takes `ref` and unpacks `d_recv_ghosts[ref & 1][dir]`.
  The pointer that arrives with the regular entry method is the sender's
  device address: `CkRdmaDeviceIssueRgets` re-enqueues a copy of the original
  message, whose marshalled `CkDeviceBuffer` still names the source, and the
  generated code hands that `ptr` to the entry method. Inside one process
  that address is valid, so single-process runs never show it; across
  processes the unpack kernel faults, the CUDA context dies, no stream
  callback fires again, and the run hangs with no output (found on Delta:
  every two-process run of the previous revision hung within two iterations
  while the base completed). The base code avoided this by unpacking from a
  fixed buffer of its own, which is what its FIXME was about. This is worth a
  runtime issue of its own: the post API's regular entry method should
  receive the posted pointer.

**`examples/charm++/cuda/gpudirect/README.txt`** — a new "Reusing a send
buffer" section states the invariant, names both mechanisms and which example
demonstrates each, and repeats the pool caveat from the issue: with a
registered device pool the source callback becomes the only signal the sender
gets, so "the pool owns the lifetime" does not mean the buffer is safe to
overwrite. The jacobi3d entry in the example list points at it.

## Why iteration parity is sufficient on its own

No acknowledgement from the receiver is needed. A neighbour issues its get of
our iteration *i* buffer while processing our iteration *i* metadata message,
and the SDAG `when recvGhostZC[my_iter]` continuation that completes its own
iteration *i* only runs once that get has completed. It sends its iteration
*i+1* ghost after that. We cannot begin packing iteration *i+2* — the next use
of that same set — until that *i+1* message has arrived, because the
`for (remote_count = 0; remote_count < neighbors; ...)` loop makes each
iteration wait for every neighbour. So by the time a set is repacked, every
neighbour has provably finished reading it.

The same argument covers teardown: the final iteration's `allDone` reduction
only completes once every chare has received every last-iteration ghost, so no
get is outstanding when the destructors free the buffers.

## Files deliberately not changed

- **`jacobi2d` and `jacobi2d-imbalance`** repack every iteration, but #3958
  already gates the repack on a source callback — the four
  `d_send_*_ghost_done` entry methods their SDAG awaits at the end of each
  iteration. The issue's premise that no example passes a source callback is
  stale for these two. Converting them to parity would also *weaken* them:
  both set `usesAtSync` and migrate at iteration boundaries under `-l`, and the
  callback wait is what orders a neighbour's get before `~Block()` frees the
  send buffers on the source PE. Parity does not order that. Their `pup()`
  comment ("iterate() only reaches AtSync() on an iteration boundary with both
  streams quiesced, so nothing in flight is being dropped here") depends on it.
  Leaving them as they are also means the tree demonstrates both fixes, which
  is what the README section now says.
- **`verify`, `sdag`, `persistent`, `bandwidth`, `latency`,
  `latency-persistent`** send buffers they never repack. `verify` and `sdag`
  fill `d_local_data` once at init and re-send the unchanged contents;
  `bandwidth` never writes its send buffer at all; `latency` memsets it once
  per message size behind a reduction and then ping-pongs one message at a
  time. `persistent` and `latency-persistent` use `CkDevicePersistent` rather
  than `CkDeviceBuffer`, and every step in `persistent` is separated by
  `CkWaitQD()`. None of them has the hazard, so none of them is touched.

## Validation

**Anvil, build: clean.** Clone of this branch at
`$PROJECT/x-lkale/software/gpudirect3963/charm` (submodule
`contrib/reconverse` @ `58921e9`), built with CUDA against reconverse in
5.5 minutes on a login node. The clone was first built at `ea96db27b` and has
since been advanced to `4d1111558`, with the example binaries rebuilt against
it, so the queued jobs below exercise **both** commits — verified by mtime
(`jacobi3d.C` 19:16:24, the binary 19:16:47 Anvil-local) and by the receive
parity line being present in the source that was compiled.

    module load modtree/gpu; module load gcc/11.2.0; module load python/3.9.5
    module load hwloc; module unload cuda; module load cuda/13.1.0
    ./build charm++ reconverse-linux-x86_64 cuda -j8 --with-production \
        --with-cmake-args="-DRECONVERSE_ENABLE_CPU_AFFINITY=ON \
                           -DHWLOC_ROOT_DIR=$RCAC_HWLOC_ROOT"

`jacobi3d`, `jacobi2d`, `verify` and `benchmarks/.../latency` all compile with
exit 0 and no warnings in `jacobi3d`, both at `CUDA_ARCH=sm_80` and (in a
separate staging tree) `sm_90`. The build-tree example directories are
symlinks to the sources, so the compiled `jacobi3d.C` is the edited one.

**Anvil, GPU runs: queued, not yet executed.** Every GPU on the machine was
allocated — 52/52 across the 13 usable GPU nodes, `gpu-debug` at zero jobs
running with 11 pending, `gpu` at ~401 pending — and Slurm's estimated start
for a 1-node 25-minute `gpu-debug` job is **2026-09-18**, backfilling to the
edge of the `anvil-maint-2026-q3` reservation. There is no in-house
alternative: the `ai` partition (H100) rejects this allocation both ways, by a
site submit filter on `-A asc050025-gpu` and by `DenyQos` on `-A asc050025`.

Two jobs are queued and will write their output whether or not the session that
submitted them is alive. Each runs seven steps, every one through
`srun --mpi=pmi2`, with a `RESULT <name> exit=$?` line after it:

| job | shape | partition |
|---|---|---|
| 20751399 | 1 node, 2 A100, 16 CPUs, 25 min | gpu-debug |
| 20751674 | 1 node, 1 A100, 16 CPUs, 25 min | gpu-debug |

The seven steps: `jacobi3d -c 32 -x 512 -i 40 -w 5 -d` at 2 PEs in one process
and again as two processes (32 chares over 2 PEs, so 16x overdecomposed);
`jacobi3d` with no `-d` (host-staged) and with `-s` (persistent) for
comparison, the `-s` run specifically because it shares the buffers this commit
touches; `jacobi2d -W 4096 -H 4096 -w 512 -h 512 -i 40 -u 5 -z` at 2 PEs
(64 chares); `verify` at exactly 2 PEs; and the `latency` benchmark at 2 PEs.

`verify` is the only value-checking program among them — it compares every
received element and aborts on a mismatch. `jacobi3d` checks no values, so for
it a clean exit and a plausible per-iteration time is all the run can show.
Collecting the results needs no rebuild:

    ssh anvil 'squeue -u x-lkale; grep -E "^RESULT |ALL_DONE" \
      /anvil/projects/x-asc050025/x-lkale/software/gpudirect3963/run{1,3}.out'

A resume note with the full recipe, the job table and the per-run expected
output is at `.../gpudirect3963/RESUME.md`. **This PR should not merge before
those two jobs report exit 0 and `verify` reports PASS.**

The reference implementation named in the issue could not be consulted: branch
`jacobi-overdecomp-bench` on origin has no `-a` flag and no double-buffering in
`jacobi_od.C`, so that work appears never to have been pushed. The design here
is derived from the issue text instead.

`git diff --stat` for the commit:

    examples/charm++/cuda/gpudirect/README.txt         | 35 +++++++++++++-
    examples/charm++/cuda/gpudirect/jacobi3d/jacobi3d.C| 56 +++++++++++++++-----
    2 files changed, 78 insertions(+), 13 deletions(-)

No build artifacts are included.

Fixes #3963


lvkale:
## GPU validation on NCSA Delta (A100-SXM4-40GB, CUDA 13.2, reconverse over Slingshot/cxi)

Build: `./build charm++ reconverse-linux-x86_64 cuda -j8 --with-production`, examples at `CUDA_ARCH=sm_80`, launcher `srun --mpi=pmix` (see note below). Partition `gpuA100x4-interactive`.

**Revision `4d1111558` (send + receive double-buffering, unpack from the delivered pointer) — jobs 22106040 / 22106132:** every two-process jacobi3d run hung within two iterations (exit 124 after 300 s, no output after `Init time`), one node and two nodes, full and tiny problem size; the base `b17f4ac18` completed the same runs back to back in the same allocation. Single-process jacobi3d, jacobi2d `-z`, `verify` (chare array / group / nodegroup all PASS) and `latency` passed at both shapes.

Cause: the `nocopydevice` pointer handed to the regular entry method is the sender's device address (`CkRdmaDeviceIssueRgets` re-enqueues a copy of the original message, whose marshalled `CkDeviceBuffer` still names the source). Valid by accident inside one process; a foreign address across processes, so the unpack kernel faults the context and nothing completes. Filed as #3981; Aditya's `rate-aware-gpu-lb` branch already carries the runtime fix (`c8833efb0`).

**Revision `a27143bab` (unpack from the example's own posted buffer) — job 22106353, nodes gpua050 + gpua057:**

```
RESULT run1_jacobi3d_d_1proc exit=0
RESULT run2_jacobi3d_d_2proc_1node exit=0
RESULT run2tiny_jacobi3d_d_2proc_1node exit=0     (-x 128 -i 2 -w 0)
RESULT run2_jacobi3d_d_2proc_2node exit=0
```

Average iteration time, 32 chares, 512^3, 40 iterations:

| shape | base b17f4ac18 | this PR |
|---|---|---|
| 1 node, 2 processes | 1299.1 us | 1266.0 us |
| 2 nodes, 2 processes | 1235.0 us | 1264.1 us |

Both within run-to-run noise; the double-buffering costs nothing measurable. Full log and scripts: `/projects/mzu/lkale/software/gpudirect3963/` on Delta.

Still to come: the same seven runs on Anvil A100s (jobs 20751399 / 20751674, queued, est. 2026-09-18) and the HIP build on Frontier (down for maintenance today).

Delta note for reviewers running there: `srun --mpi=pmi2` does not bootstrap a multi-process reconverse job on Delta and fails silently (each rank prints "Starting Reconverse with 1 process"); use `--mpi=pmix`.

adityapb (approved):

lvkale:
**Frontier (AMD MI250X, HIP) validation of the jacobi3d double-buffering fix — all pass.**

Built `reviewed-with-reconverse` tip 0130de232 (reconverse submodule 58921e9) with `./build charm++ reconverse-linux-x86_64 amd --with-production` (ROCm 6.2.4, gcc 13.2; examples auto-detected GPU=hip, `--offload-arch=gfx90a`; all four targets built with no changes). GPU: `AMD INSTINCT MI200 (MCM) OAM LC MBA HPE C2` (MI250X). Jobs 5495903 (1 node, 2 GPUs) and 5495904 (2 nodes, runs 2/6/7 one process per node).

One node (frontier01822):
```
RESULT run1_jacobi3d_d_1proc exit=0      avg iteration 4494 us
RESULT run2_jacobi3d_d_2proc exit=0      avg iteration 2996 us   (2 processes, RDMA-get path)
RESULT run3_jacobi3d_host exit=0         avg iteration 5669 us
RESULT run4_jacobi3d_persistent exit=0   avg iteration 4399 us
RESULT run5_jacobi2d_z exit=0            avg iteration 4622 us
RESULT run6_verify exit=0                chare array PASS / chare group PASS / chare nodegroup PASS
RESULT run7_latency exit=0               regular 29.3 us @1B .. 1275 us @4MiB; zerocopy set printed
```
Two nodes (frontier[01822,01825]):
```
RESULT run1_jacobi3d_d_1proc exit=0      avg iteration 4489 us
RESULT run2_jacobi3d_d_2proc exit=0      avg iteration 2754 us   (inter-node RDMA gets, Slingshot)
RESULT run3_jacobi3d_host exit=0         avg iteration 5437 us
RESULT run4_jacobi3d_persistent exit=0   avg iteration 4421 us
RESULT run5_jacobi2d_z exit=0            avg iteration 4618 us
RESULT run6_verify exit=0                chare array PASS / chare group PASS / chare nodegroup PASS
RESULT run7_latency exit=0               regular 30.1 us @1B .. 1044 us @4MiB; zerocopy 12.8 us @1B .. 186 us @4MiB
```
Both two-process jacobi3d runs report `Starting Reconverse with 2 processes, 2 PEs` and `CkRdmaDevice> device RDMA registrations released on completion`. Every step ran under `timeout 300`; each took 1-3 s.

Two launch-side notes, neither in the code under test:
1. `+pe 2 +ppn 1` is rejected by this reconverse ("only one of +pe, +ppn and +p may be specified"); run 2 used `+ppn 1` alone (2 processes x 1 PE).
2. On Frontier `srun --mpi=pmi2` starts each task as its own 1-process reconverse job (the two ranks never join; verify/latency then abort on the PE count). Plain `srun` (cray_shasta) forms the 2-process job. The reconverse-site-run.sh frontier case currently inherits the `--mpi=pmi2` default, so the 2026-09-13 Frontier tier's "2-process" cases were in fact pairs of independent single-process runs; I will follow up separately.

Issue #238, opened 2013-06-20
Intel MPI dual binary method for execution on stampede


ericjbohm:
*Original issue: https://charm.cs.illinois.edu/redmine/issues/238*

---
Proper wrappers should be developed to support building host and MIC executables which can be launched using ibrun.symm as a combined host+MIC application with PEs running on both host and MIC cores.

There is some fussy business in getting -mmic passed correctly where it belongs and not passed where it does not belong. Quickthreads use of -seq is part of the problem. I have a manual process which works and produces dual executables which run across both host and MIC.  Rest of this subtask is wrapping this up in a clean build time option.

ericjbohm:
*Original date: 2013-06-20 23:14:21*

---
Checked in config files to support addition of mic to mpi-linux-x86_64 build target.

Assuming you have already run something like : 
module swap mvapich2 impi/4.1.0.030
on stampede, then binaries built against this target will now be for the MIC.
They should be run using ibrun.symm:  like
ibrun.symm -m yourmicbinary
if you also build a host version (same target name but without mic)
then 
ibrun.symm -m yourmicbinary -c yourhostbinary
can be used to launch a combined application running on both the hosts and MICs.

See the stampede user guide for more instructions.

A proper wrapper charmrun wrapper script for such shenanighans coming to a distribution near you, hopefully next week.

PR #3985, opened 2026-09-16
Converse manual: message ownership contract and the nokeep flag

lvkale:
## What

Documents, in the Converse manual (`doc/converse/manual.rst`), the ownership
contract for a message delivered to several destinations and the per-message
nokeep flag.

New subsection **Message Ownership and the nokeep Flag** under *Building
Messages*:

- Default contract: a broadcast, list send or multicast delivers a separate
  copy to each destination; the receiving handler owns its copy and may keep,
  modify or free it, and must eventually `CmiFree` it. The `AndFree` variants
  transfer the caller's buffer to the runtime; the other variants leave it
  with the caller.
- The nokeep flag: a header bit set by the sender, promising that no receiving
  handler retains the message past its return or modifies it. In exchange the
  runtime may deliver one shared buffer to several PEs of the same process,
  each PE's `CmiFree` dropping one reference and the last one releasing the
  buffer. A handler must therefore treat such a message as read-only and still
  call `CmiFree` exactly once, or hand it back unchanged.
- How it is set: Charm++ sets it for `[nokeep]` entry methods; a Converse
  program uses `CmiSetMsgNokeep(msg)` (accessor added to reconverse in a
  companion change) or the existing macro `CMI_MSG_NOKEEP(msg)`, which reads
  and writes the bit directly.
- Where it takes effect on reconverse today: `CmiWithinNodeBroadcast` and the
  within-node delivery of list sends and multicasts. Without the flag, one
  copy per destination.

Other changes in the same file:

- New entry for `CmiWithinNodeBroadcast`, which the manual did not document,
  including that the caller's buffer is consumed.
- *Writing Handler Functions* and *Broadcasting Messages* now point at the new
  subsection.
- `CmiSyncMulticast` and `CmiSyncListSend` state the reconverse behaviour:
  one network message per destination process, with delivery within a process
  sharing the buffer for nokeep messages and copying otherwise (the behaviour
  of reconverse PR #238).

## Why

The manual's only statement about message lifetime was that a handler must
`CmiFree` what it receives. Nothing said that a multi-destination send copies
per destination by default, or that a sender can opt into buffer sharing, so
the constraints a nokeep message places on the receiving handler were
undocumented outside the Charm++ `[nokeep]` attribute description.

## Testing

Documentation only; no code changes. `sphinx-build -W -n -b html doc/` builds
warning-free. Function names cited were checked against
`contrib/reconverse/include/converse.h`.

**Depends on** reconverse PR charmplusplus/reconverse#238, which adds `CmiSetMsgNokeep` / `CmiMsgIsNokeep` and the per-process fan-out this text describes; merge after that PR and the submodule pin bump that follows it, so the manual does not describe functions the pinned reconverse lacks.


lvkale:
Folding into the reconverse pin-bump PR that brings CmiSetMsgNokeep and the per-process fan-out (#238), so the manual text lands with the code it describes and there is one PR to review instead of two. Branch doc-converse-nokeep kept.

PR #3998, opened 2026-09-19
reconverse-cuda: build the CUDA surface against reconverse, not only netlrts

ritvikrao:
## Summary

`reconverse-cuda.yaml` gates the reviewed line, but both of its build invocations targeted `netlrts-linux-x86_64`:

```
./build all-test netlrts-linux-x86_64 cuda -j4 -g
./buildold charm++ netlrts-linux-x86_64 cuda -j4 -g
```

Meanwhile `reconverse-ci.yaml` builds `reconverse-linux-x86_64`, `reconverse-linux-arm8` and `reconverse-darwin-arm8`, and none of those jobs pass `cuda`. The two axes were tested separately and never crossed, so the reconverse machine layer — the one the 9.x device work is aimed at, and the one `tests/charm++/cuda/d2dtest/Makefile` names as its own build root — had never been built with CUDA anywhere in CI.

That gap was not cosmetic. `CMakeLists.txt` has reconverse+CUDA-specific code that no netlrts build reaches:

- it forces `RECONVERSE_ENABLE_CUDA` before adding the submodule, so reconverse's GPU-aware loopback copies and LCI's GPUDirect registration come up with the same backend as Charm++;
- it appends `-L${CUDA_DIR}/lib64/stubs -lcuda` to every charmc link line **under reconverse+CUDA and nowhere else**, because `liblci.so` uses the CUDA driver API directly.

The path filters already listed `src/ck-core/ckrdmadevice.*`, `src/ck-core/ckrdma.*`, `src/xlat-i/xi-Parameter.C` and the `contrib/reconverse` gitlink — so changes to the reconverse device path did trigger this workflow, and were then compiled against the machine layer they do not run on.

Adds `reconverse-cuda-buildonly`, a matrix over `reconverse-linux-x86_64` (ubuntu-latest) and `reconverse-linux-arm8` (ubuntu-24.04-arm), running the same build-only checks as the existing job: `libhybridapi.a` present and `cudahybridapi` gone, the charm++ CUDA examples and GPU-direct benchmarks, and `hapitest`, `d2dtest` and `gpumigrate` compiling.

It is a **subset** of `cuda-buildonly`, not a copy of it:

- `./build charm++` rather than `all-test`, and no `examples/ampi/cuda`: AMPI rides on TCharm, which `CMakeLists.txt` still excludes from reconverse builds (*"TCharm (not yet ported to reconverse builds)"*). When TCharm lands, the target should be added explicitly rather than by widening to `all-test`.
- no `buildold` leg: `buildold` does not mention reconverse anywhere, so the cmake path is the only path this machine layer has ever had.

`-Werror=vla` on arm8 carries over from `reconverse-ci.yaml`'s job of the same name. `fail-fast: false` so an arm-side toolkit problem cannot mask the x86_64 result. The netlrts `cuda-buildonly` job is untouched.

Still build-only and still advisory. Running any of this needs a GPU — `d2dtest` needs GPUs on more than one host to mean anything — so runtime validation stays with the HPC-site field below.

## What the new jobs found

The job earned its place on the first run. Every failure below was real, pre-existing, and invisible to every job in the tree; none were caused by this PR.

**1. Ubuntu's nvcc cannot build on arm64 at all.** The arm8 leg died in CMake's CUDA compiler-id probe, inside **LCI's** `enable_language(CUDA)` — code reached only because `RECONVERSE_ENABLE_CUDA` is forced for reconverse+CUDA:

```
/usr/include/aarch64-linux-gnu/bits/math-vector.h(108): error: identifier "__SVBool_t" is undefined
Call Stack: lci-src/cmake_modules/CudaHelpers.cmake:42 (enable_language)
```

Noble's `nvidia-cuda-toolkit` is CUDA 12.0, and glibc 2.39 declares the SVE types behind `#if __GNUC_PREREQ(10, 0) || __glibc_clang_prereq(11, 0)` — a compiler-**version** test, not an SVE feature test. nvcc advertises the host gcc's `__GNUC__` (13 on noble), so the block is always taken and nvcc's frontend has no such builtin types. No flag escapes it, and noble ships no gcc older than 12. The arm8 leg therefore takes CUDA 12.6 from NVIDIA's `ubuntu2404/sbsa` repo — five packages, not the `cuda-toolkit-12-6` metapackage, which pulls several GB of math libraries and profilers nothing here touches.

**2. A real VLA in hybridAPI.** With the toolchain fixed, the build stopped at `src/arch/cuda/hybridAPI/hapi_impl.cpp:1887`: `variable length array 'n_buffers_to_allocate' is used [-Werror=vla]`. `n_slots` is a runtime parameter of `createPool`, so this is a true VLA — a GNU extension, not standard C++. It survived because the two jobs that could catch it are disjoint: `reconverse-ci.yaml`'s arm8 job passes `-Werror=vla` but builds without CUDA, so it never compiles this file; the CUDA job built x86_64 without the flag. This leg is the first place the two meet. Fixed with `std::vector<int>`, which is what the rest of `createPool` already uses.

**3. cuBLAS was implicit.** `examples/charm++/cuda/matmul` includes `cublas_v2.h` and links `-lcublas`; x86_64 gets cuBLAS only incidentally, via `nvidia-cuda-toolkit`'s dependencies. Named explicitly for the curated arm set. A sweep of the compiled dirs shows `cublas_v2.h`, `cupti.h` and `nvtx3/nvToolsExt.h` as the only CUDA library headers reached — CUPTI is installed, and NVTX needs no package (every example's `-lnvToolsExt` is commented out, and `hapi_nvtx.h` compiles only under `-DHAPI_NVTX_PROFILE`).

The `-L${CUDA_DIR}/lib64/stubs -lcuda` line I flagged as a risk in review turned out **not** to be a problem on either leg: `/usr` on x86_64 and `/usr/local/cuda-12.6/lib64/stubs` on sbsa both resolve.

## Validation
- [x] CI green (reconverse-ci) — and all four `reconverse-ci` jobs plus both new legs and the untouched `cuda-buildonly` pass.

| check | result |
|---|---|
| `reconverse-cuda-buildonly (reconverse-linux-x86_64)` | pass |
| `reconverse-cuda-buildonly (reconverse-linux-arm8)` | pass |
| `cuda-buildonly` (netlrts, untouched) | pass |
| `reconverse-linux-x86_64` / `-arm8` / `-stress` / `-darwin-arm64` | pass |

Every step in both new legs ran and succeeded — none skipped — including `d2dtest` and `gpumigrate` compiling against reconverse on both architectures.

- Local runs (platform, tests): **none.** Author is on macOS/arm64 with no CUDA toolkit, so neither new job can be run locally; CI is the only execution these have had. Verified statically before pushing: the workflow YAML parses and both jobs resolve; `cuda` is a generic option in `buildcmake` with no per-arch whitelist, and `builddir_extra` makes the build dir `<triplet>-cuda`; `CMK_SMP=1` in the reconverse `conv-mach.sh` does not set `opt_smp`, so no `-smp` enters the directory name.
- **Runtime-core changes only:** n/a — no runtime code touched. The one source change is a build-time container swap in hybridAPI with identical semantics.

## Review notes

**The scope question is the one worth a decision.** This is no longer pure CI config: it carries the `hapi_impl.cpp` VLA fix, because the new job cannot be green without it. My view is that a test arriving with the fix for the bug it found is the right pairing, but splitting the VLA into its own PR and dropping `-Werror=vla` from the arm8 leg is a coherent alternative — say so and I will restructure.

Points to look hardest at:

- **The CUDA 12.6 pin on arm8.** It is a pin to a series, not to a documented fix: I could not find a release note naming the nvcc version that added SVE type support, so 12.6 was chosen as new enough and as what Grace-class sites actually run. It works, but a reviewer with better knowledge of the nvcc changelog may want a different floor. The two legs now use different toolkit sources, which is genuine divergence in exchange for arm64 coverage.
- **The deb cache key hashes `cuda_pkgs`.** Without that, editing the package list is a silent no-op on a warm cache — the restored deb set satisfies the "already downloaded?" guard, so the new package is never fetched and the failure reappears at compile time. Worth confirming you want that mechanism rather than hand-bumped key versions.
- **Whether the arm8 leg should be required or advisory.** It is advisory like the rest of this workflow, but it is now the only CI anywhere that compiles the GPU device path for the architecture the 9.x work targets.


ritvikrao:
First CI run resolved both of the open questions from the review notes, in opposite directions.

**x86_64: green end to end**, including `d2dtest` and `gpumigrate` compiling against reconverse. The `-L${CUDA_DIR}/lib64/stubs -lcuda` concern I raised does not bite there — `CUDAToolkit_TARGET_DIR` is `/usr`, and the link resolves. No build-system change needed.

**arm8: failed, and the cause is a distro toolchain bug rather than anything in this tree.** It died in CMake's CUDA compiler-id probe, before a line of Charm++ was reached:

```
/usr/include/aarch64-linux-gnu/bits/math-vector.h(96): error: identifier "__SVFloat32_t" is undefined
/usr/include/aarch64-linux-gnu/bits/math-vector.h(108): error: identifier "__SVBool_t" is undefined
5 errors detected in the compilation of "CMakeCUDACompilerId.cu".
Call Stack (most recent call first):
  lci-src/cmake_modules/CudaHelpers.cmake:42 (enable_language)
  lci-src/CMakeLists.txt:339 (enable_cuda_language_and_find_cuda_toolkit)
```

Ubuntu noble's `nvidia-cuda-toolkit` is CUDA 12.0, and glibc 2.39 declares the SVE types behind

```c
#if __GNUC_PREREQ(10, 0) || __glibc_clang_prereq(11, 0)
typedef __SVFloat32_t __sv_f32_t;
typedef __SVFloat64_t __sv_f64_t;
typedef __SVBool_t    __sv_bool_t;
#endif
```

That is a compiler-**version** test, not an SVE feature test. nvcc advertises the host gcc's `__GNUC__` (13 on noble), so the block is always taken and nvcc's frontend has no such builtin types. There is no flag out of it: the guard reads only `__GNUC__`, and noble ships no gcc older than 12. The only lever is a newer nvcc, so the arm8 leg now takes CUDA 12.6 from NVIDIA's `ubuntu2404/sbsa` repo — four packages (`cuda-nvcc`, `cuda-cudart-dev`, `cuda-cupti-dev`, `cuda-driver-dev`), not the `cuda-toolkit-12-6` metapackage, which pulls several GB of math libraries and profilers nothing here touches.

The toolkit root consequently differs per leg, so it moves into the matrix as `cuda_home` alongside the package set and the repo to add; the deb cache key follows. The netlrts `cuda-buildonly` job is untouched.

Worth noting where the arm8 failure landed: `enable_language(CUDA)` inside **LCI**, reached because `CMakeLists.txt` forces `RECONVERSE_ENABLE_CUDA` for reconverse+CUDA and nothing else. No netlrts build has ever executed that code — the new job hit previously-unexercised configuration on its first run, which is the argument for the job.

Remaining risk is narrowed to one thing: whether CUDA 12.6's nvcc on sbsa parses these headers cleanly and whether its layout satisfies `conv-mach-cuda.sh`'s `$CUDA_DIR/lib64` and `$CUDA_DIR/extras/CUPTI` expectations. The `report cuda toolkit layout` step now prints `nvcc --version`, the stub dirs and the CUPTI include dir before the build, so a layout mismatch is one line in the log rather than another round trip.

ritvikrao:
**All checks green.** Final state after four rounds — every failure along the way was a real, pre-existing defect that no job in the tree could see, and each one was in the reconverse+CUDA combination specifically:

| round | failure | cause |
|---|---|---|
| 1 | arm8: `__SVBool_t is undefined` in LCI's `enable_language(CUDA)` | Ubuntu's CUDA 12.0 nvcc cannot parse noble's own glibc 2.39 aarch64 headers |
| 2 | arm8: `variable length array 'n_buffers_to_allocate'` | a true VLA in `hapi_impl.cpp`, unreachable by any existing job |
| 3 | arm8: `cublas_v2.h: No such file` | cuBLAS is implicit on x86_64 via `nvidia-cuda-toolkit`'s deps |
| 4 | — | green on both legs |

x86_64 was green from round 1 and stayed green; the risk I flagged in the original review notes (`-L${CUDA_DIR}/lib64/stubs -lcuda`) does not bite on either leg.

The PR body has been rewritten to describe the final state rather than the open questions it was opened with. Two things there need a maintainer's call rather than mine: whether the `hapi_impl.cpp` VLA fix belongs in this PR or split out (the arm8 leg cannot be green without it, since `-Werror=vla` carries over from `reconverse-ci.yaml`), and whether the CUDA 12.6 pin on arm8 is the floor you want — it is reasoned from what Grace-class sites run, not from a release note naming the nvcc version that added SVE type support.

adityapb (approved):

